PR #18, opened 2016-04-08
Use gendered placeholder images

andylolz:
I don’t know if this is actually a desirable change… But
sending as a PR partly for consistency with YNR.


symroe:
Yeah, I wasn't sure if we needed it, but thanks, I think it does make things a little better!

PR #21, opened 2016-04-08
Copy edits to home page

JoeMitchell:
Clearer simpler text. Although I don't want the first exclamation mark. But I can't work out how to edit a pull request. Gah.

(I wonder about spelling out what we'll give you when you click through, but let's keep it super short.)

And removed the bold around 'everyone'...


tfgg:
This has now been handled in #38.

PR #42, opened 2016-05-01
Add Election to PersonPost

symroe:
This is a quick way to fix the fact that posts should have more than one election.  This will allow for filtering of people based on (election,post).

There is still a problem with posts and elections, but that will be addressed later.


wcivf/apps/people/managers.py
```diff
@@ -63,20 +63,23 @@ def update_or_create_from_ynr(self, person):
...
+                        slug=membership['election']['id'])
                     elections.append(election)
 
                 if membership['post']:
                     post, _ = Post.objects.update_or_create(
                         ynr_id=membership['post']['id'],
                         label=membership['post']['label'],
                     )
+                    if election:
+                        post.election = election
```
tfgg: Why do posts still have an election attribute?

symroe: Sorry, this is confusing.  The election here is only being used later in this call…in `PersonPost.objects.update_or_create`.  This is only to pass the object through to that method.

symroe: I mean, it's setting the post's election too – removing that will come in a future PR

tfgg: That is quite confusing. Can the field be properly killed off and the election object be passed through properly?

symroe: That creates quite a lot more work right now, and I'd rather just get this fix deployed so we're not showing wrong candidates, and then work on doing it properly when I'm more with it / can pair with someone…

tfgg: Ok sure, go ahead then.

PR #63, opened 2016-05-11
Removing 'Everyone has an election on 5 May'

JoeMitchell:
Because it's Old News.

### DIFF
--- a/wcivf/templates/home.html
+++ b/wcivf/templates/home.html
@@ -8,8 +8,7 @@
 <section class="columns large-6 large-centered">
   <div style="margin-top:2rem;">
     {% block pre_post_code_copy %}
-      <p style="font-size: 1.4rem;">Everyone in the UK has an election on 5 May 2016.</p>
-      <p>Find your next election! Enter your postcode below!</p>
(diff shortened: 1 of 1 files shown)


andylolz:
Could it say 23 June instead?

andylolz:
So that looks like this:
screen shot 2016-05-11 at 16 23 20

 * “Enter your postcode below!” can probably be removed, since it is repeated below.
 * “Find your next election!” is interesting… It hadn’t occurred to me that in the context of a referendum, “Who Can I Vote For?” is the wrong question. I suppose adding this heading makes the site relevant in a referendum context.

JoeMitchell:
Doesn't need to be relevant to referendum really - we're not doing anything
for it. But that's the first thing we can do before we tell people about
candidates... so saying 'Find your next election' seems right to me...

Agree on the double Enter your postcode. One design suggestion was just to
get all the copy in that white box, which is probably wise.

JoeMitchell:
@andylolz Maybe just this? 
image

andylolz:
Hey you know… The New News here is that everyone has an election on the 23 June!

PR #72, opened 2017-02-10
Lose the ref to 2016 on the homepage

JoeMitchell:
This is one suggestion...there might be other things that are better!

Or maybe just remove that line entirely...

### DIFF
--- a/wcivf/templates/home.html
+++ b/wcivf/templates/home.html
@@ -8,7 +8,7 @@
 <section class="columns large-6 large-centered">
   <div style="margin-top:2rem;">
     {% block pre_post_code_copy %}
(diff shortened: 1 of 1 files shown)


andylolz:
> Or maybe just remove that line entirely...

+1 for this.

> Who Can I Vote for
>
> Find out about the elections in your area by entering your postcode below.

^^ this is sufficient IMO.

JoeMitchell:
Thanks Andy, will change...

JoeMitchell:
Now hopefully good to go, though I'm not sure what it'll actually look like.

symroe:
Thanks both!

Issue #73, opened 2017-02-10
Update to account for wheredoivote api changes


chris48s:
When we deploy the recent updates to the wheredoivote API, we will need to update the code in

https://github.com/DemocracyClub/WhoCanIVoteFor/blob/ba26790610253bd6d7fefcea37e032e054e73871/wcivf/apps/elections/views/mixins.py#L125
and
https://github.com/DemocracyClub/WhoCanIVoteFor/blob/8c32f80e813ad744daec1e9c4487415f465d2ab8/wcivf/apps/elections/templates/elections/includes/_polling_place.html

to account for these changes.

I think the things we will need to account for are:
* API root location has changed (breaking change)
* Points are now represented as GeoJson features, not the previous bespoke format (breaking change)
* We can now find polling stations by address where postcodes cross polling districts (enhancement, non-breaking)

chris48s:
Now that wheredoivote api is deployed, this is *mildly* broken. Page still renders, but polling card section says:

![screenshot at 2017-02-13 21 22 03](https://cloud.githubusercontent.com/assets/6025893/22904004/aa54dc26-f232-11e6-8d7b-e9a6ac5a3e09.png)

as it is expecting to be able to populate "you should call [council name] on [number]" from a valid api call.

@symroe - if you're working on this, assign yourself or let me know otherwise assign me and I'll try and have a go at it this week.

PR #74, opened 2017-02-15
Fixes for Polling Place block

chris48s:
This is just a quick sticking plaster to point at the beta API endpoint and restore the previous functionality (note I've left the code that actually displays the map commented out since we are not yet serving up current data). This doesn't do anything with the address list or `/address` endpoint.

refs #73


symroe:
Thanks!

PR #89, opened 2017-04-11
Maybe using icons that exist 

davidmiller:
would increase your social media conversion clickthrough rate synergistic leveraging.

### DIFF
--- a/wcivf/templates/base.html
+++ b/wcivf/templates/base.html
@@ -70,10 +70,10 @@
         the digital infrastructure needed for a 21st century democracy</p>
       <p class="inline-list social">
         <a href="https://www.facebook.com/democracyclub/">
-            <i class="fa fa-facebook-square" aria-hidden="true" style="color: white"></i>
+            <i class="fa fa-facebook" aria-hidden="true" style="color: white"></i>
         </a>
         <a href="https://twitter.com/democlub">
-            <i class="fa fa-twitter-square" aria-hidden="true" style="color: white"></i>
+            <i class="fa fa-twitter" aria-hidden="true" style="color: white"></i>
         </a>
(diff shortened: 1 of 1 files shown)


chris48s:
Icons that exist are my favourite kind - I have a strong preference for them :D

http://fontawesome.io/icon/facebook-square/
http://fontawesome.io/icon/twitter-square/

Do you not see the icons on live? If not, this may be a more subtle issue...

screenshot at 2017-04-11 19 39 50

chris48s:
...or that, only with a screenshot of the right site:

screenshot at 2017-04-11 19 42 54

davidmiller:
@chris48s ooh interesting - I like subtle :) 

This is what I see on live: 

screen shot 2017-04-12 at 09 25 32

When I change the class names in the inspector to match this PR I see this:

screen shot 2017-04-12 at 09 27 48

davidmiller:
Also interesting: 
screen shot 2017-04-12 at 09 29 43

Is that related to me rather than WCIVF ? :)

chris48s:
I also get those 404s.
I can't reproduce the behaviour you're seeing in any OS/browser combo though. Tried it on Linux/Firefox, Linux/Chromium, Linux/Konquerer, Window/Firefox and Windows/Chrome (won't bother posting the screenshots, but I see all 3 icons every time). I thought it might be that I've already got it cached or something, but I know at least one of those browsers has never visited that site.

> ooh interesting - I like subtle :)

I, on the other hand, have a strong preference for the blindingly obvious!

chris48s:
@davidmiller - actually, quick thought.. do you use an adblocker/other content blocker? If so, can you try disabling it on that page and reloading? Do you see the icons now?

PR #90, opened 2017-04-11
all_settings Context processor

davidmiller:
Note: I've not actually run this code ever, I'm coding this exclusively on Github and relying on Travis to tell me if I screw up.
That might not be advisable.


wcivf/apps/core/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""
+Magic file names do magic things.
```
symroe: Actually, they magically break things. I don't know why, but pytest doesn't want them.

wcivf/apps/core/context_processors.py
```diff
@@ -2,9 +2,15 @@
 
 from .forms import PostcodeLookupForm
 
+# Remove this once we add all_settings to settings.CONTEXT_PROCESSORS
 def canonical_url(request):
     return {'CANONICAL_URL': settings.CANONICAL_URL}
 
+
+def all_settings(request):
+    return dict(settings={x: getattr(settings, x) for x in dir(settings)})
```
symroe: I reckon an `if not x.startswith('_')` would be good here – we don't want python's magic `__whatever__` vars in the context I don't think.

wcivf/apps/core/tests/__init__.py
davidmiller: Oh that's interesting ? What is it breaking ?

wcivf/apps/core/context_processors.py
davidmiller: I mean, I know where you're coming from, but I don't think you really need to because Django will blow up with a `TemplateSyntaxError` if you ever try to access anything that starts with an underscore... 

The only reason I was doing this is that I'm used to it from e.g. https://github.com/openhealthcare/opal/blob/v0.8.2/opal/context_processors.py#L20 which has not caused us a problem in the 3 years since we've been putting it in every application : )

davidmiller:
I'm going to return to my original problem, which was thinking about SEO - you can refactor to settings / context processors / whatever as you see fit :)

wcivf/apps/core/context_processors.py
```diff
@@ -2,9 +2,15 @@
 
 from .forms import PostcodeLookupForm
 
+# Remove this once we add all_settings to settings.CONTEXT_PROCESSORS
 def canonical_url(request):
     return {'CANONICAL_URL': settings.CANONICAL_URL}
 
+
+def all_settings(request):
+    return {x: getattr(s, x) for x in dir(s)}
```
mhl: I guess you mean `settings` instead of `s` here?

Sorry for the drive-by comment, but this worries me a bit. e.g. it'll make `SECRET_KEY` and `DATABASES` available in the context, which people should never even think of using in a template, and could be leaked in DEBUG mode if you're using DjDT for example. (Not that a production site should be running with `DEBUG = True`, but these things happen sometimes.)

Also, using `dir` will mean that the magic methods of the `LazySettings` object get added to the context, which might be confusing... Personally, I prefer a safe-list approach like: https://github.com/DemocracyClub/yournextrepresentative/blob/161c2004a4b701a3429837ab4e9631b8f8640f4c/mysite/context_processors.py#L24-L47

mhl: Oh, I see, there are subsequent commits that have fixes for this. Never mind. That's the risk of of reviewing commit-by-commit, I guess....

Issue #99, opened 2017-04-30
On  , registered voters in {{postcode}} will be able to vote in 0 elections


chris48s:
`og:description` reads "On  , registered voters in {{postcode}} will be able to vote in 0 elections" even if there are elections
e.g: https://whocanivotefor.co.uk/elections/EH11%201US/

PR #102, opened 2017-05-01
Fix rendering of og:description - addresses #99

mrwilson:
- Use postelections as the list to be grouped
- Use 'grouper' rather than pulling the date off the first election object.
- Default text if there are no elections

### DIFF
--- a/wcivf/apps/elections/templates/elections/postcode_view.html
+++ b/wcivf/apps/elections/templates/elections/postcode_view.html
@@ -5,7 +5,7 @@
 
 {% block page_title %}Elections in {{ postcode }}{% endblock page_title %}
 {% block og_title %}Elections in {{ postcode }}{% endblock og_title %}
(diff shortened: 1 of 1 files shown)


chris48s:
Hi @mrwilson. Thanks for having a look at this and sorry none of us had a chance to review it before the local elections last week.

I've had a quick look at your changes on my local copy. Your changes fix the problem correctly where there are >=1 upcoming elections, but your check for the case where there are 0 elections doesn't pick up the condition correctly. I still get "On , registered voters in {postcode} will be able to vote in 0 elections".

You can fix it by changing the check to either
`if not header_elections_by_date` or
`if postelections.count == 0`.
Can you update and then we'll merge.

Thanks

mrwilson:
Hi @chris48s - thanks for the feedback! I've added a commit to the pull request.

chris48s:
Sorted -thanks

PR #103, opened 2017-05-03
To create a link in your text, enclose the URL in <> angle brackets. …

davidmiller:
…For example: payload={"text": "<https://slack.com>"}

see here: https://api.slack.com/incoming-webhooks#adding_links

### DIFF
--- a/wcivf/apps/feedback/models.py
+++ b/wcivf/apps/feedback/models.py
@@ -52,7 +52,7 @@ class Feedback(TimeStampedModel):
                 },
                 {
                     "title": "Feedback URL",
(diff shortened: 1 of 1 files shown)


symroe:
Unless it's in their own message builder (where I tested the messages) where it hyperlinks any URL!

https://api.slack.com/docs/messages/builder?msg=%7B%22text%22%3A%22I%20am%20a%20test%20message%20http%3A%2F%2Fslack.com%22%2C%22attachments%22%3A%5B%7B%22text%22%3A%22http%3A%2F%2Fexample.com%22%7D%5D%7D

Issue #98, opened 2017-04-29
Schedule Feedback notifications


chris48s:
We have a small "Did you find what you were looking for?" micro-feedback form.

As well as storing this in a database, if someone fills this in, we also sling the data into a slack webhook so we get notifications about it.

This has the potential to become a bit annoying on really busy days (e.g: polling day). It would be nicer if we had an aggregated version of this. It would be really handy if someone could write a small django management command to run on a schedule, pick up any responses since the task was last run, summarise them (number of Yes responses, number of No responses + list any comments) and then send that to the webhook so that we get a maximum of one notification per hour/day/however often we schedule it.

See https://github.com/DemocracyClub/WhoCanIVoteFor/blob/2f20ef9571e615b6a75b52366792380393a34df7/wcivf/apps/feedback/models.py#L26 for the current code

PR #106, opened 2017-05-08
Create management command to batch feedback to slack - addresses #98

mrwilson:
- Look back one hour, unless --hours-ago flag is used
- Aggregate YES and NO into summary
- Take some non-empty comments and add them to the post


symroe:
Lovely, thanks! Sorry it took a while to merge!

PR #112, opened 2017-05-14
First pass at displaying hustings information on a postcode page.

davidmiller:
What this doesn't do is: 
```Should only need to add to the
"select_related" section of PostcodeToPostsMixin.postcode_to_posts if
it's a FK
```

I'm still not entirely sure I understand how the data model works here, but the select_related incantations I attempted resulted in errors :( 

I would have thought that e.g.
`pes = pes.select_related('post__husting')` would work - but presumably am missing something blindingly obvious.

Also: I am not attempting to pull data directly from a google spreadsheet, but from a local CSV of page 3 of that spreadsheet if that works...


wcivf/apps/elections/templates/elections/postcode_view.html
```diff
@@ -99,6 +99,36 @@
 
 
   </div>
+
+  {% if postelection.post.husting_set.count %}
```
symroe: I think `exists` is better here than `count`

wcivf/apps/elections/templates/elections/postcode_view.html
```diff
@@ -99,6 +99,36 @@
...
+    <div class="card">
+      <h3>Hustings</h3>
+      {% for husting in postelection.post.husting_set.all %}
```
symroe: Do we care about splitting these up in to "past" and "future" lists?

wcivf/apps/hustings/management/commands/import_hustings.py
```diff
@@ -0,0 +1,134 @@
...
+                starts, data.start_time
+            )
+        if data.end_time:
+            ends = dt_from_string(data.date)
+            ends = set_time_string_on_datetime(
+                ends, data.end_time
+            )
+        # This seems absurd. Maybe there's a better way to spell this
+        # in the ORM ? Maybe I don't understand the data model properly?
+        election = Election.objects.get(slug=data.electionid)
```
symroe: Sorry, my fault. I think the FK needs to be on `PostElection` not `Post`. `Post` is the area, `PostElection` is the place with an election. That will make this query simpler and the `select_related` work, I think.

wcivf/apps/hustings/models.py
```diff
@@ -0,0 +1,19 @@
...
+from django.db import models
+
+from elections.models import Post
+
+
+class Husting(models.Model):
+    """
+    A Husting.
+    """
+    post    = models.ForeignKey(Post)
```
symroe: See comment above, this should be to `PostElection`

tfgg:
image

wcivf/apps/elections/templates/elections/postcode_view.html
```diff
@@ -99,6 +99,36 @@
...
+        <h4>{{ husting.title}}</h4>
+      <p>
+        {{ husting.starts | date:"D b d P"}}
```
tfgg: Capitalize the month?

wcivf/apps/elections/templates/elections/postcode_view.html
davidmiller: Meh, if you say so...

wcivf/apps/elections/templates/elections/postcode_view.html
davidmiller: I don't as a first pass, I'd rather have them in.
But quite possibly as a second iteration.

wcivf/apps/hustings/management/commands/import_hustings.py
davidmiller: > "Post is the area, PostElection is the place with an election."

I'm not sure I even know what the difference between area and place with an election is.
Names are hard etc...

wcivf/apps/elections/templates/elections/postcode_view.html
davidmiller: I don't really understand why Django `| date` doesn't implement B as a filter param. Seems obviously useful to me.

wcivf/apps/hustings/management/commands/import_hustings.py
symroe: Post: Cambridge constituency
Election: GE2017
PostElection the election in Cambridge ward (there have been past elections here, so we need one of them for each election).

PostElection is the `through` table of the many to many relationship

symroe:
I think including the migrations are confusing here, since this is a new feature. Can you delete them and just create a single one for this change?

Also, because I can't run the migrations on my computer because of some odd DB problem…

davidmiller:
@symroe squashed

symroe:
Lovely!

PR #115, opened 2017-05-17
Support OpenSearch - resolves #114

### LINKED ISSUE #114: Implement Opensearch spec so I can search WCIVF from my address bar
c.f. http://www.opensearch.org/Specifications/OpenSearch/1.1#OpenSearch_description_document
e.g. https://github.com/opensearch.xml

Would mean changing the param name and adding a meta tag and xml doc I think...

mrwilson:
Support an OpenSearch xml file at /opensearch.xml, using the one from [GitHub](https://github.com/opensearch.xml) as a starting point.

### DIFF
--- a/wcivf/apps/core/urls.py
+++ b/wcivf/apps/core/urls.py
@@ -1,7 +1,7 @@
 from django.conf.urls import url
 from django.views.generic import TemplateView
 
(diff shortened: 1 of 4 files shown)


symroe:
Amazing!

PR #127, opened 2017-05-19
Show election dates

annapowellsmith:


annapowellsmith:
The most common user experience will be to type in a postcode on the home page and see the postcode view. I think it would be good to have the date of the election on the postcode view, just to remind people when it is! Also updating post view for consistency.

symroe:
Excellent point!

PR #129, opened 2017-05-19
Add candidate CVs.

annapowellsmith:
New app to import CVs from http://cv.democracyclub.org.uk/.

Display them on candidate pages, with thumbnail if available.

This will soon be replaced with separate cards, so styling is simple:

screen shot 2017-05-19 at 16 05 40


annapowellsmith:
How much do you care about test coverage? Happy to add a test for the management command...

wcivf/apps/peoplecvs/models.py
```diff
@@ -0,0 +1,13 @@
+from django.db import models
+
+from people.models import Person
+
+
+class CV(models.Model):
+    """
+    A candidate's CV.
+    """
+    person = models.ForeignKey(Person)
```
symroe: I think this can be a OneToOneField, unless there are likely to be more than one CV per person? Does the CVs API support this?

annapowellsmith: no, it's definitely one-to-one (checked with Francis) - so yes this should be a `OneToOneField`

wcivf/apps/people/views.py
```diff
@@ -35,6 +36,10 @@ def get_person(self, queryset=None):
                 'post',
                 'election'
             )
+        try:
```
symroe: If this is a OneToOneField then I think this is done for us. It should be added to the `select_related` tuple above though, as that will reduce the number of queries on the person pages.

symroe:
Excellent, thanks! I left a couple of comments but I'm happy to merge this and fix them later.

On test coverage: I'm not _very_ fussed about testing management commands – the only thing I've done it to make a smoke test that mocks the running of the command. Like I say, I'm not very fussed by this – it's more important to have tests for the core data models.

symroe:
FYI, I did these two commits to address the comments I made above:

9339c00f6b45108eb6184effecc59131a981e579

04e09a03fbf25f241028bd2f082cc496cb730ea1

PR #131, opened 2017-05-19
Add a candidate's latest tweets, where available.

annapowellsmith:
I'm not spending a lot of time styling and customising this, since
I assume these cards are going to get a revamp soon anyway.

screen shot 2017-05-19 at 18 04 01

### DIFF
--- a/wcivf/apps/people/templates/people/person_detail.html
+++ b/wcivf/apps/people/templates/people/person_detail.html
@@ -59,9 +59,7 @@
           Twitter
         </dt>
         <dd>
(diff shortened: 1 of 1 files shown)


davidmiller:
`I assume these cards are going to get a revamp soon anyway.` -> @symroe how valid is that assumption?

symroe:
The general idea is that each "thing" gets its own card. This might end up being really messy over time, we'll find out :)

davidmiller:
I've heard about this TDD thing, but didn't realise it meant Thing Driven Design.

symroe:
Thing David Does?

PR #138, opened 2017-05-22
Corporate overlords

davidmiller:
@Sym you should maybe check this over with a version of WCIVF that has run the full import_x more recently than me.

The CSV this is coming from is here: https://github.com/EdwardBetts/companies-house/blob/master/companies.csv

The real question with this is "How sure are we" / do we need disclaimer text.


wcivf/apps/core/models.py
```diff
@@ -18,7 +18,7 @@ class LoggedPostcode(TimeStampedModel):
     def __str__(self):
         return "{0}".format(self.postcode)
 
-def log_postcode(log_dict, blocking=True):
+def log_postcode(log_dict, blocking=False):
```
symroe: Can you explain this commit? Currently this change on its own will cause a write query per request to the 'postcode' page. Maybe the tests need to be updated rather than the code?

davidmiller: `Can you explain this commit? ` :) 

I think that's the way it is on Master: https://github.com/DemocracyClub/WhoCanIVoteFor/blob/master/wcivf/apps/core/models.py#L21?

Instead of installing Redis, I changed that default.
The original commit was a mistaken/subconscious `-a` flag to `git commit` - certainly I'm not proposing that you change from what's on master as part of this change, and I have no opinions on the matter either way, I just don't want to sysadmin a Redis :)

davidmiller: I could obv. re-write the history of this branch if you had loads of opinions about it.

symroe: Aaah, right, I see :)

symroe: Yeah, would you mind squashing this in to a single commit?

PR #148, opened 2017-05-23
position header better on mobile

### LINKED ISSUE #139: Header image is poorly positioned on mobile
This is from Chrome on Android:

screenshot_20170522-165927

BobHarper1:
should fix #139

Of course that change should also be made on [dc_base_theme](https://github.com/DemocracyClub/dc_base_theme/blob/master/dc_theme/static/dc_theme/scss/dc/_header.scss) (but I don't want to break something else!)


symroe:
Thanks Bob! This line is actually auto generated, so it'll get squished out next time I build the site.

Can you make the change (I can't actually see it in the diff!) to one of the sass files, or explain what change you made and I'll apply it to the base theme?

symroe:
Closing as this was fixed in the base theme

PR #164, opened 2017-05-24
fixes problems with wheredoivote

BobHarper1:
This should fix two issues with `_polling_place` cards for `postcode_view` I think this might have been due to changes made at some point to wheredoivote api @symroe?

1. Where a postcode did not have a known polling station, currently no card shows e.g. https://whocanivotefor.co.uk/elections/sa126hf/ because `polling_place.custom_finder` is `null`. The changes to `postcode_view.html` fixes this.

2. Where a postcode does not have a known polling station but `custom_finder` is a url _string_ (rather than the expected _object_), the current  results in the postcode_view giving no link to the third party site e.g. https://whocanivotefor.co.uk/elections/BT546RQ/. The changes to `_polling_place.html` fixes this. (Ok, it's a bit hacky in that it doesn't test for a url but it works.)

Tested that still working where there is full information for a polling station e.g https://whocanivotefor.co.uk/elections/sa126yr/. There may be more situations (?) but I don't know about them.

### DIFF
--- a/wcivf/apps/elections/templates/elections/includes/_polling_place.html
+++ b/wcivf/apps/elections/templates/elections/includes/_polling_place.html
@@ -12,21 +12,9 @@
     </p>
   {% else %}
     {% if polling_station.custom_finder %}
-
-    <!-- hacky fix if custom_finder is a url to election office/council website rather than an object -->
-      {% if not polling_station.custom_finder.base_url %}
         <p>You can find your polling station by
         <a href="{{ polling_station.custom_finder }}">
         following this link</a>.
-      {% elif polling_station.custom_finder.can_pass_postcode %}
-        <p>You can find your polling station by
-        <a href="{{ polling_station.custom_finder.base_url}}{{polling_station.custom_finder.encoded_postcode}}">
-        following this link</a>.
-      {% else %}
-        <p>You can find your polling station by
-          <a href="{{ polling_station.custom_finder.base_url }}">following this link and entering your postcode</a>.
-        </p>
-    {% endif %}
     {% else %}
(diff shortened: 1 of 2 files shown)


symroe:
Thanks @BobHarper1. @chris48s can you take a look at this one?

chris48s:
Well spotted @BobHarper1 . You're right that there have been API changes which we haven't properly accounted for, but the `custom_finder` property of the URL is now much simpler. It is either `null` or a `string`. See the API docs https://wheredoivote.co.uk/api/

As such, you don't need to bother accounting for an object with properties like `can_pass_postcode` and `base_url`. Its either going to be:

* `null` (in which case we probably just want to hide the div), or
* it will return a string containing either:
  * a link to the finder base_url  (if we can't pass it a postcode) or
  * a pre-assembled string with the postcode already in it (if we can), for example: https://stage.wheredoivote.co.uk/api/beta/postcode/BT282EY/

in either case, if it is a string we can just link to it and call it a day.

Can you have a go at simplifying that and update please.

BobHarper1:
Am I also right in thinking WDIV API doesn't need the auth code? (I'm working without one)

chris48s:
You can make up to 1,000 requests per day without an API key so for dev purposes you should be fine with anonymous usage, but if you want a key you can use give me a shout on slack :)

Obviously in production, WhoCIVF is making many more than 1,000 requests/day to the WhereDIV API but I think giving it an API key to use is handled by the deployment scripts .

BobHarper1:
That's ok - don't need it personally was just wondering why there was one for production but hadn't realised the rate limit :man_facepalming:

chris48s:
Cheers.

PR #189, opened 2017-05-30
Add manifestos where available on candidate pages.

### LINKED ISSUE #150: Add links to party manifestos in useful places
Joe made this sheet. It might be useful to include them somewhere:

https://docs.google.com/spreadsheets/d/1ag0FuUqUOJlP8nvVVxjxFh_2HDNH6WJ9srZSJh9KG_c/edit#gid=816418254

annapowellsmith:
Fixes #150.

Displays GE 2017 UK manifesto wherever available, and national manifestos if available for the candidate's country.

Uses a slightly hacky method to determine the candidate's country (by looking at the ONS ID), but should work for the purposes of this election.

Doesn't attempt to deal with there being two Labour parties and two Conservative parties - just attaches to the party ID shown in the [source spreadsheet](https://docs.google.com/spreadsheets/d/1ag0FuUqUOJlP8nvVVxjxFh_2HDNH6WJ9srZSJh9KG_c/edit#gid=816418254).

screen shot 2017-05-30 at 18 42 45


wcivf/apps/parties/management/commands/import_manifestos.py
```diff
@@ -0,0 +1,41 @@
...
+        )
+
+    def handle(self, **options):
+        with open(options['filename'], 'r') as fh:
+            reader = csv.DictReader(fh)
+            for row in reader:
+                party_id = row['party_id'].strip()
+                try:
+                    party = Party.objects.get(party_id='party:%s' % party_id)
+                    election = Election.objects.get(
```
symroe: Do you need to get the election for each row?

wcivf/apps/parties/management/commands/import_manifestos.py
```diff
@@ -0,0 +1,41 @@
...
+                        name='2017 General Election')
```
symroe: I think using the election ID here would be better (`slug="parl.2017-06-08"`)

wcivf/apps/people/views.py
```diff
@@ -66,9 +67,37 @@ def get_object(self, queryset=None):
         obj.title = self.get_title(obj)
         obj.intro = self.get_intro(obj)
         obj.text_intro = strip_tags(obj.intro)
-
+        obj.post_country = self.get_post_country(obj)
+        if obj.personpost and\
+                (obj.personpost.election.name == '2017 General Election'):
```
symroe: Same comment as above re using the election ID

wcivf/apps/people/views.py
```diff
@@ -66,9 +67,37 @@ def get_object(self, queryset=None):
...
+            obj.manifestos_2017 = Manifesto.objects.filter(
+                party=obj.personpost.party,
+                election=obj.personpost.election).filter(
+                Q(country='UK') | Q(country=obj.post_country))
+            obj.manifestos_2017 = sorted(
+                obj.manifestos_2017,
+                key=lambda n: n.country != 'UK')
         return obj
 
+    def get_post_country(self, person):
```
symroe: As you say, this is a bit of a hack, and fine for now (it works!), but I think I'd rather this were moved on to the `Post` model directly, in future. This can wait for a while, as we store the `territory_code` in [EveryElection](https://elections.democracyclub.org.uk/) and eventually this should be imported directly on to the Post model.

wcivf/apps/parties/management/commands/import_manifestos.py
annapowellsmith: doh

Issue #157, opened 2017-05-24
All elections should have "election" in the name


annapowellsmith:
The elections in the [list of elections](https://whocanivotefor.co.uk/elections/) should all have "election" in the name, because various places expect that. 

[Here's an example](https://whocanivotefor.co.uk/elections/parl.copeland.2017-02-23/copeland) of something that looks odd as a result.

Obviously we could write code to handle names in different formats, but it's probably easier just to make sure the names are consistent (could maybe enforce a check on save).

BobHarper1:
Just to add, though it may be obvious, "local election" and "by-election" should apply where relevant.

Most currently do but some don't eg [Copeland](https://whocanivotefor.co.uk/elections/parl.copeland.2017-02-23/copeland)

JoeMitchell:
An up-to-date example of an odd result: 

![image](https://user-images.githubusercontent.com/8096305/36665912-c66f5d78-1ae0-11e8-8ef3-f1dc7224ad38.png)

https://whocanivotefor.co.uk/elections/M30%209AG/

PR #216, opened 2017-06-07
If 'election' not in election name display 'Election'

BobHarper1:
Re #157. e.g. https://whocanivotefor.co.uk/elections/parl.manchester-gorton.2017-05-04/manchester-gorton

Only a temporary resolution, as it would be preferable if it displayed 'Local Election' or 'By-Election' or whatever, but as the issue says, that would be achieved by fixing the database so that that info was in the election name itself.


symroe:
Thanks for this – and sorry for the long time in addressing it!

I think this is mostly fixed now, so will close.

PR #217, opened 2017-06-07
WIP: Results

symroe:
## Postcode & post pages
screen shot 2017-06-07 at 17 06 06

## Results page (list of results as we get them)
screen shot 2017-06-07 at 17 32 56


wcivf/apps/people/models.py
```diff
@@ -16,6 +16,7 @@ class PersonPost(models.Model):
     party = models.ForeignKey(Party, null=True)
     election = models.ForeignKey(Election, null=False)
     list_position = models.IntegerField(blank=True, null=True)
+    elected = models.NullBooleanField()
```
symroe: Yep, this is used in most places, the ResultEvent model is just for recording times really

wcivf/apps/results/views.py
```diff
@@ -0,0 +1,21 @@
...
+from people.models import PersonPost
+
+class ResultsListView(TemplateView):
+    template_name = "results/results.list.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context['elections'] = []
+        election_qs = Election.objects.filter(
+            current=True).order_by('election_date', '-election_weight')
```
symroe: Ah yes, I'll add that as a data migration

wcivf/apps/results/templates/results/results_list.html
```diff
@@ -0,0 +1,21 @@
+{% extends "base.html" %}
+
+{% block content %}
+<section class="card">
+  <h1>Latest election results</h1>
+  {% for election_dict in elections %}
+    <h2>{{ election_dict.election }}</h2>
+    {% for result in election_dict.results %}
+      {% for person_post in result.person_posts.all %}
```
annapowellsmith: Ah - I think this template logic needs another look - just tried it against the live Atom feed and it's (i) printing results for non-elected candidates and (ii) iterating more than it should be. I see 16 cards for 4 candidates in Woodville ward, of whom only 1 was elected.

wcivf/apps/results/views.py
annapowellsmith: Actually South Derbyshire will still be shown top because it's on 11 May. Maybe instead we should do `.order_by('-election_date', '-election_weight')` to show the soonest first.

symroe: I think this is correct – the `-` would make oldest show at the top, wouldn't it?

annapowellsmith: Is there a way we could show GE2017 at the top, just for now?

PR #219, opened 2017-06-08
Tweak wording in polling place card

andylolz:
This isn’t great, really… It’s a bit clunkier in the case where the polling station is found (because it says “Your polling station” in consecutive sentences).


andylolz:
This was fixed in 347dd6bc.

PR #225, opened 2017-08-11
update mailing list signup forms

chris48s:
TODO:
* ~I've removed most of the `notifications` app already in 1dd6438, but on reflection it probably makes sense to get anything we need out of the existing database table and then delete the rest of the app (`models` and `migrations`) in this PR as well.~

* ~Blocked on https://github.com/DemocracyClub/Website/pull/77~

* Once deployed, we will need to generate an API key for it and then add `EMAIL_SIGNUP_API_KEY = 'f00ba2'` to `local.py`
  See https://github.com/DemocracyClub/dc_signup_form#configuration

* ~Blocked on https://github.com/DemocracyClub/dc_base_theme/pull/18~
* ~Once I've looked at implementing this on WhereDIV, I'll tag https://github.com/DemocracyClub/dc_signup_form with a `1.0` release and pin the dependency to it.~


chris48s:
Build is failing on master so failure is _probably_ unrelated to this, but needs fixing anyway as would be useful to be able to test this against working tests..

chris48s:
Good for review. There is an API key at https://democracyclub.org.uk/admin/signup_server/token/

Issue #235, opened 2017-11-03
Better SEO for by-elections


chris48s:
We would like to have better SEO for by-election searches by name. For example, we would ideally like people searching Google for "Gospel Oak by election" or "Camden Gospel Oak by election" to see
https://whocanivotefor.co.uk/elections/local.camden.2017-11-09/post-LBW:E05000134/gospel-oak
as a highly ranked result.

There are a few things we can probably do to help this:

1. Improve meta-data for this page/view - get relevant search terms like the ward name and "by election"/"by-election" into the meta description.
2. Improve discoverability of pages for current by elections by adding a "this week's by-elections" or an "election feed" to the front-page of the site. This will also help the site to show what we do when we're outside a major election cycle.

This probably requires us to differentiate a bit between by-elections and scheduled elections in the data model. For example, I wouldn't want the "this week's elections" panel on the home page to be listing hundreds of posts when there is a major scheduled local election coming up.

andylolz:
sitemap.xml currently includes https://whocanivotefor.co.uk/elections/local.camden.2017-11-09/camden-local-election but not the URL mentioned.

chris48s:
Interesting.. we probably don't want to go down to the post/ward level for scheduled elections otherwise the sitemap will be enormous, but for by-elections I think the ward/post name is much more important as an identifier/search term.

andylolz:
> otherwise the sitemap will be enormous

I think it’s fine to split the sitemap into multiple files…

chris48s:
so you can: https://support.google.com/webmasters/answer/75712?hl=en :+1:

chris48s:
election feed added in #237 - metadata/sitemap work still to do

symroe:
Site map still not done, but we now title by-elections properly, so that might help

michaeljcollinsuk:
@symroe doesnt look like the sitemap has been touched since this ticket was created so it think this should remain open

PR #237, opened 2017-11-29
Add upcoming by-elections to the home page

symroe:
screen shot 2017-11-29 at 10 59 45

### DIFF
--- a/wcivf/apps/elections/models.py
+++ b/wcivf/apps/elections/models.py
@@ -152,11 +152,21 @@ class PostElection(models.Model):
 
 
     def friendly_name(self):
-        # TODO clearly not every election is a by-election
-        # This will get better once we import more info from YNR/EE
-        return "{} ward by-election".format(
-            self.post.area_name
-        )
+        # TODO Take more info from YNR/EE about the election
+        # rather than hard coding not_wards and not_by_elections
+        name = self.post.area_name
+
+        not_wards = [
+           'W09000007',
+        ]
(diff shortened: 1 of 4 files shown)


symroe:
[Add upcoming elections to Who home page](https://trello.com/c/gjwYTQtw/1081-add-upcoming-elections-to-who-home-page)

wcivf/apps/elections/models.py
```diff
@@ -145,16 +145,20 @@ class Post(models.Model):
...
 class PostElection(models.Model):
     post = models.ForeignKey(Post)
     election = models.ForeignKey(Election)
     contested = models.BooleanField(default=True)
 
+
+    def friendly_name(self):
+        # TODO clearly not every election is a by-election
+        # This will get better once we import more info from YNR/EE
+        return "{} ward by-election".format(
```
chris48s: Given we are going to have to deal with a Welsh Assembly by-election soon (before Feb 7th - source: https://www.itv.com/news/wales/update/2017-11-20/alyn-deeside-vacant-formal-declaration/ ), lets not hard-code "ward" here. If we can pick up the post/geography type from the `Post` model, can we put in a special case? If not, lets stick to `return "{} by-election".format(` here.

chris48s:
refs #235

symroe:
Cool, fixup pushed that should deal with the one case we know about, we can add more if we need to, but obviously I hope we'll have fixed it properly before then

PR #244, opened 2018-01-10
Don't expect the area keys from YNR any more

symroe:
Future commits will:

1. Remove the use an Area at all
2. Use the ballot_paper_id from YNR

However this commit will fix the importer for the time being.

### DIFF
--- a/wcivf/apps/elections/managers.py
+++ b/wcivf/apps/elections/managers.py
@@ -59,8 +59,8 @@ class PostManager(models.Manager):
                 'role': post_dict['role'],
                 'group': post_dict['group'],
                 'organization': post_dict['organization']['name'],
-                'area_name': post_dict['area']['name'],
-                'area_id': post_dict['area']['identifier'],
+                'area_name': post_dict['label'],
+                'area_id': post_dict['id'],
             }
         )
 


chris48s:
This looks like it should sort the `KeyError: 'area'` error. Is a second fix needed for 

```
Traceback (most recent call last):
  File "/var/www/wcivf/code/wcivf/apps/people/managers.py", line 98, in update_or_create_from_ynr
    post = all_posts[post_id]
KeyError: 'WAC:W09000007'
```

Presumably this ends up changing the post-slugs for WhoCIVF? Do we need to think about impact on any downstream consumers, for example the post slug gets used in https://whocanivotefor.co.uk/api/candidates_for_postcode/?postcode=CT125AN

symroe:
No the `KeyError: 'WAC:W09000007'` is because the importer is trying to add a person to the post `WAC:W09000007`, but that post isn't being created. The importers need to run in order: elections -> posts -> people (with parties being run before people too)

symroe:
Ok `post_slug` I think it's the same – we're not really using the area concept in Who any way.

chris48s:
OK. One more question:

When WhoCIVF imports, it doesn't do a full import every time - it only grabs stuff that has been recently changed. If I go and edit a candidate that stood in an election last year, does that cause the same problem again? Do we need to do a 'full' import on deploy to update these for historic elections or anything, or will that also be safe?

symroe:
1. The recent edits import will work because the post exists and the ID hasn't changed.
2. A full import happens every hour anyway, so at most any error of that class would fix itself anyway.

symroe:
Also worth noting that, because "area" wasn't a real thing from EE, all the info should be the same anyway.

This is because, for example, `area_name`'s history was:
Previously:

EE: Division name
YNR: area_name & post_label
WCIVF: area_name & post_label

Now:
EE: Division name
YNR: post_label
WCIVF: area_name & post_label

PR #245, opened 2018-02-06
stop running tests on python 3.4

chris48s:
all the cool kids are running 3.5 or higher 😎


chris48s:
I'll do the same on WhereDIV once https://github.com/DemocracyClub/UK-Polling-Stations/pull/1124 is in because these 2 things change all the same lines, so I'd just end up doing it again.
EE is still deployed on 3.4 for the moment

PR #251, opened 2018-02-26
Update donation form for new GC API

symroe:
Note that before deploying this we'll need to update the `GOCARDLESS_ACCESS_TOKEN` in `local.py` in the deploy repo.


symroe:
Also, this is the same code as used for the DC Website, so it should all be fine. 

This only change here is adding [`GC_SESSION_KEY` as a random string](https://github.com/DemocracyClub/WhoCanIVoteFor/pull/251/files#diff-7410aaca0c376cf0087666b63b24e9d0R26). This is because we're using signed cookies where the session key is the data, meaning that the key hash changed between requests. This matters because gocardless uses a session token of some sort to confirm the payment after redirect. (see `confirm_redirect_flow`)

wcivf/settings/base.py
```diff
@@ -207,8 +209,9 @@
 }
 
 GO_CARDLESS_PAYMENT_NAME = "Democracy Club Donation"
-GO_CARDLESS_PAYMENT_DESCRIPTION = "Helping Democracy Club increase the quantity,"\
-" quality and accessibility of information on election candidates, politicians and democratic processes"
+GO_CARDLESS_PAYMENT_DESCRIPTION = "Helping Democracy Club "\
+"increase the quality of information on elections & the democratic processes"
+
 GOCARDLESS_REDIRECT_URL = "https://democracyclub.org.uk/donate/thanks/"
```
chris48s: After going through the donate process, I'm getting dumped to the corporate site, but you've also added a `/donate/thanks` in this PR. Seems like the _intent_ is that you want people to end up at https://whocanivotefor.co.uk/donate/thanks ?

symroe:
[Fix Who donate form](https://trello.com/c/NEglV3tj/1206-fix-who-donate-form)

PR #252, opened 2018-03-04
YNR CTA

### LINKED ISSUE #233: What should we do about elections with unlocked candidate lists?
It's quite cool that we're giving people loads of notice about upcoming elections on Who — and maybe it doesn't matter because there are so few searches at the mo' — but we are getting strange results like this: 

image

'0 candidates' is technically true at the moment, but not very helpful. And if I added a joke candidate via candidates.dc now, it would show until like April or something. I vaguely remember thinking that Who should only show elections once they're locked. But maybe I've made this up.

Ideally, we'd say something like 'There's an upcoming election. The deadline for nomination as a candidate is [`electiondate minus 19 working days`]. Find out more about standing as a candidate at beacouncillor.org.' Then flip it to show candidates at the deadline.

Any thoughts?

symroe:
Closes #233 I hope

1. Import post locking from YNR so we can use it in the project / templates
2. Deep link to YNR pages (Thanks @jf1)
3. Better wordsing on the postcode pages See images.

### Has candidates, post locked ("You can choose from")
screen shot 2018-03-04 at 17 13 09

### No candidates, post not locked (YNR CTA)
screen shot 2018-03-04 at 17 10 25

### One candidates, post not locked
screen shot 2018-03-04 at 17 10 12

### _N_ candidates, post not locked
screen shot 2018-03-04 at 17 10 08


jf1:
Looking good :+1: 

The phrases "You can choose from" and "We know of" both work for me. Expectations come with risk so I'd avoid them. The only slight tweak I'd do is for consistency: "We don't yet know of any candidates in [Area]. CTA-text."

Contact info can be added to candidates in locked posts so ideally the CTA would be present at least until a day or so before the election date.

I wonder if the CTA would be better below the candidates, then it wouldn't distract users who aren't interested in it. I figure this would be worthwhile at all times, not just near or on election day.

Trivially, for its text how about:
You may [add information to our candidates database](somewhe.re) to help improve this page - please do if you can.

symroe:
Updated with @JoeMitchell's wording:

screen shot 2018-03-05 at 10 13 34


screen shot 2018-03-05 at 10 14 49

As for a CTA for adding more info – that's a good idea @jf1. We do have a CTA on the person page like this:

screen shot 2018-03-05 at 10 16 46

But that could do with improving I expect. Let's do that in a issue different though, and keep this PR as a smaller improvement that we can deploy ASAP.

wcivf/settings/base.py
```diff
@@ -214,6 +214,7 @@
 
 GOCARDLESS_REDIRECT_URL = "https://whocanivotefor.co.uk/donate/process/"
 
+YNR_BASE_URL = "https://candidates.democracyclub.org.uk/"
```
chris48s: I don't think we need to define this twice: https://github.com/DemocracyClub/WhoCanIVoteFor/blob/9375842052b611d81e7e5d089fcc2641c5ade977/wcivf/settings/base.py#L176

Apart from that, looks good as long as @JoeMitchell is happy with the text

JoeMitchell: I'm happy!

Issue #257, opened 2018-04-03
Show info on voter ID trials


symroe:

chris48s:
From #255

> Now that we are serving this info up in Every Election and WhereDIV we should display it to users in affected areas on WhoCIVF.

symroe:
Gah, used waffle.io search and didn't see the other issue 🙄

chris48s:
Closed in #260

Issue #242, opened 2018-01-07
Remove CV block for non-parliamentry elections


chris48s:
One of the top blocks in our information hierarchy on candidate profiles for all election types is the 'Candidate CVs' block, but the CVs site is explicitly scoped to only deal with MPs. The 2 first bits of text on http://cv.democracyclub.org.uk/ are "Curriculum Vitae of future Members of Parliament" and "Ask your MP candidates for their CV!"

As such, for all other election types, we waste a chunk of real-estate right at the top of the page with a card that says

> We haven't got {{name}}'s CV yet. We think candidates should share their full CV with voters, to help us assess their skills and experience. Can you ask {{name}} to add a CV?

and then we link users to a page like http://cv.democracyclub.org.uk/candidates/DIW:E05005100 which throws a 500 error.

For all non-parliamentry elections we should replace this with a content block which has a higher chance of containing useful information.

chris48s:
Closed in #260

Issue #213, opened 2017-06-06
invalid postcode error is a bit ugly


chris48s:
This could look nicer:

![screenshot at 2017-06-06 17 06 37](https://user-images.githubusercontent.com/6025893/26839320-9632381e-4ada-11e7-9961-03e4fe6bd295.png)

JoeMitchell:
Quite aggressive innit! Perhaps something like:

>Oops, did you mis-type the postcode? We didn't recognise it. Please try
again.

>Sorry, we're not sure that's a UK postocde. Have you entered it correctly?

chris48s:
I was thinking more about the layout than the wording

BobHarper1:
Is it meant to show like this https://github.com/DemocracyClub/WhoCanIVoteFor/blob/master/wcivf/templates/home.html#L17-L19 - why doesn't it?

chris48s:
This is fixed in the latest theme so we just need to upgrade WhoCIVF to use the current theme version to fix this.

chris48s:
Closed in #260

Issue #203, opened 2017-06-02
Add friendlier 400 and 500 pages


annapowellsmith:
Would be nice if these were served by Django, and friendlier

chris48s:
Closed in #260

PR #260, opened 2018-04-16
2018 locals updates

symroe:
Single PR, but split in to hopefully meaningful commits.

@chris48s I don't think this needs a really in-depth review as it's mostly template changes (and more will come). A sanity check of the election metadata stuff would be welcome. We can chat in the morning – I'll do a re-image and deploy first thing (not going to attempt that this late in the evening).


chris48s:
This looks like it should be fine now you've added the pagination on the call to EE. Other changes look broadly sensible.

You'll be pleased to know This PR also solves #257, #242, #213 and #203 as well as the issues you've noted in the commit messages :D

I feel like we should be able to make a better job of dealing with 'charismatic' elections vs by-elections in a more sophisticated way to save doing stuff like 3101df8 and having to remember to undo it, but it the work probably begins upstream in EE/YNR

PR #282, opened 2018-08-28
Fix CI

symroe:
A few things here, best to read the commit messages.

* 69c1b35 is a missing migration from some time ago. Adding now as the `YNR_ID` depends on it.
* 632a087 is a collection of smaller changes mainly focused on the CI runs
* The other two should speak for themselves in the commit messages


wcivf/apps/people/migrations/0028_person_pk_to_int.py
```diff
@@ -0,0 +1,68 @@
...
+            schemaEditor._delete_constraint_sql(
+                schemaEditor.sql_delete_unique,
+                mp,
+                index_name
+            )
+        )
+
+    for model in models:
+        index_names = schemaEditor._constraint_names(model, index=True)
+        for index_name in index_names:
```
chris48s: This took a while to understand because in this function we are re-using `index_names` and iterating `for index_name in index_names` twice for 2 different reasons. For clarity this could probably be 2 functions rather than one.

chris48s:
Did you come to a conclusion about whether this will be an issue for models where the PK is legitimately an string (e.g: a ballot id, or whatever)?

wcivf/apps/people/migrations/0028_person_pk_to_int.py
```diff
@@ -0,0 +1,70 @@
...
+    models = [
+        apps.get_model("mentions", "mention_people"),
+        apps.get_model("mentions", "mention"),
+        apps.get_model("people", "AssociatedCompany"),
+        apps.get_model("people", "Person"),
+        apps.get_model("peoplecvs", "CV"),
+        apps.get_model("pledges", "CandidatePledge"),
+        apps.get_model("leaflets", "Leaflet"),
+    ]
+    indexed_to_delete = []
```
chris48s: Should this var name be `indexes_to_delete` rather than `indexed_to_delete`

symroe: Spot 👀  Fixed and squashed

PR #288, opened 2018-09-13
Improve docs (a bit)

chris48s:
Having just followed through the process of setting this project up from scratch again, I've made a few corrections to the setup instructions.

More generally, it would be useful to refresh the readme a bit as it explains the project from the perspective of "We've just started this project - these are our ambitions for it" as opposed to "We've been running this project for several years - this is what it does" (spoiler: they're not exactly the same :wink: ), but that is a slightly different job.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ htmlcov/
 .coverage
 .coverage.*
 .cache
+.pytest_cache/
 nosetests.xml
 coverage.xml
 *,cover
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -3,16 +3,15 @@ WCIVF requires Python 3.
 To install:
 
     sudo apt-get install python3-dev libpq-dev libjpeg-dev redis-server
-    sudo gem install --no-ri --no-rdoc sass -v 3.4.21
     pip install -r requirements/local.txt
 
 Create a Postgres database as detailed below, then:
 
     python manage.py migrate
     python manage.py import_elections
-    manage.py import_posts
-    manage.py import_parties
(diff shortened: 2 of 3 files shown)


chris48s:
Note: the reason this PR didn't trigger a build first time was because we need to turn on "Build forked pull requests" in CircleCI. This seems to be off by default (I guess this is because there is an abuse vector here: on a paid plan, someone can just rinse all your build minutes by submitting loads of spurious PRs to your public repos). I've turned it on for all our other projects while I was there, but we'll need to remember to enable it if/when we migrate more projects.

I'll have a look at the coveralls fail now. This is probably also a fix we'll need in lots of places..

chris48s:
Hmm. As far as I can tell this _should_ work. We've got `COVERALLS_REPO_TOKEN` correctly defined in the settings :/ I'll continue to chew it over..

symroe:
As I'm fairly sure none of these changes are test breaking, should I merge this, or is it useful to keep here for testing circle CI?

chris48s:
Aah yes I'd forgotten about this.

> I'm fairly sure none of these changes are test breaking

If you look at the build, you can see the "run tests" step passed. Its the "Update Coveralls" stage that is failing.

> is it useful to keep here for testing circle CI?

I'd quite like to figure this out if we're going to make wider use of Circle CI but I don't think there is anything special about this PR. I expect this will fail with any PR where the source branch is on a fork.. I'm just not sure what the fix is. All the posts I found about this problem said "set `COVERALLS_REPO_TOKEN`" but we've already done that :man_shrugging:

PR #292, opened 2018-09-26
Use YNR "next" API

symroe:
Changes to YNR needed after updating the Party endpoint.

Note that the images PR doesn't change anything in this code base as the 'thumbnail' key hasn't changed on `Person`.


symroe:
Depends on https://github.com/DemocracyClub/yournextrepresentative/pull/651

PR #294, opened 2018-10-02
Remove mentions

symroe:
Election mentions is dead and we don't have any of the data from it.

This app hasn't been used for a couple of election cycles, so let's just
remove the code

There is no data in the tables on prod or any recent backup


wcivf/apps/people/templates/people/person_detail.html
```diff
@@ -15,7 +15,7 @@ <h2 class="person_name">{{ object.name }}</h2>
     {% if object.photo_url %}
       <img src="{{ object.photo_url }}" class="person_photo detail" alt="profile photo of {{ object.name }}">
     {% endif %}
-  
+
```
davidmiller: Boo - less additions please

andylolz: But trailing whitespace is bad.

PR #305, opened 2018-11-01
Provide information about cancelled elections

chris48s:
We've already chatted about this a bit, so there shouldn't be any major surprises here, but just to summarise quickly..

* Add fields to store data on cancelled ballots
* Pull in data from YNR and/or EE APIs as applicable
* Show it in the front-end
  * I've kept the messaging fairly consistent with https://github.com/DemocracyClub/UK-Polling-Stations/pull/1433
  * I've taken the decision that if a ballot is cancelled we shouldn't show much additional info about anything other than the cancellation itself. In the context of WCIVF this is the key info and showing other info distracts from it.
  * We can always iterate on this as we start seeing real cases
* I ended up needing to do a bit of restructuring to make it easier to write tests which needed mock data replicating specific conditions. b0e20cf is a bit of a yak-shave, but I think it is a useful abstraction and doing that also fixed another open issue while I was there.
* As discussed on slack, I've not done anything about the API here as there are some wider issues to consider. We do need to sort that out though because now that we're not going to be deleting cancelled ballots in YNR, doing nothing here will mean we're serving up incorrect data if we don;t build in any explicit handling for cancelled elections (i.e: doing nothing about the API makes things worse - they don't stay the same). I'll open a new issue for this.


wcivf/apps/elections/models.py
```diff
@@ -216,6 +217,16 @@ def ynr_link(self):
             settings.YNR_UTM_QUERY_STRING,
         )
 
+    @property
+    def short_cancelled_message(self):
```
symroe: Just to make it clear what this does, can you call it `short_cancelled_message_html` or something that indicates it (might) output HTML?

I worry with this pattern (not new in this PR) that we introduce a catch all `mark_safe` and use it in templates when future refactoring might end up taking some sort of query string or use input and opening us up to an XSS. 

Appending "html" to a method name doesn't stop this of course, but it might make us think a little about how it's used.

chris48s: Have changed the name as suggested. In terms of how we handle this in general, as far as I can see there are a few ways this can work:

* run `mark_safe()` on some text in the model/view layer
* don't do that but render it with the `|safe` filter in the template layer
* make small template fragments which you can `{% include %}` in lots of places

If you reckon one of the others is a better way of doing it, I can revise. Regardless of whether you use `mark_safe()` or `|safe` as long as we're calling a model method/`@property` from inside a template, I don't think we can pass in any user input. In order to add a param we'd have to convert to calling in the view with a param and then passing to the template in the context object.

symroe: Yep, that's a good point about it coming from the model. In general `|safe` or a `*_html` method is something that tells me "this is unsafe", where as typing `foo.no_worries_mate` wouldn't. Th more things we can do to trigger that thought the better – I think this is a good pattern in this case. 

I already approved the PR so I can't do that again, other than to say 👍

PR #307, opened 2018-11-02
delete some dead code

chris48s:
we don't need this any more


symroe:
Thanks! 

I think the initial idea was sound, so I'll make an issue for us to think about in future. Clearly we don't care about optimizing the SEO for 2016 elections any more though.

PR #310, opened 2018-11-15
fix bug in person import

chris48s:
I just ran `import_people` on my local WCIVF and it did this:

```
$ ./manage.py import_people --since 2018-11-01
Downloading https://candidates.democracyclub.org.uk/api/next/persons/?page_size=200&updated_gte=2018-11-01T00:00:00
Downloading http://candidates.democracyclub.org.uk/api/next/persons/?page=2&page_size=200&updated_gte=2018-11-01T00%3A00%3A00
Importing page-200.json
Importing page-1.json
```

which looked a bit.. odd

I suspect in some situations we have been missing pages worth of edits by writing >1 API page to `page-200.json`. This will fix it, but I wonder if it would be better to just encode the whole URL so it is safe for use as a filename.

### DIFF
--- a/wcivf/apps/people/management/commands/import_people.py
+++ b/wcivf/apps/people/management/commands/import_people.py
@@ -81,7 +81,7 @@ class Command(BaseCommand):
             filename = url.split("/")[-1]
         else:
             if "page=" in url:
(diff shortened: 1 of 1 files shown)


symroe:
Ha! Good spot, thanks!

PR #311, opened 2018-11-21
API improvements

chris48s:
- expose more data via API
- improve API error handling
- add `candidates_for_ballots` endpoint
- update tests

### DIFF
--- a/wcivf/apps/api/tests/test_api.py
+++ b/wcivf/apps/api/tests/test_api.py
@@ -8,26 +8,10 @@ from elections.tests.factories import (
     ElectionFactory, PostFactory, PostElectionFactory)
 
 
-class TestAPI(APITestCase):
-    API_BASE = reverse('api:api-root')
-
-    def setUp(self):
-        self.election = ElectionFactory(
-            name="2017 General Election",
-            election_date="2017-06-08",
-            slug="parl.2017-06-08",
-        )
-        self.post = PostFactory(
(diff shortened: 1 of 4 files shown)


chris48s:
phew. I was fairly convinced I had gone completely mental for a moment there :smiley:

wcivf/apps/api/tests/test_api.py
```diff
@@ -34,6 +35,8 @@ def test_candidates_for_postcode_view_raises_error(self):
 class TestAPISearchViews(APITestCase):
 
     def setUp(self):
+        cache.clear()
```
symroe: This is fine, but I reckon the real solution is to use `django.core.cache.backends.dummy.DummyCache` in the settings the tests use. 

It might be a good idea to change the default cache to `DummyCache` and leave production to use Redis / whatever.

symroe:
Yep, all good just with that single comment. Happy to turn that comment in to an issue if you'd rather ignore for now.

chris48s:
Happy to do that, but I noted you made the opposite move in the relatively recent past: https://github.com/DemocracyClub/WhoCanIVoteFor/commit/632a08733b870d5fcb2a47ad5b2c87b9e4e37141

(presumably for reasons)

symroe:
Oh, hmm. I expect that was for reasons. Ok, let's ignore for now as it looks like a battle with circle looms if we touch it.

PR #326, opened 2019-01-18
New election IDs

symroe:
This:

1. Moves to using hybrid election / ballot paper IDs in URLs like YNR
2. Makes the slugs optional, but keeps them in for SEO funtimes
3. Sets up redirects for old slugs to new ones


symroe:
Whoops, I forgot Black was in the CI so I didn't format the commits as I made them. Sorry about that, and for the stray `blacken` commit.

chris48s:
Hmm. I'm struggling to get my head round this, but there's some proper mad stuff going on here.

We're trying to do redirects on stuff that may not have ever been a valid call in WCIVF in the first place, but also failing to redirect on stuff that was valid in the past. For example if call https://whocanivotefor.co.uk/elections/2015 it thinks "2015" is a postcode but http://127.0.0.1:8000/elections/2015/ redirects me to http://127.0.0.1:8000/elections/parl.2015-05-07/ but then conversly https://whocanivotefor.co.uk/elections/2015/2015-general-election is a valid call in WCIVF but if I call http://127.0.0.1:8000/elections/2015/2015-general-election it just throws a 404.

I'll try and go through this in more detail, but it seems worth raising this point now. In retrospect it seems like a bad design decision to have overloaded `/elections` with postcode search as well :( but meh - the past is the past. We are where we are. Any thoughts on how to attack this?

symroe:
> For example if call https://whocanivotefor.co.uk/elections/2015 it thinks "2015" is a postcode but http://127.0.0.1:8000/elections/2015/ redirects me to http://127.0.0.1:8000/elections/parl.2015-05-07/ 

The first one might be odd, but it's not "new odd" if you see what I mean. I think `/elections/2015/` -> `/elections/parl.2015-05-07/` is reasonable.

> http://127.0.0.1:8000/elections/2015/2015-general-election it just throws a 404.

Ah right. I was only testing on slugs with `.` in. I've pushed a new commit with a URL pattern in that matches this case.

chris48s:
Some more bugs:

elections index:
* http://127.0.0.1:8000/elections/ throws `NoReverseMatch at /elections/`

specific slugs:
* http://127.0.0.1:8000/elections/mayor.liverpool-city-ca.2017-05-04/ throws `NoReverseMatch at /elections/mayor.liverpool-city-ca.2017-05-04/`
* http://127.0.0.1:8000/elections/elections/mayor.greater-manchester-ca.2017-05-04/ throws `NoReverseMatch`
* http://127.0.0.1:8000/elections/local.castle-point.2016-05-05/ throws `NoReverseMatch`

There's also a bunch of places where we're setting up redirects for stuff that isn't in use. e.g: we've set up a redirect for `naw-2016-05-05-c` but the real URL is already using `naw.c.2016-05-05` https://whocanivotefor.co.uk/elections/naw.c.2016-05-05/welsh-assembly-election-constituencies
Similarly, there's a redirect for `mayor.greater-manchester.2017-05-04` --> `mayor.greater-manchester-ca.2017-05-04` but we're already using https://whocanivotefor.co.uk/elections/mayor.greater-manchester-ca.2017-05-04/foo

..or is that because we've re-imported? I'm getting confused now..

Should we only be specifying redirects for stuff that ever actually existed, or is it easier to keep the various different copies of this dictionary the same across applications? (I'm not sure I have a strong view on it - do you already have thoughts?)

symroe:
> `NoReverseMatch at /elections/`

I think this is because you have the old Elections in your database still. Can you either dump your DB and do a full import (slow) or just manually delete `2010` and `2015` in the shell?

> [*] throws NoReverseMatch

I can't reproduce this, but I think it might be caused by database related things too. Can you at least run `import_elections` and try again?

As for old redirects, yeah, I think it's easier to maintain a single dict and copy it about the place. I doubt `naw-2016-05-05-c` was ever (part of) a URL on Who, but I'd rather just have one mapping if poss (same for the CSV nginx redirects)

symroe:
Hmm. I don't see that. Can you give more context? It normally gives quite a good error…

chris48s:
Yeah. Just did a complete DB rebuild following https://github.com/DemocracyClub/WhoCanIVoteFor/blob/master/INSTALL.md Error I've got is:

```
django.urls.exceptions.NoReverseMatch: Reverse for 'election_view' with arguments '('tmp_2010.65686', 'member-of-parliament-for-thirsk-and-malton')' not found. 1 pattern(s) tried: ['elections/(?P<election>[a-z\\-]+\\.[^/]+)(?:/(?P<ignored_slug>[^/]+))?/$']
```

It may be this points to an issue in YNR rather than with this code..

symroe:
Ah ha! Yes, it's the by-election that should be deleted from YNR. What I'll do as a precaution is stop importing any `tmp_` ballot paper IDs or slugs.

symroe:
Yes good point and it also messes with the fixtures (that need updating at some point).

I've deleted that commit and removed the Ballot from YNR (as it should have been removed anyway – the election didn't happen).

If should work now, or at least once the cached API has updated itself at the YNR end

chris48s:
OK. I expect that same condition exists for some selection of our friends:

```
local.chichester.chichester-east.by.2018-08-16
local.wigan.bryn.by.2018-02-22
parl.bridgend.by.2016-05-05
tmp_2010.65686
tmp_gla.a.2016-05-05.unit_id:41441
tmp_local.falkirk.2017-05-04.UTW:boaness-and-blackness
tmp_local.lancashire.2017-05-04.CED:sklemersdale-east
tmp_local.south-buckinghamshire.farnham-royal-and-hedgerley.2016-05-05.DIW:E05010574
tmp_local.st-edmondsbury.haverhill-north.2016-05-05.DIW:E05007174
tmp_local.torquay.tormohun.2016-05-05.UTW:E05002110
parl.copeland.2017-02-23
parl.stoke-on-trent-central.2017-02-23
local.south-buckinghamshire.farnham-royal-and-hedgerley.2016-05-05
local.st-edmondsbury.haverhill-north.2016-05-05
local.torquay.tormohun.2016-05-05
local.glasgow.anderston-city.2016-05-05
local.southend-on-sea.2016-05-05
parl.manchester-gorton.2017-05-04
```

I think the actual code in this PR is fine independent of that though. I think until we fix those lot, something somewhere is going to be broken. I'll leave it with you to work out what order to attack those issues in.

symroe:
Thanks – I'll get this deployed and make a note to deal with the rest on Monday 👍

PR #337, opened 2019-02-26
Only delete LocalParty objects for current elections

symroe:
This is a bit of a hack, as we don't have a good way to manage local
parties.

We don't want to delete the 2018 parties from candidates, as they're
still useful. But we have different sheets for 2018 and 2019.

At some point, this should be done properly, but for now, just take the
2019 sheet and leave the data from the 2018 sheet as it is.


symroe:
[Import 2019 local parties in to Who](https://trello.com/c/i1C2PfI0/1675-import-2019-local-parties-in-to-who)

PR #352, opened 2019-03-26
display 2019 id pilot info

chris48s:
Relies on running `import_elections` to pull in the meta-data from EE


symroe:
Nice, thanks!

PR #363, opened 2019-04-06
Display expected SoPN publish date for posts

mrwilson:
This pull request aims to address a piece of feedback we get on the website when we don't have candidate lists yet (e.g. in the window between notice of poll and sopn publishing)

* "I'm looking for list of candidates."
* "You don't know who is standing!"
* "Looking to see who's standing for election"

It adds a piece of markup to show when we expect the candidate list to be/have been published, and a tooltip to explain why a candidates list might be delayed.

Screen Shot 2019-04-06 at 21 21 56


symroe:
Thanks for this!

Super rough feedback after looking quickly: given that so much of the traffic to this site is on mobile, do you think that `abbr` is accessible / usable enough? Might it be nice just to spell this out in full with a card / paragraph? There's normally not a lot of content on this page before the SOPNs are out, so I think we can add more text to the page.

symroe:
Also, we have the locked status of the ballot, so I expect we shouldn't show this message at all once the candidates are locked…

wcivf/apps/elections/helpers.py
```diff
@@ -79,3 +81,16 @@ def __call__(self, request, *args, **kwargs):
...
+
+
+def expected_sopn_publish_date(slug):
+    if not expected_sopn_publish_date.lookup:
+        expected_sopn_publish_date.lookup = StatementPublishDate()
+
+    try:
+        return expected_sopn_publish_date.lookup.for_id(slug)
+    except AmbiguousElectionId:
+        return None
```
chris48s: I accept this is probably currently non-trivial with the data we hold in WCIVF, but given the vast majority of elections we cover on WCIVF are local, the most common case is that this will throw `AmbiguousElectionId` and return `None`. For example, if we had this in place a week or two ago, this output would only be shown for something like 6 of the 5,000+ ballots happening on 2nd May.

In order to make this return a useful answer in most cases, we will need to find a way to derive the country.

mrwilson: Agreed.

To my mind, the easiest thing would be keeping keeping a list of "these wards are Scotland/Northern Ireland" and assuming everything else is England and Wales, but that's probably not sustainable.

I suppose the future-proof way of doing this would be to add country to the WDIV API so it can be pulled into WCIVF that way? WDIV feels like the correct place for country to exist.

symroe: We do store `territory_code` in EE, e.g https://elections.democracyclub.org.uk/api/elections/local.tower-hamlets.lansbury.by.2019-02-07/

I expect a quick way of doing this would be to do something like the existing [`import_metadata` on the election import](https://github.com/DemocracyClub/WhoCanIVoteFor/blob/master/wcivf/apps/elections/management/commands/import_elections.py#L28) but for [`import_posts`](https://github.com/DemocracyClub/WhoCanIVoteFor/blob/master/wcivf/apps/elections/management/commands/import_posts.py).

(This would require a DB change too)

mrwilson: Oh, sweet! I noticed there used to be "group" attached to the post object but that's now blank?

```sql
wcivf=> select "group", count(*) from elections_post group by 1 order by 2;
      group       | count 
------------------+-------
 Northern Ireland |    36
 Wales            |    40
 Scotland         |    59
 England          |   533
                  | 12005
```

If this field could be backfilled/made current, this would be even easier to pull out and I don't think a need for a migration?

mrwilson: Had a stab at this below. Interesting outputs:

```
wcivf=> select territory, count(*) from elections_post group by 1;
 territory | count 
-----------+-------
           |   410
 SCT       |   441
 ???       |     2
 GBN       |  1300
 WLS       |   902
 ENG       |  9520
 NIR       |    98
(7 rows)
```

The two `???` are from election ids starting `tmp_local`, but there are 410 posts without corresponding PostElections. What's the case for these, is this due to e.g. redrawing wards?

mrwilson:
This is about ready to review, I think @symroe @chris48s 

* It supports `local.*` by having an explicit branch.
* The message only appears on the post view, and only if there are (a) 0 candidates and (b) the election is not in the past
* It doesn't support `europarl` yet - is this a dealbreaker for merge?

Here's how the copy would look on a currently live election.

Screen Shot 2019-04-17 at 18 06 42

wcivf/apps/elections/management/commands/import_posts.py
```diff
@@ -63,3 +64,24 @@ def populate_any_non_by_elections_field(self):
...
+
+        post_elections_without_territory = PostElection.objects.filter(post__territory='')
+
+        for post_election in post_elections_without_territory:
+
+            ee_data = ee.get_data(post_election.ballot_paper_id)
+
+            territory = self.extract_territory(ee_data)
+
+            print("%s - %s" % (post_election.ballot_paper_id, territory))
```
symroe: Couple of points here:

1. We try (but don't always, for legacy reasons) to use `self.stdout.write` in management commands – this makes testing them easier as you can [replace the classes stdout class](https://github.com/DemocracyClub/UK-Polling-Stations/blob/03d1378103a0921b329edccf1abeeb8e611e9429/polling_stations/apps/councils/tests/test_council_importer.py#L58).

2. Having said that, I'm not sure if this is useful beyond initial debugging?

mrwilson: Gotcha, happy to make that change.

I included it mainly as a feedback mechanism - other management commands list out e.g. parties, posts as they're imported, and it felt right to include a way of demonstrating that something is actually happening.

Happy to change the copy to something cleaner too like "Adding territory to some.ballot.id - NIR"

symroe: At right, yes – ok let's keep the output, but convert to `self.stdout` 👍

mrwilson: These changes have been added in, merges resolved, and migrations updated to avoid conflicts with migrations made in the meantime.

symroe:
Woop!

PR #371, opened 2019-04-12
 only show You will have {n} votes for FPTP elections 

### LINKED ISSUE #370: Local elections: 'You will have {{X}} votes' not correct
If I go by my postcode (postcode_view?) I get a message for the upcoming that **tells me how many votes I will have,** e.g:
> You can choose from {{9}} candidates in the {{Newry}} ward. You will have {{6}} votes.
 https://whocanivotefor.co.uk/elections/BT35%208HF/

However, if I go directly to the `election/*electionid*/*ballot*` namespace (election_view?) for the same thing I do not get this message:
> {{9}} candidates are standing in this ward.
https://whocanivotefor.co.uk/elections/local.newry-mourne-and-down.newry.2019-05-02/newry/

Now, under the first example, the statement is incorrect. There is **only one vote per person** (in my example of Northern Ireland, we use the single transferable vote preference system, but still it counts as one vote - same in Scotland).

I believe the "number of votes" is taken from the number of places that there are e.g. there are 6 seats for councillors in the example. So this isn't the same thing and it's confusing for a voter.

chris48s:
Closes #370

Disclaimer: All of this is terrible code and I am a bad person for writing it.

In my defence:

* This does fix our immediate problem and makes things mostly not worse than they already were
* I've declared that "fix our entire data model/pipeline" is out-of-scope for a task I started at 4pm on a Friday :)

### DIFF
--- a/wcivf/apps/elections/management/commands/import_elections.py
+++ b/wcivf/apps/elections/management/commands/import_elections.py
@@ -16,8 +16,8 @@ class Command(BaseCommand):
         for page in pages:
             self.add_elections(page)
 
-        # Now import any metadata from EE
-        self.import_metadata()
+        # Now import any other bits from EE that aren't in YNR
+        self.import_extras()
 
     def add_elections(self, results):
         for election in results["results"]:
@@ -25,15 +25,19 @@ class Command(BaseCommand):
             if created:
                 self.stdout.write("Added new election: {0}".format(election["name"]))
 
-    def import_metadata(self):
+    def import_extras(self):
         pages = self.get_paginator(
-            "{}/api/elections/?current=true&metadata=1".format(settings.EE_BASE)
+            "{}/api/elections/?current=true".format(settings.EE_BASE)
         )
         for page in pages:
             for election in page["results"]:
                 try:
                     election_obj = Election.objects.get(slug=election["election_id"])
                     election_obj.metadata = election["metadata"]
+                    if election["voting_system"]:
+                        election_obj.voting_system_id = election["voting_system"][
+                            "slug"
+                        ]
                     election_obj.save()
-                except:
+                except Election.DoesNotExist:
                     pass
--- a/wcivf/apps/elections/migrations/0025_auto_20190412_1528.py
+++ b/wcivf/apps/elections/migrations/0025_auto_20190412_1528.py
@@ -1,0 +1,60 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def insert_voting_systems(apps, schema_editor):
+    VotingSystem = apps.get_model("elections", "VotingSystem")
+    db_alias = schema_editor.connection.alias
+    VotingSystem.objects.using(db_alias).bulk_create(
+        [
+            VotingSystem(
+                slug="AMS",
+                name="Additional Member System",
+                wikipedia_url="https://en.wikipedia.org/wiki/Additional_Member_System",
+                description="In an election using the Additional Member System, each voter casts two votes: a vote for a candidate standing in their constituency (with or without an affiliated party), and a vote for a party list standing in a wider region made up of multiple constituencies.",
+            ),
+            VotingSystem(
+                slug="FPTP",
+                name="First-past-the-post",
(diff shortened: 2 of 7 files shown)


wcivf/apps/elections/migrations/0025_auto_20190412_1528.py
```diff
@@ -0,0 +1,60 @@
...
+
+from django.db import migrations
+
+
+def insert_voting_systems(apps, schema_editor):
+    VotingSystem = apps.get_model("elections", "VotingSystem")
+    db_alias = schema_editor.connection.alias
+    VotingSystem.objects.using(db_alias).bulk_create(
+        [
+            VotingSystem(
```
chris48s: This is a straight copy & paste job from
https://github.com/DemocracyClub/EveryElection/blob/cfe79ec3b8d5b9da057ff3f3b84f9ff5841e0749/every_election/apps/elections/constants.py
It would probably be nicer if we could get this from an API endpoint or something, but for now we can't.

wcivf/apps/elections/migrations/0025_auto_20190412_1528.py
```diff
@@ -0,0 +1,60 @@
...
+                description="In an election using the Additional Member System, each voter casts two votes: a vote for a candidate standing in their constituency (with or without an affiliated party), and a vote for a party list standing in a wider region made up of multiple constituencies.",
+            ),
+            VotingSystem(
+                slug="FPTP",
+                name="First-past-the-post",
+                wikipedia_url="https://en.m.wikipedia.org/wiki/First-past-the-post_voting",
+                description="A first-past-the-post (abbreviated FPTP, 1stP, 1PTP or FPP) or winner-takes-all election is one that is won by the candidate receiving more votes than any others.",
+            ),
+            VotingSystem(
+                slug="sv",
```
chris48s: Wasn't me :see_no_evil: 
This is odd, but I'm replicating it because we want the slugs in EE and WCIVF to be the same

wcivf/apps/elections/management/commands/import_elections.py
```diff
@@ -16,24 +16,28 @@ def handle(self, **options):
...
         for election in results["results"]:
             election_obj, created = Election.objects.update_or_create_from_ynr(election)
             if created:
                 self.stdout.write("Added new election: {0}".format(election["name"]))
 
-    def import_metadata(self):
+    def import_extras(self):
         pages = self.get_paginator(
-            "{}/api/elections/?current=true&metadata=1".format(settings.EE_BASE)
+            "{}/api/elections/?current=true".format(settings.EE_BASE)
```
chris48s: Like various other things in this codebase, this will only work properly for `current` elections, so we break historic elections when we rebuild, but that's not a new problem.

wcivf/apps/elections/templates/elections/postcode_view.html
```diff
@@ -92,7 +92,7 @@ <h2>{{ postelection.friendly_name }}</h2>
               {{ postelection.post.label }} ward.
               {% endif %}
 
-              {% if postelection.winner_count %}
+              {% if postelection.winner_count and postelection.election.voting_system_id == 'FPTP' %}
               You will have {{ postelection.winner_count|intword }} vote{{ postelection.winner_count|pluralize }}.
```
chris48s: I'm just removing this for everything which isn't definitely FPTP (i.e: we know its something else, or `voting_system` is unknown) for now. We can probably do some better stuff here if we wanted. Do note though that populating the `voting_system` key here does make this code: https://github.com/DemocracyClub/WhoCanIVoteFor/blob/3c01c12e1bd1d7f7e9bc2f44e3bf95451c13221c/wcivf/apps/elections/templates/elections/postcode_view.html#L139-L147 work (which it didn't before), and that does provide some more general information about the voting system further down the page.

symroe:
> Disclaimer: All of this is terrible code and I am a bad person for writing it.

I like to think of it as you being consistent with the code style of the project ;)

👍 for reviewing your own PR too :D 

Given that, I reckon this is good. I'm about for a bit longer this evening, so I think it's better to :shipit: and fix thing later if needs be. Wooyeah, Friday night deploy!

PR #379, opened 2019-04-22
fix display for party list elections on post page

### LINKED ISSUE #377: Display candidates grouped by party on ballot page for party list elections
This page:
https://whocanivotefor.co.uk/elections/OX26%202NA/#election_europarl.2019-05-23
shows the candidates grouped in a sensible way

but this one:
https://whocanivotefor.co.uk/elections/europarl.uk-south-east.2019-05-23/south-east/
doesn't.

chris48s:
Closes #377 

Side note: I'm sure we talked about there being a way to make using a non-existent variable in a template a hard error in dev or testing or something at some point (that sounds too dull for me to have hallucinated it). I think this project in particular would benefit from that if its an easy thing to do as it is heavy on template logic.

### DIFF
--- a/wcivf/apps/elections/templates/elections/post_view.html
+++ b/wcivf/apps/elections/templates/elections/post_view.html
@@ -38,7 +38,7 @@
         <p><a href="{{ object.ynr_link }}">Add more at our candidate crowd-sourcing site</a></p>
       {% endif %}
 
-      {% if post.election.uses_lists %}
+      {% if postelection.election.uses_lists %}
         <h3>List candidates by party</h3>
         <p>You will vote for your preferred party rather than a candidate on this ballot paper</p>
         {% include "elections/includes/_people_list_with_lists.html" with people=person_posts.all %}


symroe:
> make using a non-existent variable in a template a hard error

In YNR we have this:

https://github.com/DemocracyClub/yournextrepresentative/blob/master/pytest.ini#L4

We should add this, it's a quick win here and for all our projects.

_However_ , that setting would not have caught this bug, as:

```
{% if thing_thats_not_in_context %}
```
is valid in Django template parlance. This means that `FAIL_INVALID_TEMPLATE_VARS` ignores anything in conditional blocks that don't render to the page. It also ignores missing vars for logic branches that are never hit in tests. Also, because exceptions are squashed to `None` (or maybe `False` in templates, you can do `{% if method.that.doesnt.exist %}` and never raise anything. I'm not sure if it's possible to change this at all.

chris48s:
Yeah that makes sense. There is still some class of error we can catch easier even if this exact example isn't in that class :)

PR #388, opened 2019-04-29
Updating main.scss

edjw:
Making `<summary>` tags have a pointer so they appear clickable.

It's being used in the 'How you vote' section and maybe elsewhere too

### DIFF
--- a/wcivf/assets/scss/main.scss
+++ b/wcivf/assets/scss/main.scss
@@ -125,3 +125,7 @@ dd {
       padding-left:10px;
     }
 }
+
+summary {
+  cursor: pointer;
+}


jf1:
It would be helpful if there was also a pointer before the "Show X list candidates" text that appears on EU elections such as https://whocanivotefor.co.uk/elections/europarl.uk-north-west.2019-05-23/north-west/
Let me know if you'd prefer this as a separate issue.

edjw:
> It would be helpful if there was also a pointer before the "Show X list candidates" text that appears on EU elections such as https://whocanivotefor.co.uk/elections/europarl.uk-north-west.2019-05-23/north-west/
> Let me know if you'd prefer this as a separate issue.

This PR would deal with that as well

symroe:
Thanks!

PR #393, opened 2019-05-02
Say when polls start if they start later today

zuzak:
At the moment the election view just says that "polls are open till
10pm", which implies that polls are currently open. As polls in the UK
open at 7am and not midnight, this is vaguely misleading.

This commit attempts to add a "from 7am" if one is looking at the
election view on the day of the poll, but prior to polls opening. After
7am you don't care when they opened, just when it closes.

(I haven't written any Django before, nor have it running locally, so I'm fully expecting this commit to be very broken, but I'm hoping CI will prove me wrong)


symroe:
Thanks for this! Can you explain what the `|add:"0"` is about?

zuzak:
I suspect it's me not knowing what I'm doing; it's supposed to coerce the string to something numeric that the `<` works with.

michaeljcollinsuk:
Closing this PR as the changes have been made in #623

PR #432, opened 2019-07-22
Lookup PostElection objects using ballot_paper_id

symroe:
This deals with the case that the ballot has been added in the past but
the election or post object has changed. The post will normally be the
thing to change, especially if the division was assigned a GSS code over
the time the election was active, and EE was updated.

Using the ballot_paper_id as a look up will mean that the ballot always
gets imported with the post and election from EE, but it might cause
orphan posts to exist.

### DIFF
--- a/wcivf/apps/elections/managers.py
+++ b/wcivf/apps/elections/managers.py
@@ -60,8 +60,7 @@ class PostManager(models.Manager):
 
         for election_dict in post_dict["elections"]:
             election = Election.objects.get(slug=election_dict["id"])
-            kwargs = {}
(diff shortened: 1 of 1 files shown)


chris48s:
> but it might cause orphan posts to exist

Do you reckon there's any advantage in sticking something to cleanup orphan posts on the end of the relevant import command? As far as I can tell, there isn't really a problem with this in WCIVF and they will eventually get cleared when we do a clean DB init, so we can _probably_ just ignore it...

symroe:
I reckon that's a good idea, but I'd like to be able to run tests on something that deletes things automatically. I'll make an issue to do do a clean up and then write some tests for it.

PR #446, opened 2019-08-15
Get tests passing

chris48s:
Before we start trying to upgrade this project to django 2, we're going to want passing tests to rely on


chris48s:
going to crack on and merge this so we've got it available for django 2 upgrade work

PR #478, opened 2019-10-22
WIP: YNR next importer

symroe:
This is a WIP PR but also can't be deployed until https://github.com/DemocracyClub/yournextrepresentative/pull/1044 is merged and deployed.

The major changes are that I'm importing posts, elections, ballots and candidacies in one command.

This is mainly because, well, that's what's in the new ballot endpoint. Importing posts and elections as two processes didn't make a lot of sense, even before, as there was no point in having one without the other.

We may as well set up candidacies when importing everything else as it's in the endpoint, and it actually gives us everything _essential_ for running the site – people on ballots.

The person profile information is managed by `import_people`.

Still to do:

1. The old `import_people --recent` would add and remove candidacies from people. Now this is managed by `import_ballots`, but there isn't a `--recent` flag. I expect it's more useful to have `import_people` manage this, so we can just check the candidacies when importing people and add and remove them as we need to. 

2. EE is used a lot (and was before), but that means we end up making a lot of calls to it. I expect we can be smarter about caching here, so I need to think about that a bit more


wcivf/apps/elections/import_helpers.py
```diff
@@ -0,0 +1,250 @@
...
+
+    Manages caching, and updating metadata from EE
+
+    """
+
+    def __init__(self, ee_helper=None):
+        if not ee_helper:
+            ee_helper = EEHelper()
+        self.ee_helper = ee_helper
+        self.election_cache = {}
```
chris48s: Isn't this just duplicating functionality that is already implemented in `EEHelper`? (same with `post_cache` )

https://github.com/DemocracyClub/WhoCanIVoteFor/blob/ccf440c333f47fc646990e3a757e1e8493cadc71/wcivf/apps/elections/helpers.py#L15-L16

wcivf/apps/elections/import_helpers.py
```diff
@@ -0,0 +1,250 @@
...
+                updated = True
+
+            if updated:
+                election.save()
+
+
+class YNRPostImporter:
+    """
+
+    """
```
chris48s: Insightful comment :D

wcivf/apps/elections/import_helpers.py
symroe: No, because one is the `election` from YNR and the other is the `election` from EE…

PR #481, opened 2019-11-01
expose email and photo_url in API person object

chris48s:


symroe:
This is fine – but let's make sure it's 

a) safe to deploy without breaking the devs.DC API (I'm sure you've done this, but would like confirmation) and 

b) documented on that site too

PR #483, opened 2019-11-02
fix(election): Start time is 7am, not 6am. Make consistent with copy.

mrwilson:
Joe pointed this out on Twitter.

The `start_time` being pulled in (and used for our Event data in Google) is hard-coded to 6am, whereas almost all copy in the codebase declares that polls open at 7am.

This minor change makes the code consistent with the copy.


symroe:
Thanks!

PR #522, opened 2020-02-10
API improvements

symroe:
This adds some fields we need for https://github.com/DemocracyClub/aggregator-api/issues/125 and related improvements to the devs.DC API


wcivf/apps/api/views.py
```diff
@@ -74,6 +80,11 @@ def list(self, request, *args, **kwargs):
                 "ballot_locked": postelection.locked,
                 "replaced_by": postelection.replaced_by,
                 "candidates": candidates,
+                "voting_system": VotingSystemSerializer(
+                    postelection.voting_system
+                ).data,
+                "seats_contested": postelection.winner_count,
+                "organisation_type": postelection.post.organization_type,
```
chris48s: What are your thoughts on what to do with this field once the data gets to devs.DC? Are you thinking of just using it for sorting the ballots and then binning the value, or exposing it?

If we're just going to use it for sorting and discard it, I don't suppose this matters:

If we're going to also expose it via the `/postcode` and `/address` endpoints, now that devs.DC has the `/elections` endpoint, the direction we should really be heading in is that a ballot object from the [/postcode](https://developers.democracyclub.org.uk/api/v1/#postcode-search-2) or [/address](https://developers.democracyclub.org.uk/api/v1/#address-search-2) is a cut-down version of a ballot object from [/elections](https://developers.democracyclub.org.uk/api/v1/#elections-2), so I wouldn't want to randomly expose `organisation_type` as a top level ballot key when it isn't a top-level key on `/elections`. We should probably include (minimal?) `organization` / `division` keys in the `/postcode` and `/address` ballot objects.

symroe:
Good to know I did it the hard way!

_But_, the `voting_system` and `seats_contested` fields were basically the same work as doing it in WDIV (as we imported them already), and the `organisation_type` in WCIVF will allow us to order ballots there in the same way as in the API (should we want to keep the same order), so I think it's still useful work

wcivf/apps/api/views.py
symroe: I was assuming we'd just use it for ordering and not present it to the front-end. Not sure there's any value in exposing it?

chris48s: In that case, it doesn't really matter

chris48s:
Yep, so my point wasn't that we should undo this and do it differently somewhere else. Its worth bearing in mind for future though, especially if you don't already have the fields imported here.

PR #528, opened 2020-03-18
Coronavirus updates

symroe:
1. Generic message
2. Show metadata with a key "coronavirus_message" under the ballot title

### DIFF
--- a/wcivf/templates/base.html
+++ b/wcivf/templates/base.html
@@ -50,7 +50,6 @@
 {% endblock extra_site_css %}
 
 {% block top_banner %}
-<main>
   <div class="container">
     <div class="card postcode_card">
       <h2>Coronavirus</h2>
@@ -66,7 +65,6 @@
         contact your council</a> for more information</p>
     </div>
   </div>
-</main>
 {% endblock top_banner %}
 
 {# {% block bottom_banner %} #}
--- a/wcivf/templates/base.html
+++ b/wcivf/templates/base.html
@@ -55,7 +55,7 @@
     <div class="card postcode_card">
       <h2>Coronavirus</h2>
       <p>
-        <strong>Local by-elections</strong>: most elections will continue as planned.
+        <strong>Local by-elections</strong>: we anticipate most elections will continue as planned.
       </p>
       <p><strong>May 2020 scheduled elections</strong>: The Government have announced
         they plan to postpone the election in May until 2021. We will update
--- a/wcivf/templates/base.html
+++ b/wcivf/templates/base.html
@@ -49,6 +49,26 @@
     {% endif %}
 {% endblock extra_site_css %}
 
+{% block top_banner %}
+<main>
+  <div class="container">
+    <div class="card postcode_card">
+      <h2>Coronavirus</h2>
+      <p>
+        <strong>Local by-elections</strong>: most elections will continue as planned.
+      </p>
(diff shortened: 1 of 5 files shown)


wcivf/templates/base.html
```diff
@@ -49,6 +49,26 @@
...
+      <p>
+        <strong>Local by-elections</strong>: we anticipate most elections will continue as planned.
+      </p>
+      <p><strong>May 2020 scheduled elections</strong>: The Government have announced
+        they plan to postpone the election in May until 2021. We will update
+        ths information on this site once the postponement has become law.
+      </p>
+
+      <p>You should <a href="https://www.electoralcommission.org.uk/i-am-a/voter/your-election-information">
+        contact your council</a> for more information</p>
```
chris48s: The text here says "contact your council", but its a link to the Electoral Commission

wcivf/templates/base.html
```diff
@@ -49,6 +49,26 @@
     {% endif %}
 {% endblock extra_site_css %}
 
+{% block top_banner %}
+<main>
```
chris48s: Our primary content is already inside a `<main>`. You're only supposed to have one `<main>` element in a document.

chris48s: Having looked at it in a bit more detail (I'm just looking at nicking the same thing for WDIV), I can see why you've done this. My proposed solution would be to fix this in the theme: https://github.com/DemocracyClub/dc_base_theme/pull/52 tag a release and bump the theme in WDIV/WCIVF rather than have 2 `<main>`s

wcivf/templates/base.html
symroe: Where should it link to?

wcivf/templates/base.html
symroe: Great – I'll update the theme when that's in ad bumped

wcivf/templates/base.html
symroe: To be clear: this shows on every page, including those where we don't have a postcode. The link is to the EC's "enter your postcode" page, where we provide council contact information. I could link to WhereDoIVote where this info is also, but that doesn't seem right really.

wcivf/templates/base.html
chris48s: :+1:  `0.3.11` is tagged

wcivf/templates/base.html
chris48s: I clocked what link it was. I just totally didn't get what the point of sending me there was, or assumed that wasn't the right text to describe it.

Just taking one step back from this, is "contact your council" actually the CTA here?
i.e: does your council actually know any more about this specific issue right now? (as opposed to an issue around say registration/applying for a postal vote/knowing where your polling station/etc is where its a local issue).

I don't have a better suggestion for a "go here for more information about elections being cancelled due to coronavirus" link because there isn't really one (yet).

If you're totally convinced its the right CTA/link, feel free to crack on..

symroe: Right, got you. You're right that the council won't know, but if they say "yeah, you can vote tomorrow" it's more likely to be true than us saying it. They will know first if an election is cancelled / changed / etc, and we might be giving misinformation for some time that wouldn't help anyone.

I reckon I'll keep it there for now, but it's not like it's hard to change later.

PR #542, opened 2020-05-05
If an election is cancelled don't add it to ical feed.

GeoWill:
Based on error report here: https://twitter.com/DanielJUK/status/1257596394250084356

Any pointers on getting a test running would be great. When I inspect the current 'response' object in `wcivf/apps/elections/tests/test_postcode_views.py:PostcodeViewTests:test_ical_view`, it hasn't picked up any elections. So changing the cassette to include  a cancelled election seems like it would need something changing with the existing test first.

### DIFF
--- a/wcivf/apps/elections/views/postcode_view.py
+++ b/wcivf/apps/elections/views/postcode_view.py
@@ -77,6 +77,8 @@ class PostcodeiCalView(
         cal.add("prodid", "-//Elections in {}//mxm.dk//".format(postcode))
 
         for post_election in self.postcode_to_ballots(postcode):
+            if post_election.cancelled:
+                continue
             event = Event()
             event["uid"] = "{}-{}".format(
                 post_election.post.ynr_id, post_election.election.slug


symroe:
Thanks for this!

In other places on the site, and across DC, we still show the election, but display it as cancelled. I think this is useful, as it's better to tell someone about a cancelled thing than to just make it disapear.

To that end, I think using the `CANCELLED` flag on the event would be better. I'm not sure, but I think that looks like:

`event.add("status", "CANCELLED")`

Maybe adding `(cancelled)` after the title would be useful, too? I'm not sure how calendar clients show cancelled events.

(we could add and notifiactions to the description too, but that can wait)

As for tests - I'm less and less convinced that VCR is the right way to go about testing, as it's so hard to maintain. It's great for getting a quick snapshot of an API at that time, but if the upstream API changes (like, if the elections aren't current any more, in our case) then it's almost impossible to reproduce the response.

I've wanted to move towards a set of mocked responses for a while, so maybe it's best to start here and just add the elections and EE response as a mock?

Is see that none of the ical view is tested anyway, so feel free to punt this to an issue if you'd rather!

GeoWill:
Tested this a bit with google calendar and gnome-calendar. 
* If you import a cancelled event on either of them it just doesn't show up. 
* If you import an even, and then import a cancelled version of it google removes it while gnome does nothing (known issue: https://gitlab.gnome.org/GNOME/gnome-calendar/-/issues/503).
Both of them act silently, so there's no information that anything has updated. They can both link to a url too, but I haven't tested how that works. I guess similarly to you I don't really know what the expected behaviour is...

symroe:
Hmm, that's frustrating. Thanks for investigating!

Given what you've found, we could either do as the PR currently does and remove events from the feed, or change the title to mark it as cancelled?

Do you have opinions about it? Happy to just remove it, as I exect that's the best thing to do when populating other people's calendars!

GeoWill:
> Do you have opinions about it?
Not strong ones! I'm a fairly light user of digital calendars. If pushed, my hunch is that it's not a good communication pathway, so I would just magic it away by setting it to `CANCELLED`.  Leaving it in, but appending ' - cancelled' to the title still takes up a slot in someone's day, until they notice, and manually remove it. If it just vanishes, either they don't notice it's gone and there's no harm done, or they do, and then there's lot's of places they can go investigate to find out more.

PR #609, opened 2020-12-28
feat(sopn-publish-date): Switch to using uk-election-timetables

mrwilson:
This supercedes the sopn-publish-date library and enables wider usage of election timetables in future.


symroe:
Nice, thanks!

Also a great indication that the new API changes are working for us, just based on the number of lines deleted in this PR :+1:

PR #616, opened 2021-01-05
2021

pmk01:


wcivf/templates/base.html
```diff
@@ -97,7 +97,7 @@ <h3>Made by Democracy Club</h3>
       <li><a href="https://facebook.com/DemocracyClub">Facebook</a></li>
       <li><a href="https://github.com/DemocracyClub">GitHub</a></li>
     </ul>
-    <p>Copyright &copy; 2019 Democracy Club Community Interest Company No: <a href="https://beta.companieshouse.gov.uk/company/09461226">09461226</a></p>
+    <p>Copyright &copy; 2021 Democracy Club Community Interest Company No: <a href="https://beta.companieshouse.gov.uk/company/09461226">09461226</a></p>
```
symroe, line 100: Again, if changing a PR is hard then I can do this, but a better thing to do here would be to add `{% now 'Y' %}` rather than the year. This will pull the year out of the current time and mean we never need to change it again :)

PR #618, opened 2021-01-07
Update INSTALL.md with mac os installation instructions

michaeljcollinsuk:
Add instructions for installing on Mac-os


INSTALL.md
```diff
@@ -1,11 +1,47 @@
...
+
+Pyenv is recommended to install python versions. You can find installation instructions for pyenv [here](https://github.com/pyenv/pyenv#installation) - installation via homebrew is recommended. You can then install python 3 with:
+
+    pyenv install 3.6.12
+
+If using pyenv, after installation use the command `pyenv local 3.6.12` to set this version for the project locally. This will create a `.python-version` file with the specified python version.
+
+You can then create a [virtual environment](https://docs.python.org/3/tutorial/venv.html) with the command:
+
+    python -m venv env
```
VirginiaDooley, line 25: This command didn't work for me but `python3 -m venv env` did.

.gitignore
```diff
@@ -23,6 +23,7 @@ var/
 *.egg-info/
 .installed.cfg
 *.egg
+.python-version
```
symroe, line 26: Is there a reason this shouldn't be checked in? I think it's enough of a standard that it's useful to have, especially if we're all using pyenv

INSTALL.md
```diff
@@ -1,11 +1,47 @@
...
     sudo apt-get install python3-dev libpq-dev libjpeg-dev redis-server
     pip install -r requirements/local.txt
 
-Create a Postgres database as detailed below, then:
+
+## Installation requirements on macOS
+
+Use homebrew to install the following dependencies:
+
+    brew install postgresql libjpeg-dev postgis
```
symroe, line 15: FWIW, I always used to use https://postgresapp.com/ when I was on a mac, as it resolved a tonne of problems installing with Brew. Also, multiple version support was super useful.

Also, I'm not 100% sure, but I don't think postgis is a requirement for this project any more. We use it elsewhere though, so good to have it installed.

INSTALL.md
```diff
@@ -1,11 +1,47 @@
...
+
+If using pyenv, after installation use the command `pyenv local 3.6.12` to set this version for the project locally. This will create a `.python-version` file with the specified python version.
+
+You can then create a [virtual environment](https://docs.python.org/3/tutorial/venv.html) with the command:
+
+    python -m venv env
+
+This will create a virtual environment in the directory `env`. You can then activate the virtual environment with:
+
+    source env/bin/activate
```
symroe, line 29: This might not be helpful here, but you don't have to make your virtualenvs in this directory. I have a `~/.envs/` directory that I use for everything (e.g, `~/.envs/wcivf/`). This means I can e.g `grep` the project root without getting results from 3rd party packages (although this might be useful sometimes).

To manage the envs, I use `virtualenvwrapper` - just some shell scripts around the base `virtualenv` (there's also [`venvwrapper`](https://github.com/antebandov/venvwrapper)) - this allows me to activate the virtualenv by typing e.g `workon wcivf`.

INSTALL.md
symroe: This is strange...is `python` found at all? I'd have thought that `pyenv` would have put both `python` and `python3` on your PATH. What does `python --version` show?

michaeljcollinsuk: I think this may have been something to do with the pyenv installation. Now that I have commited the `.python-version` file hopefully this should help with this. Whenever there is a `.python-version` file present in a directory, Pyenv will attempt to activate and use that version. If the specified is not installed, you'd get an error when running `python` command like:

```
michaelcollins@Michaels-MBP WhoCanIVoteFor % python -V                          
pyenv: version `3.6.1' is not installed (set by /Users/michaelcollins/code/dc/WhoCanIVoteFor/.python-version)
```

More info here https://realpython.com/intro-to-pyenv/

INSTALL.md
michaeljcollinsuk: I'm gonna explore these myself and then think it could be a good thing to include in the tech handbook

INSTALL.md
michaeljcollinsuk: FWIW here is a step by step shell output of me 
- opening a new terminal session 
- displaying the default python version as system 
- then changing to the WhoCanIVoteFor dir and displaying the 3.6.12 version being used
- then creating the venv, activating it and again checking the version

Might be useful to reference against steps you are taking
```
michaelcollins@Michaels-MBP ~ % pwd
/Users/michaelcollins
michaelcollins@Michaels-MBP ~ % python -V
Python 2.7.16
michaelcollins@Michaels-MBP ~ % cd code/dc/WhoCanIVoteFor 
michaelcollins@Michaels-MBP WhoCanIVoteFor % python -V
Python 3.6.12
michaelcollins@Michaels-MBP WhoCanIVoteFor % pyenv versions
  system
* 3.6.12 (set by /Users/michaelcollins/code/dc/WhoCanIVoteFor/.python-version)
  3.7.9
  3.8.6
  3.9.0
michaelcollins@Michaels-MBP WhoCanIVoteFor % python -m venv env
michaelcollins@Michaels-MBP WhoCanIVoteFor % source env/bin/activate
(env) michaelcollins@Michaels-MBP WhoCanIVoteFor % python -V
Python 3.6.12
(env) michaelcollins@Michaels-MBP WhoCanIVoteFor % 
```

VirginiaDooley: Very helpful, thank you! These steps worked for me. 👍

INSTALL.md
```diff
@@ -1,11 +1,61 @@
...
+
+Check that the virtual environment is using the correct python version:
+
+    python --version
+
+The output should match the python version set in the `.python-version`. If it does not, something has gone wrong - delete the `env` directory and retry the steps to set the python version and create the virtual environment.
+
+You can now install the project dependencies in to your activated virtualenv:
+
+    pip install -r requirements/local.txt
```
symroe, line 53: Not in the existing install, but I think we should document running `pytest` here - this will make sure the install has worked, but also make sure that the tests can be run right from the start.

symroe (approved):
This looks good - as long as it works for @VirginiaDooley as well!

PR #619, opened 2021-01-08
Hotfix/empty postcode error

### LINKED ISSUE #316: Show a nicer error for blank postcode query
e.g: https://whocanivotefor.co.uk/?postcode=

We get a fair chunk of `Reverse for 'postcode_view' with keyword arguments '{'postcode': ''}' not found. 1 pattern(s) tried: ['elections/(?P<postcode>[^/]+)/$']` in the logs

Not sure if this is people using browsers that don't support the HTML5 `required` attribute or 3rd parties submitting to us without validation, but we could do something better here

michaeljcollinsuk:
Resolves #316 

Checks that a value is actually present in the request.GET rather than simply the key being present. If not, page is rendered as normal, without 500 error

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ var/
 *.egg-info/
 .installed.cfg
 *.egg
+.vscode/
 
 # PyInstaller
 #  Usually these files are written by a python script from a template
--- a/requirements/local.txt
+++ b/requirements/local.txt
@@ -1,5 +1,4 @@
 -r testing.txt
 
 ipdb
-
-
+pre-commit
--- a/wcivf/apps/core/tests/test_views.py
+++ b/wcivf/apps/core/tests/test_views.py
@@ -1,0 +1,31 @@
+from django.test import TestCase
+from django.shortcuts import reverse
+
+
+class TestPostcodeFormView(TestCase):
+    def test_no_postcode(self):
+        """
+        When postcode is not present in query string no redirect occurs.
+        """
+        response = self.client.get("/?postcode=")
+        assert response.status_code == 200
+
+    def test_has_postcode(self):
+        """
+        When poscode is present in request GET redirect to the postcode view
+        occurs.
+        """
+        response = self.client.get("/?postcode=TE11ST")
+
+        assert response.status_code == 302
+        assert response.url == reverse(
+            "postcode_view", kwargs={"postcode": "TE11ST"}
+        )
+
+    def test_has_invalid_postcode(self):
+        """
+        When invalid_postcode is in the request GET redirect does not occur.
+        """
+        response = self.client.get("/?postcode=TE11ST&invalid_postcode=1")
+
+        assert response.status_code == 200
--- a/wcivf/apps/core/views.py
+++ b/wcivf/apps/core/views.py
@@ -18,7 +18,7 @@ class PostcodeFormView(FormView):
 
     def get(self, request, *args, **kwargs):
         if (
-            "postcode" in request.GET
+            request.GET.get("postcode")
             and "invalid_postcode" not in self.request.GET
         ):
             redirect_url = reverse(


michaeljcollinsuk:
One thing that came up in the process of this minor change was that it seems that there are places in the code that makes assumptions that Redis is running which could be refactored - I've added an issue for now as its not particularly pressing https://github.com/DemocracyClub/WhoCanIVoteFor/issues/620

.gitignore
```diff
@@ -23,6 +23,7 @@ var/
 *.egg-info/
 .installed.cfg
 *.egg
+.vscode/
```
symroe, line 26: As per my comment about `.python-version`, but the opposite I guess - is this a useful config file for others using this project and vs code, or should we be excluding them?

symroe:
The actual code looks great :+1: 

As for the PR - it would be good to define some style guides for making PRs - I like the `hotfix` / GitFlow naming, but it would be nice to understand the systems you've used in the past and write down a guide for them.

We might be able to fork the [mySociety one](https://mysociety.github.io/coding-standards.html) and adapt to our needs.

For example, I'd like:

1. The commit message in ee51f51 to meet the 72 char title / 72 wrapped body standard (if nothing else, to prevent GitHub from wrapping the text)
2. The tests to be included in the same commit as the code change (to make `git bisect` work better)
3. The black commit to be squashed in to the other one - by the way, I use a pre-commit hook to stop myself from committing non-black standard code, else I always get caught out by CI fails!

In general for PRs I prefer `!fixup` commits and rebasing / force pushing to get a clean commit history. Happy to talk this through in more detail on Monday.

michaeljcollinsuk:
> The actual code looks great 👍
> 
> As for the PR - it would be good to define some style guides for making PRs - I like the `hotfix` / GitFlow naming, but it would be nice to understand the systems you've used in the past and write down a guide for them.
> 
> We might be able to fork the [mySociety one](https://mysociety.github.io/coding-standards.html) and adapt to our needs.
> 
> For example, I'd like:
> 
> 1. The commit message in [ee51f51](https://github.com/DemocracyClub/WhoCanIVoteFor/commit/ee51f514dbf58e10beee52b7686283b5684b0c29) to meet the 72 char title / 72 wrapped body standard (if nothing else, to prevent GitHub from wrapping the text)
> 2. The tests to be included in the same commit as the code change (to make `git bisect` work better)
> 3. The black commit to be squashed in to the other one - by the way, I use a pre-commit hook to stop myself from committing non-black standard code, else I always get caught out by CI fails!
> 
> In general for PRs I prefer `!fixup` commits and rebasing / force pushing to get a clean commit history. Happy to talk this through in more detail on Monday.

Totally agree about defining some styles for PRs - I think git best practices in general is probably a bit of a blank spot in my knowledge. So would be good to discuss this next week.

.gitignore
GeoWill: If it is useful, might it be better to use a [.editorconfig](https://editorconfig.org/) (Eg [Django one](https://github.com/django/django/blob/master/.editorconfig), so it is editor agnostic?

michaeljcollinsuk: I'm not sure that it would be right to commit this as although I am sure there are some settings that are useful for everyone (e.g.enabling linting) there is probably others that could fall under personal preferences?

GeoWill: I don't know if it does, but it would be neat if `.editorconfig` extended and overrides `.editor-specific-config`. Then we could check in the basics and any additional personal preferences would be taken care of in the local repo - with the local editor specific config file added to `.git/info/exclude` (or `.gitignore` if preferred - though again if I'd argue that's more for per project things (e.g. build artefacts) than individual preferences/tooling (e.g. `.idea/`)). Though it's probably easier to just say follow PEP8, run black, etc and let everyone take care of it however they say best - with CI saying 'nope' if it hasn't been done.

symroe: How do `.editorconfig` and `pyproject.toml` interact?! Some projects like `black` use`pyproject.toml` for settings, but I don't know if that's standard or not?

This is all quite confusing. I suggest we invent our own file format to settle this once and for all :)

symroe (approved):

PR #623, opened 2021-01-13
Hotfix/future elections

### LINKED ISSUE #499: Always set future elections as current
Future (not current) elections are referred to in the past tense, e.g. https://whocanivotefor.co.uk/person/1478/shaun-bailey

image

image

michaeljcollinsuk:
Closes #499

### DIFF
--- a/wcivf/apps/elections/models.py
+++ b/wcivf/apps/elections/models.py
@@ -49,8 +49,36 @@ class Election(models.Model):
     def __str__(self):
         return self.name
 
+    @property
     def in_past(self):
+        """
+        Returns a boolean for whether the election date is in the past
+        """
         return self.election_date < datetime.date.today()
+
+    @property
+    def is_city_of_london(self):
+        """
+        Returns boolean for if the election is within City of London district
+        """
+        return "local.city-of-london" in self.slug
+
+    @property
+    def polls_close(self):
+        """
+        Return the time the polls close
+        """
+        if self.is_city_of_london:
+            return datetime.time(20, 0)
+
+        return datetime.time(22, 0)
+
+    @property
+    def is_election_day(self):
+        """
+        Return boolean for whether it is election day
+        """
+        return self.election_date == datetime.date.today()
 
     def friendly_day(self):
         delta = self.election_date - datetime.date.today()
@@ -272,7 +300,7 @@ class PostElection(models.Model):
                     url, title
                 )
         if not message:
-            if self.election.in_past():
+            if self.election.in_past:
                 message = "(The poll for this election was cancelled)"
             else:
                 message = "<strong>(The poll for this election has been cancelled)</strong>"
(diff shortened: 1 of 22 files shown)


wcivf/apps/elections/models.py
```diff
@@ -49,9 +49,37 @@ class Meta:
...
+        """
+        return "local.city-of-london" in self.slug
+
+    @property
+    def polls_close(self):
+        """
+        Return the time the polls close
+        """
+        if self.is_city_of_london:
+            return datetime.time(20, 0)
```
symroe, line 72: Would it be possible to add a comment, maybe with a link to a page on the City of London website citing this? @pmk01 do you know of a good page?

wcivf/apps/elections/models.py
```diff
@@ -49,9 +49,37 @@ class Meta:
...
     def in_past(self):
+        """
+        Returns a boolean for whether the election date is in the past
+        """
         return self.election_date < datetime.date.today()
 
+    @property
+    def is_city_of_london(self):
+        """
+        Returns boolean for if the election is within City of London district
```
symroe, line 62: Can you expand this comment a little to explain _why_ we special case the city? Just "The city often has different rules to other UK elections so it's useful to know when we need to special case" or something along those lines.

symroe (commented):
I really like how this is looking now. I've left a few comments that I admit are fairly nit-picky, but I figured it's worth pointing out a few things in the interest of improving the overall quality of this code base!

One other thing that might be worth looking at is a couple of checks of the actual wording on templates. A few tests along the line of:

```py
req = self.client.get(election.get_absolute_url())
self.assertContains(req, "Polls are open from 8am")
```

would go a long way to making sure the templates are doing what we expect. I think this could just be dropped in to your existing test set ups if that made sense, or in some new tests for the views/templates - up to you.

wcivf/apps/elections/models.py
pmk01: This? https://www.cityoflondon.gov.uk/about-us/voting-elections/elections/ward-elections

michaeljcollinsuk: The only place I could find where the actual opening/closing times were explicitly mentioned as being different was here https://www.citymatters.london/city-london-voters-may-go-polls-twice-2021/ there's a paragraph
 which states:

"This is because City elections open at 8am and close at 8pm, whilst GLA polling runs from 7am  at 10pm, so it would be impractical to share the same venue. An Act of Parliament dating back to 1957 would have to be changed to change the hours in the Square Mile."

michaeljcollinsuk: I'll link to both Peter's link and this old blog post of ours that references it https://democracyclub.org.uk/blog/2017/03/22/eight-weird-things-about-tomorrows-city-london-elections/

Issue #572, opened 2020-10-19
Use twitter:card=summary rather than summary_large_image


symroe:
At the moment, we use `twitter:card=summary_large_image`. This looks ugly when it's just the logo, so we should use `twitter:card=summary`

PR #627, opened 2021-01-19
Update twitter card size 

VirginiaDooley:
https://github.com/DemocracyClub/WhoCanIVoteFor/issues/572

Update twitter card for greater flexibility when no image exists in the most recent tweet.


symroe (approved):
:tada: 

By the way, the coverage drop is something to do with coveralls - it seems to change the coverage a small amount each time it runs. Not sure why.

PR #630, opened 2021-01-20
Hotfix/delete deleted elections

### LINKED ISSUE #456: Handle deleted elections
`import_elections` / `import_posts` should handle the case where an election is created in error and then soft-deleted in EE

michaeljcollinsuk:
Closes #456 

Implementation notes/queries:
- Wasn't sure if the delete needs to replicate the checks made in YNR `safe_delete` method [here](https://github.com/DemocracyClub/yournextrepresentative/blob/bd0db75e46ef27efb2d2f459087f26f5e03cba21/ynr/apps/elections/models.py#L226) so for now have gone for a more straightforward MVP and I can update to add this check if necessary. This is also ensures that any associated PostElection objects are deleted as it is using `CASCADE` on the FK on PostElection.
- I've added the call to actually make the delete directly to the management command as this seemed most clear and explicit. Although there is an argument to be made for having it run on the `YNRBallotImporter` itself somewhere (e.g. `do_import` or `add_ballots`) so interested to hear any thoughts on this

Some notes about the tests:
- I have used mocks as it allowed me more control of what data to test against. I also had a look at using `vcrpy` since that is already used in the project, but I think this would have meant in my `test_deleted_election_ids` I'd have to parse the created fixtures in the same way as I do in the method itself, so it didn't feel right to me. But interested to hear any feedback about this.
- Since I was already using mocks, I also used these for my `test_deleted_deleted_elections` rather than using factories to create objects and then allowing the tests to hit the DB. However this does make them somewhat more abstract, so they could be amended if necessary

### DIFF
--- a/wcivf/apps/elections/helpers.py
+++ b/wcivf/apps/elections/helpers.py
@@ -53,7 +53,6 @@ class EEHelper:
 
         elections_count, _ = Election.objects.filter(
             slug__in=self.deleted_election_ids,
-            current=False,
         ).delete()
 
         post_elections_count, _ = PostElection.objects.filter(
--- a/wcivf/apps/elections/helpers.py
+++ b/wcivf/apps/elections/helpers.py
@@ -1,7 +1,11 @@
 import sys
 from functools import update_wrapper
 
+from django.db import transaction
 from django.conf import settings
+from django.utils.functional import cached_property
+from django.utils.timezone import datetime, timedelta
+from django.utils.http import urlencode
 from uk_election_timetables.election_ids import from_election_id
 from uk_election_timetables.calendars import Country
 
@@ -12,10 +16,56 @@ class EEHelper:
 
     ee_cache = {}
 
+    @property
+    def base_elections_url(self):
+        """
+        Builds the URL for the elections endpoint of EveryElection API
+        """
+        return f"{settings.EE_BASE}/api/elections/"
+
+    @cached_property
+    def deleted_election_ids(self):
+        """
+        Returns a generator object that retreives ids for elections that have
+        been deleted in Every Election. Accepts optional date arg, otherwise
+        defaults to objects with a Poll Open date within the last fifty days
+        """
+        params = {
+            "deleted": 1,
+            "poll_open_date__gte": datetime.now().date() - timedelta(days=50),
+        }
+        querystring = urlencode(params)
+        url = f"{self.base_elections_url}?{querystring}"
+        pages = JsonPaginator(page1=url, stdout=sys.stdout)
+        for page in pages:
+            return [result["election_id"] for result in page["results"]]
+
+    @transaction.atomic
+    def delete_deleted_elections(self):
+        """
+        Deletes Election and PostElection objects that were soft-deleted in
+        EveryElection
+        """
+        from elections.models import (
+            Election,
+            PostElection,
+        )  # noqa avoid circular import
+
+        elections_count, _ = Election.objects.filter(
+            slug__in=self.deleted_election_ids,
+            current=False,
+        ).delete()
+
+        post_elections_count, _ = PostElection.objects.filter(
+            ballot_paper_id__in=self.deleted_election_ids,
+        ).delete()
+
+        return elections_count, post_elections_count
+
     def prewarm_cache(self, current=False):
-        page1 = "{}/api/elections/".format(settings.EE_BASE)
+        page1 = self.base_elections_url
         if current:
-            page1 = "{}?current=True".format(page1)
+            page1 = f"{page1}?current=True"
         pages = JsonPaginator(page1, sys.stdout)
         for page in pages:
             for result in page["results"]:
@@ -24,9 +74,7 @@ class EEHelper:
     def get_data(self, election_id):
         if election_id in self.ee_cache:
             return self.ee_cache[election_id]
-        req = requests.get(
-            "{}/api/elections/{}/".format(settings.EE_BASE, election_id)
-        )
+        req = requests.get(f"{self.base_elections_url}{election_id}/")
         if req.status_code == 200:
             self.ee_cache[election_id] = req.json()
             return self.ee_cache[election_id]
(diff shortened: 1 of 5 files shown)


wcivf/apps/elections/helpers.py
```diff
@@ -12,10 +13,40 @@ class EEHelper:
...
+        """
+        return f"{settings.EE_BASE}/api/elections/"
+
+    @cached_property
+    def deleted_election_ids(self):
+        """
+        Returns a generator object that retreives ids for elections that have
+        been deleted
+        """
+        url = f"{self.base_elections_url}?deleted=1"
```
symroe, line 29: Because we intend to run this regularly, and over time the list will get longer, it might be a good idea to also add a data filter here too. We allow filtering on [`poll_open_date`](https://github.com/DemocracyClub/EveryElection/blob/master/every_election/apps/api/views.py#L48), so adding `now() - timedelta(days=50)` (50 has no special reasoning there!) should prevent this taking longer to run over time.

wcivf/apps/elections/helpers.py
```diff
@@ -12,10 +13,40 @@ class EEHelper:
...
+            for result in page["results"]:
+                yield result["election_id"]
+
+    def delete_deleted_elections(self):
+        """
+        Deletes elections that were soft-deleted in EveryElection
+        """
+        from elections.models import Election  # noqa avoid circular import
+
+        return Election.objects.filter(
```
symroe, line 41: `self.deleted_election_ids` will also include ballots (`group_type=None`), and we do sometimes delete a single ballot. In fact, this might be the most common path: for example when creating the IDs, we normally get the list of organisations right, but might make a mistake when adding by selecting "ward A" rather than "ward B" (this normally expresses itself as "Foo North ward" rather than "Foo North West Ward").

Relying on the CASCADE to delete the elections is smart when we've created the wrong _organisation_ group level election (`local.foo.2021-05-06`), but this doesn't address the ballot level soft deletes (`local.foo.foo-north.2021-05-06`).

I think adding the same query but for `PostElection` would deal with this.

wcivf/apps/elections/tests/test_helpers.py
```diff
@@ -31,3 +35,56 @@ def test_with_territory_code_malformed_id(self):
...
 
         assert expected is None
+
+
+class TestEEHelper:
+    @pytest.fixture
+    def ee_helper(self):
+        return EEHelper()
+
+    def test_base_elections_url(self, ee_helper):
```
symroe, line 45: It's possible for this to fail in some situations - for example, I have changed `settings.EE_BASE` in my `local.py` in the past when testing the relationship between local EE and local WCIVF. In that case, `            ee_helper.base_elections_url` would return `http://localhost:8001/api/elections/`. 

I expect it's worth adding a [settings override](https://docs.djangoproject.com/en/2.2/topics/testing/tools/#django.test.override_settings) here just in case.

wcivf/apps/elections/management/commands/import_ballots.py
```diff
@@ -39,6 +39,15 @@ def populate_any_non_by_elections_field(self):
             election.any_non_by_elections = any_non_by_elections
             election.save()
 
+    def delete_deleted_elections(self):
```
symroe, line 42: I think this should be decorated with a `@transaction.atomic`, just in case something goes wrong.

symroe:
> Wasn't sure if the delete needs to replicate the checks made in YNR safe_delete method

I'd say no - in general we can be less careful with this site as it's not the source of any data. If we delete something by mistake here, it's going to be back again in a future import. With YNR it's (too) easy to delete the canonical data in a way that's hard to recover.

> I've added the call to actually make the delete directly to the management command

I think this is reasonable in this case, but in general I like to have the management commands quite small, with the logic living somewhere else. I find this makes things easier to test, but also more reusable if other parts of the code want to share logic

> I have used mocks as it allowed me more control of what data to test against. I also had a look at using vcrpy

Yes, this is good. I don't think VCR is right for us, as we normally want to write tests with very specific sets of data. VCR makes it hard to create or edit this data. This is how @GeoWill and I started talking about generated test fixtures, and that lead to the PR he's working on over on the polling stations repo.

FWIW, I think these mocked tests are much nicer than VCR, and I think we should try to migrate VCR tests to mock (or some other system) as we go.

wcivf/apps/elections/helpers.py
michaeljcollinsuk: I've updated to attempt to delete PostElection objects also. But in the process I was trying to figure out how/where `group_type` is set, and noticed there is an `identifier_type` [field](https://github.com/DemocracyClub/EveryElection/blob/5145c0bd6138555b65cd1c8c0f1f4193f9890b9a/every_election/apps/elections/models.py#L300) that is exposed over the API. So I wonder if rather than use whole list of `election_id` values that come back, would it be better to filter them by their identifier type before making the filter/delete?

wcivf/apps/elections/helpers.py
```diff
@@ -12,10 +16,56 @@ class EEHelper:
...
+        EveryElection
+        """
+        from elections.models import (
+            Election,
+            PostElection,
+        )  # noqa avoid circular import
+
+        elections_count, _ = Election.objects.filter(
+            slug__in=self.deleted_election_ids,
+            current=False,
```
symroe, line 56: Can you explain this line? I don't _think_ we want to exclude `current` elections here...

michaeljcollinsuk: @symroe it was to match how YNR handles deleting elections https://github.com/DemocracyClub/yournextrepresentative/blob/bd0db75e46ef27efb2d2f459087f26f5e03cba21/ynr/apps/elections/models.py#L221

But if that's not important here I won't

michaeljcollinsuk: In fact I think we've already covered this in previous comments - so I'm going to update to remove the current check

symroe: Oh that is interesting about EE->YNR imports - I'm not sure it's right in call cases! Either way, we care less here as none of the data is canonical in WCIVF.

symroe (approved):

Issue #596, opened 2020-11-17
The 0 candidates


pmk01:
![Zero](https://user-images.githubusercontent.com/61510609/99378165-06a0d400-28bf-11eb-8522-aed27d7c88bf.png)

The preview text on WCIVF tabs reads '0 candidates' regardless of the number of candidates on page.

symroe:
* [ ] Use @michaeljcollinsuk's new `current_or_future` queryset in the template - but consider what happens when we don't have any curent candidacies
* [ ] write a test for this

PR #632, opened 2021-01-20
Hotfix/Conditional to display number of candidates in meta title

VirginiaDooley:
Closes https://github.com/DemocracyClub/WhoCanIVoteFor/issues/596

In the context of the meta title, `person_posts` was undefined and therefore defaulted to 0 regardless of the number of candidates actually standing for a post.

### DIFF
--- a/wcivf/apps/elections/templates/elections/includes/_post_meta_description.html
+++ b/wcivf/apps/elections/templates/elections/includes/_post_meta_description.html
@@ -1,1 +1,7 @@
-{% autoescape off %}See all {{ person_posts|length }} candidates in {{ object.post.label }} in the {{ object.election.name }} on {{ object.election.election_date }}: {% for pp in person_posts %}{{ pp.person.name|safe }} ({{ pp.party.party_name }}) {% endfor %}{% endautoescape %}
+{% if postelection.people|length %}
+{% autoescape off %}See all {{ postelection.people|length }} candidates in the {{ object.election.name }} on {{
+object.election.election_date }}: {% for pp in person_posts %}{{ pp.person.name|safe }}
+({{ pp.party.party_name }}) {% endfor %}{% endautoescape %}
+{% else %}
+{{ object.election.name }}: No candidates known yet.
+{% endif %}
--- a/wcivf/apps/elections/templates/elections/includes/_post_meta_title.html
+++ b/wcivf/apps/elections/templates/elections/includes/_post_meta_title.html
@@ -1,1 +1,5 @@
-{{ object.election.name }}: The {{ person_posts|length }} candidates in {{ object.post.label }}
+{% if postelection.people|length %}
+    {{ object.election.name }}: The {{ postelection.people|length }} candidates in {{ object.post.label }}
+{% else %}
+    {{ object.election.name }}: No candidates known yet.
+{% endif %}
--- a/wcivf/apps/elections/tests/test_election_views.py
+++ b/wcivf/apps/elections/tests/test_election_views.py
@@ -1,14 +1,14 @@
 import pytest
-
+from django.shortcuts import reverse
 from django.test import TestCase
 from django.test.utils import override_settings
-from django.shortcuts import reverse
+from elections.models import PostElection
 from elections.tests.factories import (
     ElectionFactory,
-    PostFactory,
     PostElectionFactory,
+    PostFactory,
 )
-from elections.models import PostElection
+from people.tests.factories import PersonFactory, PersonPostFactory
 
 
 @override_settings(
@@ -73,3 +73,56 @@ class ElectionViewTests(TestCase):
                 )
                 self.assertContains(response, election[0].nice_election_name)
                 self.assertContains(response, election[1])
+
+
+class ElectionPostViewTests(TestCase):
+    def setUp(self):
+        self.election = ElectionFactory(
+            name="Adur local election",
+            election_date="2021-05-06",
+            slug="local.adur.churchill.2021-05-06",
+        )
+        self.post = PostFactory(label="Adur local election")
+        self.post_election = PostElectionFactory(
+            election=self.election, post=self.post
+        )
+
+    def test_zero_candidates(self):
+        response = self.client.get(
+            self.post_election.get_absolute_url(), follow=True
+        )
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, "elections/post_view.html")
+        self.assertTemplateUsed(
+            response, "elections/includes/_post_meta_title.html"
+        )
+        self.assertTemplateUsed(
+            response, "elections/includes/_post_meta_description.html"
+        )
+        self.assertContains(response, "No candidates known yet.")
+
+    def test_num_candidates(self):
+        people = [PersonFactory() for p in range(5)]
+        for person in people:
+            PersonPostFactory(
+                post_election=self.post_election,
+                election=self.election,
+                post=self.post,
+                person=person,
+            )
+
+        response = self.client.get(
+            self.post_election.get_absolute_url(), follow=True
+        )
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, "elections/post_view.html")
+        self.assertTemplateUsed(
+            response, "elections/includes/_post_meta_title.html"
+        )
+        self.assertTemplateUsed(
+            response, "elections/includes/_post_meta_description.html"
+        )
+        self.assertContains(response, f"The 5 candidates in {self.post.label}")
+        self.assertContains(
+            response, f"See all 5 candidates in the {self.post.label}"
+        )


wcivf/apps/elections/templates/elections/includes/_post_meta_title.html
```diff
@@ -1 +1,5 @@
-{{ object.election.name }}: The {{ person_posts|length }} candidates in {{ object.post.label }}
\ No newline at end of file
+{% if object.election.current_or_future_elections %}
+{{ object.election.name }}: The {{ postelection.people|length }} candidates in {{ object.post.label }}
+{% else %}
+{{ object.election.name }}: The official candidate list has not yet been published.
```
symroe, line 4: Sorry to dive in when this is still WIP, but I thought a comment on this might be useful now:

It's not quite true that the states are "have candidates" or "SOPN not published", although they are the most common states we're likely to see.

On WCIVF we import candidates before they've been "verified" (e.g, the SOPN is out, two people have compared our list to the SOPN). This means that we could have a single candidate on the site months before the SOPN. This code deals with that case well.

However, we can also have a state where the SOPN is out but we've not managed to add any candidates for it yet. It can take some days for us to get to a given area, this year especially.

There are a couple of options here:

1. We get clever with another condition (Using https://github.com/DemocracyClub/uk-election-timetables as we do [elsewhere in this project](https://github.com/DemocracyClub/WhoCanIVoteFor/blob/fa251d0910e228791b31309ad3025bd7ff2f2d30/wcivf/apps/elections/helpers.py#L98))
2. We make the above line more generic, something like "no candidates known yet" - maybe ask @pmk01 for some ideas here.

As this is just a page title, I think option 2 is the better one here, but happy for either if you think 1is better.

VirginiaDooley: No problem! With #1, are you suggesting a third condition? In the mean time, I've pushed the change as suggested in #2.

symroe: Yeah, if we were to implement #1, it would need to deal with another condition. You can see how we do that on the ballot card here: https://github.com/DemocracyClub/WhoCanIVoteFor/blob/master/wcivf/apps/elections/templates/elections/includes/_single_ballot.html#L100

I think going down that route is overkill for a `<title>` though, especially as we have it on the page already

wcivf/apps/elections/tests/test_election_views.py
```diff
@@ -39,3 +40,45 @@ def test_election_detail_view(self):
...
+                person=person,
+            )
+
+    def test_zero_candidates(self):
+        response = self.client.get(
+        self.election.get_absolute_url(), follow=True
+        )
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, "elections/election_view.html")
+        # I need to figure out how to test the page_title on the line below
```
VirginiaDooley, line 68: @michaeljcollinsuk I couldn't figure out how to test the `page_title`; A simple response + str hasn't worked here.

michaeljcollinsuk: I think the reason that the assert commented out on line 69 fails is because your test is checking the wrong view - rather than the `ElectionView` it should be the `PostView` that the test needs to hit. As the template rendered by the `ElectionView` doesn't include `_post_meta_title.html `. And also it is in the `PostView` where the `people` variable that is used is set https://github.com/DemocracyClub/WhoCanIVoteFor/blob/fd79f1ad27645c0e806484f3a9ebc8884389b6c9/wcivf/apps/elections/views/election_views.py#L106

So you should be able to change line 63 to:
```
response = self.client.get(
    self.post_election.get_absolute_url(), follow=True
)
```

You will also then need to update the assert on line 67 to check that the `post_view.html` is used.

wcivf/apps/elections/tests/test_election_views.py
```diff
@@ -39,3 +40,45 @@ def test_election_detail_view(self):
...
+        # self.assertContains(response, "No candidates known yet.", html=True)
+
+    def test_num_candidates(self):
+        response = self.client.get(
+        self.election.get_absolute_url(), follow=True
```
michaeljcollinsuk, line 73: Same as above, this should be using `self.post_election.get_absolute_url()`

wcivf/apps/elections/tests/test_election_views.py
```diff
@@ -39,3 +40,45 @@ def test_election_detail_view(self):
...
+class ElectionPostViewTests(TestCase):
+    def setUp(self):
+        self.election = ElectionFactory(
+            name="Adur local election",
+            election_date="2021-05-06",
+            slug="local.adur.churchill.2021-05-06",
+        )
+        self.post = PostFactory()
+        self.post_election = PostElectionFactory(election=self.election, post=self.post)
+        people = [PersonFactory() for p in range(5)]
```
michaeljcollinsuk, line 53: I would move this block (lines 53-60) to within your `test_num_candidates` test. Otherwise your `test_zero_candidates` test will fail, as it will have 5 candidates rather than 0.

wcivf/apps/elections/templates/elections/includes/_post_meta_title.html
```diff
@@ -1 +1,5 @@
-{{ object.election.name }}: The {{ person_posts|length }} candidates in {{ object.post.label }}
\ No newline at end of file
+{% if postelection.people|length %}
+{{ object.election.name }}: The {{ postelection.people|length }} candidates in {{ object.post.label }}
```
michaeljcollinsuk, line 2: When you have an `if` block split across lines like this, indent the text within the block e.g.:

```
{% if postelection.people|length %}
    {{ object.election.name }}: The {{ postelection.people|length }} candidates in {{ object.post.label }}
{% else %}
....
```

wcivf/apps/elections/tests/test_election_views.py
```diff
@@ -39,3 +40,45 @@ def test_election_detail_view(self):
...
+        for person in people:
+                PersonPostFactory(
```
michaeljcollinsuk, line 55: This looks like it is over-indented - should be 4 spaces. Although `black` will reformat this (and any other similar formatting errors like line length) for you. I added some _brief_ notes about running it [here](https://github.com/DemocracyClub/WhoCanIVoteFor/blob/master/INSTALL.md#code-formatting) but also might be worth getting `pre-commit` set up so that it runs black before each commit.

wcivf/apps/elections/tests/test_election_views.py
VirginiaDooley: Thanks for catching this. 👍

wcivf/apps/elections/tests/test_election_views.py
```diff
@@ -49,36 +55,29 @@ def setUp(self):
...
+            election=self.election, post=self.post
+        )
+
+    def test_zero_candidates(self):
+        response = self.client.get(
+            self.post_election.get_absolute_url(), follow=True
+        )
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, "elections/post_view.html")
+        # self.assertContains(response, "No candidates known yet.")
```
VirginiaDooley, line 68: @michaeljcollinsuk I'm not able to get line 68 and 83 to pass. Is there a better way to target the partial? Or do I need render it in the test explicitly?

VirginiaDooley: If I add:
`self.assertTemplateUsed(response, "elections/includes/_post_meta_title.html")`
before line 68, the test passes. The same strategy doesn't work for `test_num_candidates`

VirginiaDooley: Unless perhaps it's a problem with the string 🤔

symroe: This is where I'd drop in to an interactive debugger (`ipdb` to see what's going on in the response. Or you can just use `print` on `response.content` and see what it produces. Don't forget to run `pytest` with `-s` to tell it not to capture the output.

I think the response object also has a `templates` list against it - one to try with `ipdb`.

michaeljcollinsuk: Hmm that is odd as everything looks correct.. if you add in a `import pdb; pdb.set_trace()` and then check `response.content` you should see the HTML that is rendered, including the meta title, what does that look like?

wcivf/apps/elections/tests/test_election_views.py
```diff
@@ -39,3 +45,39 @@ def test_election_detail_view(self):
...
+                post_election=self.post_election,
+                election=self.election,
+                post=self.post,
+                person=person,
+            )
+
+        response = self.client.get(self.post_election.get_absolute_url())
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, "elections/post_view.html")
+        # self.assertContains(response, "The 5 candidates in Adur local election")
```
michaeljcollinsuk, line 83: I think this one may fail because of the "in Adur local election" line. In the template, "Adur" is read from the label field on the post object:
```
candidates in {{ object.post.label }}
```
And looking in the PostFactory is hardcoded to "Copeland". You can update your `PostFactory` definition in setUp to specify "Adur" as the label to use:

michaeljcollinsuk: Or alternatively, build the string you asserting against with the `Post` object that is created:
```
self.assertContains(response, f"The 5 candidates in {self.post.label}")
```

VirginiaDooley: I like this, thanks 👍

wcivf/apps/elections/tests/test_election_views.py
```diff
@@ -47,29 +46,46 @@ def test_election_detail_view(self):
         self.assertTemplateUsed(response, "elections/election_view.html")
         self.assertContains(response, self.election.nice_election_name)
 
-    @pytest.mark.freeze_time("2017-03-23")
-    def test_election_detail_day_of_election(self):
```
michaeljcollinsuk, line 51: Has this tet been deleted accidentally or was there a problem with it?

wcivf/apps/elections/templates/elections/includes/_post_meta_title.html
```diff
@@ -1 +1,5 @@
-{{ object.election.name }}: The {{ person_posts|length }} candidates in {{ object.post.label }}
\ No newline at end of file
+{% if postelection.people|length %}
```
michaeljcollinsuk, line 1: I've noticed these same changes should be applied in the `_post_meta_description.html` include also - might be worth adding in to this PR? https://github.com/DemocracyClub/WhoCanIVoteFor/blob/master/wcivf/apps/elections/templates/elections/includes/_post_meta_description.html#L1

wcivf/apps/elections/tests/test_election_views.py
VirginiaDooley: Accidentally! During a merge conflict. I'll add back.

wcivf/apps/elections/templates/elections/includes/_post_meta_description.html
```diff
@@ -1 +1,7 @@
-{% autoescape off %}See all {{ person_posts|length }} candidates in {{ object.post.label }} in the {{ object.election.name }} on {{ object.election.election_date }}: {% for pp in person_posts %}{{ pp.person.name|safe }} ({{ pp.party.party_name }}) {% endfor %}{% endautoescape %}
\ No newline at end of file
+{% if postelection.people|length %}
+{% autoescape off %}See all {{ postelection.people|length }} candidates in the {{ object.election.name }} on {{
```
michaeljcollinsuk, line 2: Really minor thing but I find it easier to read when these are indented (as done in the `_post_meta_title.html`)

michaeljcollinsuk (approved):

wcivf/apps/elections/templates/elections/includes/_post_meta_description.html
VirginiaDooley: Very strange but on save, vs code would not accept my indentation changes. I'll look into it but didn't want to hold this up any longer than I have already. Nevertheless, appreciate the reminder 👍

Issue #396, opened 2019-05-09
Make party boxes the same height


JoeMitchell:
Currently, parties get a larger 'box' due to their different logo sizes. Green gets a bigger box than Cons, for example. Labour gets a giaaant box. 

![image](https://user-images.githubusercontent.com/8096305/57443690-8e2b7280-7246-11e9-9be2-f64cfe06a521.png)

Probably fairer (and nicer to look at) if these could all be the same height. (Even Change UK, which doesn't have a logo)

JoeMitchell:
Less apparent, but still true when squished

![image](https://user-images.githubusercontent.com/8096305/57443957-0d20ab00-7247-11e9-8dc8-c04fd5ff1adc.png)

PR #637, opened 2021-01-26
Hotfix/Make party cards same height

VirginiaDooley:
Closes [396](https://github.com/DemocracyClub/WhoCanIVoteFor/issues/396)

I decreased the size of this "magic number" but there is a larger task of restructuring this layout with greater flexibility, perhaps replacing outdated float structure with flexbox.

### DIFF
--- a/wcivf/assets/scss/_cards.scss
+++ b/wcivf/assets/scss/_cards.scss
@@ -229,7 +229,7 @@ blockquote {
   text-align:center;
   .party_emblem {
     float: none;
-    width:100px;
+    width:75px;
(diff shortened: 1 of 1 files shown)


michaeljcollinsuk (approved):
LGTM - I wonder if for front end changes dropping in a screenshot would be worthwhile?

VirginiaDooley:
> LGTM - I wonder if for front end changes dropping in a screenshot would be worthwhile?

💥 

Screen Shot 2021-01-26 at 10 58 03 AM

PR #638, opened 2021-01-26
Feature/division types

### LINKED ISSUE #608: Stop saying 'area'
I think this should read:

🗳️ Senedd Cymru elections 🗳️
Blaenau Gwent constituency

(apply change to Scottish parl., London Assembly, etc, etc.

Screenshot 2020-12-22 155831

michaeljcollinsuk:
closes #608 

Screenshots of how pages now look - might be worth @pmk01 taking a look over? Can upload more upon request!:

Screenshot 2021-01-29 at 16 23 58

Screenshot 2021-01-29 at 16 24 11

Screenshot 2021-01-29 at 16 24 36

Screenshot 2021-01-29 at 16 24 47

Screenshot 2021-01-29 at 16 24 54

Screenshot 2021-01-29 at 16 25 01

Screenshot 2021-01-29 at 16 25 34

Screenshot 2021-01-29 at 17 13 19

Screenshot 2021-01-29 at 17 13 29

Screenshot 2021-01-29 at 17 49 23

Screenshot 2021-01-29 at 17 50 46


michaeljcollinsuk:
test for octobox

michaeljcollinsuk:
another test

wcivf/apps/elections/models.py
```diff
@@ -229,6 +254,27 @@ def nice_territory(self):
...
+            choice[0]: choice[1] for choice in self.DIVISION_TYPE_CHOICES
+        }
+        return mapping.get(self.division_type, "")
+
+    @property
+    def division_suffix(self):
+        """
+        Returns last word of the division_description
+        """
+        return self.division_description.split(" ")[-1].lower()
```
symroe, line 272: I can see this causing problems in future - one of the hard things about this system of taking strings from EE and using them elsewhere is it's hard to remember how and where we're actually using them (this applies in many places, not just this PR!).

Relying on a particular format might cause problems in the future - especially as we might change EE without checking here.

I think this means a couple of things:

1. Finding a way to store how strings from EE are used, and writing that in the EE docs, so next time we add something we don't have to guess how to use it

2. Sort of as a guard against the above, I think we should store 3 bits of information: ID, description, suffix, maybe as a named tuple. We can grab the description as you're doing in `division_description, but then change `division_suffix` to return `division_description.suffix` rather than splitting the description.

wcivf/apps/elections/import_helpers.py
```diff
@@ -280,6 +281,21 @@ def set_organisation_type(self, ballot):
...
+        """
+        if ballot.post.division_type and not self.force_update:
+            return
+
+        ee_data = self.ee_helper.get_data(ballot.ballot_paper_id)
+
+        if not ee_data or not ee_data["division"]:
+            return
+
+        ballot.post.division_type = ee_data["division"].get("division_type")
```
symroe, line 296: Will this raise a `ValueError` if the type isn't in `DIVISION_TYPE_CHOICES`? I think it would be useful to do this, to make sure we don't forget to update the mapping when / if we add more types in EE

wcivf/apps/elections/models.py
michaeljcollinsuk: Not sure I completely follow - `division_description` is not something that is being stored from EE (for the reason you outlined - in the issue  `division_subtype` [is mentioned as something that should not be relied upon being present all of the time](https://github.com/DemocracyClub/WhoCanIVoteFor/issues/608#issuecomment-763849830)). 

So all I am storing from EE at the moment is the `division_type` which is a 3 letter code. And then I have hardcoded the descriptions for each code using the descriptions in issue #608. Therefore any change in EE wouldn't affect anything about how we determine the description for a `division_type`. Unless a new `division_type` is added in EE, in which case we would need to update the codebase here.

So as far as `division_suffix` goes I can then be confident that what when I do the `split`, assuming we have a valid `division_type` the format will match the way I expect (e.g. final word being a descriptor like "ward") because it is set here and not in EE. And if we don't have a `division_type` code (e.g. it is a Mayoral Election) then it will always return an empty string rather than error. I was then going to update this to account for those sort of edge cases somehow.

That being said, happy to implement the changes you have suggested, but I think it would mean capturing and storing some extra bits from EE (e.g. the description)? So perhaps best to have a quick run through to check we are on same page.

wcivf/apps/elections/import_helpers.py
michaeljcollinsuk: TIL;

No, it doesn't by default. But by adding a call to `ballot.full_clean()` before the save, then yes a `ValidationError` is raised if the value isn't in `DIVISION_TYPE_CHOICES`. So I think that would help with the above issue, to make sure that if a new division_type is added in EE it is also updated here.

wcivf/apps/elections/models.py
symroe: Sorry, my mistake - as long as we're parsing text that's checked in to this repo I think we're good :+1:

pmk01:
Hurrah! These are great. This will really improve the AMS pages especially.

I have two thoughts, both of which might fall outside of the scope of the PR and neither are urgent:
1. It might be worth removing the second use of 'by-election' (so the page will read in full: "Scottish parliament by-election: Shetland Islands Constituency")
2. I'm still not entirely happy with PCCs. Thinking this over, it might be best to remove the geography reference entirely ("xxxx police force area") in the same way we might treat mayors? (apologies if you've already thought about this... or if I've mentioned it elsewhere and it's on the to do list)

wcivf/apps/elections/models.py
```diff
@@ -303,19 +306,38 @@ def expected_sopn_date(self):
...
+        return self.ballot_paper_id.startswith("mayor")
+
     @property
     def friendly_name(self):
+        """
+        Helper property used in templates to build a 'friendly' name using
+        details from associated Post object, with exceptions for by-elections
+        and mayoral elections
         # TODO Take more info from YNR/EE about the election
         # rather than hard coding not_wards and not_by_elections
```
symroe, line 360: Reckon this TODO is done?

wcivf/apps/elections/import_helpers.py
```diff
@@ -91,11 +91,10 @@ def __init__(self, ee_helper=None):
         self.post_cache = {}
 
     def update_or_create_from_ballot_dict(self, ballot_dict):
-        created = False
         # fall back to slug here as some temp ballots don't have an ID set
-        post_id = ballot_dict["post"].get("id", ballot_dict["post"]["slug"])
-        if not post_id in self.post_cache:
-            post, created = Post.objects.update_or_create(
+        post_id = ballot_dict["post"]["id"] or ballot_dict["post"]["slug"]
```
symroe, line 95: Good spot!

I added this [recently](https://github.com/DemocracyClub/WhoCanIVoteFor/commit/d4541719ad013a21d24532f033e4a0ad33a7a2a5) as a response to some posts getting assigned to the incorrect ballot - this was because `self.post_cache` contained None values for >1 post.

This fix makes the code correct here, but I wanted to check that we didn't have duplicate slugs in YNR for some reason. The good news is that all posts without an identifier have a unique slug there, so this change should work fine :+1:

michaeljcollinsuk:
> Hurrah! These are great. This will really improve the AMS pages especially.
> 
> I have two thoughts, both of which might fall outside of the scope of the PR and neither are urgent:
> 
> 1. It might be worth removing the second use of 'by-election' (so the page will read in full: "Scottish parliament by-election: Shetland Islands Constituency")
> 2. I'm still not entirely happy with PCCs. Thinking this over, it might be best to remove the geography reference entirely ("xxxx police force area") in the same way we might treat mayors? (apologies if you've already thought about this... or if I've mentioned it elsewhere and it's on the to do list)

Point 1 - agreed (actually makes the code a bit more straightforward too!)

Regarding point 2 - so do you mean simply remove the 'subheading' line. So for example:

Screenshot 2021-02-01 at 17 50 58

wcivf/apps/elections/import_helpers.py
michaeljcollinsuk: Would it make sense to update the `ynr_id` to remove `blank=True` and add `unique=True`? It is already flagged as the primary key

michaeljcollinsuk:
@symroe this has (finally) been rebased - could I get a final check over when you have chance? I've also confirmed with Peter that the latest screenshots are good. Other than that, I don't believe anything functional will have changed from your last check

VirginiaDooley:
Nice rebase and super clear commit messages - makes it much easier to understand what's happening in the code 🙌

wcivf/apps/elections/import_helpers.py
symroe: Removing `black=True` is a good idea - I'm not sure why it's there!

When a field is marked as the [primary key it automatically gets](https://docs.djangoproject.com/en/3.1/ref/models/fields/#django.db.models.Field.primary_key) `unique=True` and `db_index=True`, so no need to add them as well.

symroe (approved):
Yep, this is great - happy for you to merge this now and leave the `Post.ynr_id` comment for another time - I don't know if that will end up breaking other things that we don't want to deal with as part of this PR!

Issue #274, opened 2018-07-16
ICS calendar validation errors on phone


JoeMitchell:
Feedback from WCIVF last May:

>Earlier I told you I was pleased to find an ICS calendar showing local elections. That remains true, and it works in one calendar program, but not the one on my phone. Apparently, there are validation errors: https://icalendar.org/validator.html?url=http://whocanivotefor.co.uk/elections/M24%201NY.ics - hopefully some simple changes would allow reusing the data more widely. Thanks!

VirginiaDooley:
For an update on this issue, please see the https://github.com/DemocracyClub/WhoCanIVoteFor/pull/641

This issue is on hold [until VTIMEZONE components are built in ](https://github.com/collective/icalendar/issues/44#issuecomment-773466893
)

VirginiaDooley:
Update from user: _The iCal feed link generated a URL that was not accepted by Google Calendar. I discovered that the issue was that a space was being added in the middle of my postcode, which made it an invalid URL. Once I removed the space, I was able to add it successfully._

VirginiaDooley:
Increased support for time zones https://pypi.org/project/python-dateutil/2.8.2/

PR #641, opened 2021-01-28
ICS calendar validation errors on phone

VirginiaDooley:
Ref [#274](https://github.com/DemocracyClub/WhoCanIVoteFor/issues/274)

The source code was missing a DTSTAMP in each VEVENT as well as time zone in the start and end date. 

With these changes, we've solved DSTAMP errors, but still need to address the setting of UTC and the 'floating time' issue. After experimenting with passing in UTC to the start and end time, we satisfied errors, but introduced a potential bug around time zones and rendering the correct time in calendars. I think what we want is for events that appear in BST to appear in the calendar without the +1 time difference. This may require the introduction of time zone settings. However, doing so adds complexity that may be avoided by using a different library or by trying to solve the issue in the iCalendar library.

After careful consideration, we've decided to only solve the DTSTAMP error here and add to the issue in iCalendar that relates to the time zone errors in hopes of a solution we can implement at a future date. 

ICS Cal research [here](https://docs.google.com/document/d/1KCQKzwtDPGZA_ywlFNjaocP7GOg855Sojv0NKP9TK2Y/edit#)

In collab with @michaeljcollinsuk

### DIFF
--- a/wcivf/apps/elections/views/postcode_view.py
+++ b/wcivf/apps/elections/views/postcode_view.py
@@ -1,5 +1,5 @@
 from icalendar import Calendar, Event, vText
-
+from django.utils import timezone
 from django.conf import settings
 from django.http import HttpResponse
 from django.views.generic import TemplateView, View
@@ -86,7 +86,7 @@ class PostcodeiCalView(
             event["summary"] = "{} - {}".format(
                 post_election.election.name, post_election.post.label
             )
-
+            event.add("dtstamp", timezone.now())
             event.add("dtstart", post_election.election.start_time)
             event.add("dtend", post_election.election.end_time)
             event.add(


wcivf/apps/elections/views/postcode_view.py
```diff
@@ -87,6 +88,8 @@ def get(self, request, *args, **kwargs):
                 post_election.election.name, post_election.post.label
             )
 
+            # event.add("dtstamp", datetime.isocalendar())
+            event.add("dtstamp", datetime.now().year)
             event.add("dtstart", post_election.election.start_time)
```
VirginiaDooley, line 93: I am unclear how these params related to django params for "X-WR-TIMEZONE" which we have already defined. Perhaps DTSTART and DTEND need to concatenate TZID or "X-WR-TIMEZONE" to satisfy the TZID errors.

wcivf/apps/elections/views/postcode_view.py
```diff
@@ -1,16 +1,17 @@
-from icalendar import Calendar, Event, vText
```
GeoWill, line 1: I don't think you want to delete this line

wcivf/apps/elections/views/postcode_view.py
```diff
@@ -86,9 +86,9 @@ def get(self, request, *args, **kwargs):
             event["summary"] = "{} - {}".format(
                 post_election.election.name, post_election.post.label
             )
-
-            event.add("dtstart", post_election.election.start_time)
-            event.add("dtend", post_election.election.end_time)
+            event.add("dtstamp", timezone.now()),
```
symroe, line 89: Hard to tell from the diff, but are the trailing `,` meant to be there?

It's valid python to do this, because you can make a tuple inline, e.g `x, y = 1, 2` (I don't think this is great python FYI, but it is valid).

If it's not inside a function call, I think they should be removed.

VirginiaDooley: I don't think they are meant to be there; will remove.

symroe (approved):

PR #645, opened 2021-02-01
Hotfix/Add method to delete orphan posts

### LINKED ISSUE #433: Clean up orphan posts after import_posts
Refs #432 

Because the post for a ballot can change over time (because the IDs change), we can end up with orphan posts from time to time.

This doesn't _really_ matter, as they're not exposed anywhere, but it would be a good idea to clean up after ourselves when running `import_posts`.

VirginiaDooley:
Closes #433 
Co-authored with @michaeljcollinsuk 

This work adds a method and tests to delete orphan posts. We have not added tests on the importer because it falls outside the scope of this ticket, but it could be a nice addition.

### DIFF
--- a/wcivf/apps/elections/import_helpers.py
+++ b/wcivf/apps/elections/import_helpers.py
@@ -170,6 +170,14 @@ class YNRBallotImporter:
             # This is because there's a high chance we've not
             # got all the ballots we need yet.
             self.run_post_ballot_import_tasks()
+        self.delete_orphan_posts()
+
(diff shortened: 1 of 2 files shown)


symroe (approved):
Nice!

symroe:
By the way, GitHub has a nice method for doing commits during pair programming:

https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors

PR #649, opened 2021-02-02
Display correct polls open msg in PostcodeView

### LINKED ISSUE #441: Change opening times on the polling stations card for City of London Alderman/Councilman elections
https://github.com/DemocracyClub/WhoCanIVoteFor/commit/c4edbe9bb2d6237044090aeb55a84191cab1d171 bodges the poll close date in the ballot detail card, but if **all** of the relevant elections matched by a postcode search are City of London local elections, that info should be passed to the polling stations card so we can modify the opening times there too. At the moment we say 8pm on the election detail card but then 10pm on the polling stations card.

Edge case: I'm not entirely sure what happens if there is say a CoL Alderman by-election on the same day as a parliamentary election or something..

This may become irrelevant if we get to https://github.com/DemocracyClub/EveryElection/issues/581 first.

michaeljcollinsuk:
Closes #441 

Updates template to read the poll open/close times for the first object in pre-ordered grouping of elections.

Does not account for edge case where City of London has both a local election and e.g. parliamentary by election on the same day - it will use the opening times of the first election within the list of elections for that day. However this should be ok as @symroe has [mentioned in the issue](https://github.com/DemocracyClub/WhoCanIVoteFor/issues/441#issuecomment-771163882) that this would not occur.

### DIFF
--- a/wcivf/apps/elections/templates/elections/includes/_polling_place.html
+++ b/wcivf/apps/elections/templates/elections/includes/_polling_place.html
@@ -43,11 +43,8 @@
         <a href="{{ polling_station.custom_finder }}">
         following this link</a>.</p>
     {% else %}
-      {% now "j F Y" as current_day %}
-      {% if elections_by_date and elections_by_date.0.grouper|date:"j F Y" == current_day %}
-        {% with today_election=elections_by_date.0.list.0.election %}
-          <strong>Polling stations are open from {{ today_election.polls_open|time:"ga" }} till {{ today_election.polls_close|time:"ga" }} today.</strong>
-        {% endwith %}
+      {% if ballots_today %}
+        <strong>Polling stations are open from {{ ballots_today.0.election.polls_open|time:"ga" }} till {{ ballots_today.0.election.polls_close|time:"ga" }} today.</strong>
       {% endif %}
       {% if polling_station.addresses %}
         <p>Your polling station in {{ postcode }} depends on your address. <a href="https://wheredoivote.co.uk/postcode/{{ postcode }}/">Check the correct polling station for your address &raquo;</a></p>
--- a/wcivf/apps/elections/templates/elections/includes/_polling_place.html
+++ b/wcivf/apps/elections/templates/elections/includes/_polling_place.html
@@ -25,7 +25,6 @@
   You don't need to take your poll card with you{% if not postcode|ni_postcode %}, or any identification{% endif %}.
   {% endif %}
 
-
   {% if polling_station.polling_station_known %}
     <p>
       Your polling station is <strong>{{ polling_station.polling_station.properties.address }}</strong>.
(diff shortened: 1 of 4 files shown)


wcivf/apps/elections/templates/elections/includes/_polling_place.html
```diff
@@ -46,7 +45,9 @@ <h3>{{ voter_id_required.0.1.title }}</h3>
     {% else %}
       {% now "j F Y" as current_day %}
       {% if elections_by_date and elections_by_date.0.grouper|date:"j F Y" == current_day %}
-      <strong>Polling stations are open from 7am till 10pm today.</strong>
+        {% with today_election=elections_by_date.0.list.0.election %}
```
symroe, line 48: Untested, but I think `postelections` still exists in this context, meaning you should be able to do `today_election=postelections.0.election` - this might make it a little easier to understand?

symroe (commented):
I think this is good, accepting that we don't handle a edge case of COL + other election on the same day. 

Not really for this PR, but I wonder if we can find a system of warning about this case that doesn't rely on us remembering that we don't deal with this case? I imaging that we'll have a headache if we have to deal with this anyway, but knowing all the things we need to think about would make it easier to deal with.

chris48s:
>  I wonder if we can find a system of warning about this case that doesn't rely on us remembering that we don't deal with this case?

Not a warning, but one thing you could do as a guard here is check for the case where you've got >= 1 `local.city-of-london.*` and >=1 not `local.city-of-london.*` and just not show the "Polling stations are open .." text in that situation. Then if it does happen the fallback is saying nothing rather than saying something which is possibly wrong. In all other cases, show `elections[0].polls_open` / `elections[0].polls_close` or whatever.

michaeljcollinsuk:
I've updated this to implement the suggestion from @chris48s so that if there is a local city of London election and another election on the same day, no polling station open/closing time information is displayed. And the methods implemented should make it easy to change the messaging in the future if we did get clarity on what would happen in these circumstances

wcivf/apps/elections/views/postcode_view.py
```diff
@@ -80,6 +82,29 @@ def get_todays_ballots(self):
...
 
+    def multiple_city_of_london_elections_today(self):
+        """
+        Checks if there are multiple elections taking place today in the City
+        of London. This is used to determine if it is safe to display polling
+        station open/close times in the template. As if there are multiple then
+        it is unclear what time the polls would be open. See this issue for
+        more info https://github.com/DemocracyClub/WhoCanIVoteFor/issues/441
+        """
+        ballots = self.get_todays_ballots()
```
symroe, line 93: Just an idea on this one - both `get_todays_ballots()` and `ballot_paper_id__startswith="local.city-of-london"` will cause a DB query per request, just to catch an edge case that might never happen.

I think it might be better to change the return value of `get_todays_ballots()` (or maybe allow passing in `flat=True`) to just return a set of ballot paper IDs.

This would mean we only run a single query, but we can also do `len()` rather than `count()` on the set and then use python to work out if there's a city of London ballot in the set.

Another option that would prevent adding another query at all would be to annotate another field like we do here: https://github.com/DemocracyClub/WhoCanIVoteFor/blob/99484efa37e5a2f1b0fe49f60e2983d1956900c1/wcivf/apps/elections/views/mixins.py#L62

This would look like:
```py
pes = pes.annotate(
            is_today=Case(
                When(election__election_date=date.today(), then=1),
                default=0,
                output_field=IntegerField(),
            )
        )
```

`get_todays_ballots` could then be just `return [b for ballot in self.ballots if b.is_today]` and `multiple_city_of_london_elections_today` could just iterate over the list.

michaeljcollinsuk: Yeah you are totally right about the extra DB queries - should have thought more about the performance implications. I liked the `annote` idea, but then remembered I already added the `is_election_day` method on the Election model, so thought I could use that to do the same thing. And then updated `multiple_city_of_london_elections_today` accordingly to work with a list instead of a QS. Tests still pass and think this still results in removing the additional DB queries

symroe: Nice!

symroe (approved):

PR #654, opened 2021-02-08
Adds REDIS_LOG_POSTCODE setting

### LINKED ISSUE #620: Update code to run without Redis locally
It seems there are some assumptions in the code that Redis is running. So for example when attempting to run the project with dummy cache, when submitting a postcode a Redis connection error occurs. Similarly, there a few tests that rely on having Redis running. This could be refactored to allow using the dummy cache

michaeljcollinsuk:
Closes #620

- Allows project to be run without having redis running
- Updates postcode logging method to skip redis and add tests
- Updates existing test to skip if redis isnt running - an alternative to this would be to mock redis in existing test

Suggestions for further areas for testing welcome

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -86,6 +86,12 @@ the cache backend with a file at `./wcivf/settings/local.py` with the following:
         }
     }
 
+You will also need to add:
+
+    REDIS_LOG_POSTCODE = False
+
+This will disable using redis to log postcodes.
+
 
 ## Setting up PostgreSQL and PostGIS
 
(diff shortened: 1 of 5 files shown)


symroe:
This looks good to me - with one question about if the python redis library will install without redis installed? Do we need to try/except the import and make it optional?

michaeljcollinsuk:
> This looks good to me - with one question about if the python redis library will install without redis installed? Do we need to try/except the import and make it optional?

I've tested this on my machine and the python library will still install without redis. And the project will run fine as long as you have the new setting set. So no explicit need for a try/except when importing redis

symroe:
Great!

symroe (approved):

PR #659, opened 2021-02-10
Fix sentry errors

michaeljcollinsuk:
This has been updated to fix a few minor recurring Sentry errors

https://sentry.io/organizations/democracy-club-gp/issues/2131361570/?project=1426221&query=is%3Aunresolved
- Data migration to delete any Post objects with no `ynr_id` value
- Remove `blank=True` from the Post model
- In the ballot importer, if we still get an empty Post ID skip to the next ballot dict

https://sentry.io/organizations/democracy-club-gp/issues/1858067343/?project=1426221&query=is%3Aunresolved&statsPeriod=14d
- If a user attempted to access the iCal view directly with an invalid postcode, a 500 error would be raised
- Updates the `get` method to catch `InvalidPostcodeError` and redirect user to the home page, where invalid postcode message is displayed
- There is further scope to refactor these views and the mixins but it felt outside of the scope of this

https://sentry.io/organizations/democracy-club-gp/issues/1647283944/?project=1426221&query=is%3Aunresolved&statsPeriod=14d
- Update import_local_parties to catch an attribute error when attempting to set a party_id, and skip that row
- A wider refactor of the importers is planned to make them more like those in [EE](https://github.com/DemocracyClub/EveryElection/blob/master/every_election/apps/core/mixins.py#L58) but thought it would be worth to get the quick fix in with this PR


symroe (approved):
Hard to be 100% about the importer commit, but let's merge and run on staging to see what happens

Issue #599, opened 2020-11-23
Candidates standing in more than one election on the same day


pmk01:
candidates with multiple same-day elections only show the latest added election on their profile. 

eg. https://whocanivotefor.co.uk/person/38812/emma-best
(standing in both constituency & list seats)

Issue #404, opened 2019-05-18
Make it clearer if a candidate is one of many standing for a party


chris48s:
It is slightly odd phrasing to frame someone as "**the** party X candidate" when that party is fielding lots of candidates:

![Screenshot at 2019-05-18 22-58-16](https://user-images.githubusercontent.com/6025893/57975412-6f547b00-79c0-11e9-8672-588229589bc0.png)

This applies to party list elections, but also in local elections where a party will field multiple candidates in a multi-seat ward.

symroe:
I reckon this is done? @VirginiaDooley?

Issue #211, opened 2017-06-05
Combine cards into higher-level sections


annapowellsmith:
The person detail pages for GE2017 are starting to get long and bitty (which is good, means there's lots of information on there!).

I would like to combine the cards into higher-level sections as follows.

**{{ candidate }}'s policies**
- Statement
- Party manifesto
- Leaflets
- TWFY record

**About {{ candidate }}**
- Wikipedia
- CV
- Nesta biographical info (#210)
- Companies
- Past candidacies

**{{ candidate }} online**
- Tweets
- Facebook, LinkedIn, homepage, email

annapowellsmith:
Here's how it looks...
![screencapture-localhost-8000-person-1120-jeremy-corbyn-1496817669170](https://user-images.githubusercontent.com/78156/26865506-f7867c86-4b54-11e7-90d1-700467bb1dff.png)

JoeMitchell:
I think this is a good idea! Where would hustings go?

(Also, have we lost 2015 elections again?!)

annapowellsmith:
Cool. Hustings get their own card - `See <name>`

Yes re 2015 elections :( See #137

symroe:
I actually would prefer to keep one card per thing – I find this more confusing to read, like everything is jumbled together, where as the card concept keeps thing self-contained…

annapowellsmith:
OK! I feel like these pages do need some work, but I'm not sure exactly what form it should take - maybe something to discuss with Rose. Moving this out of "in progress", but keeping it open.

Issue #202, opened 2017-06-02
Show statements first if available


annapowellsmith:
If not available, show manifestos first: but if statements are available, we should show them first (and see #198)

PR #660, opened 2021-02-10
Feature/new person page

VirginiaDooley:
This work implements the new [DC design system](ttps://github.com/DemocracyClub/design-system) for the candidate (Person) page as well as addresses the following:
*Related issues:
    * #211
    * #599
    * #202 
    * #198 
    * #404 

To install the new DC design system as a python package and test the changes in this PR, follow [these instructions](https://github.com/DemocracyClub/design-system/pull/2).

### Note on testing: 
Some conditionals in the template require us to create new factories and we weren't sure those sections (such as CV) would be made live just yet. Once the invalid template setting PR is merged, which is dependent on the DC base theme, it will catch conditionals we haven't explicitly tested here. 

### Intro Card for candidate with more than one current candidacy
Screen Shot 2021-03-04 at 9 03 33 PM

### Policies Card with expanded long (>100 words) statement to voters for long statements 
Screen Shot 2021-03-04 at 9 03 48 PM

### Shorter statement to voters
Screen Shot 2021-02-25 at 8 02 38 AM

### About card (includes when present: wikipedia, CV, TWFY link)
Screen Shot 2021-02-24 at 11 02 04 AM

### [Contact card](#contactcard)
Screen Shot 2021-03-04 at 9 05 34 PM

### Previous elections Table
Screen Shot 2021-03-04 at 9 18 50 PM


requirements/base.txt
```diff
@@ -20,7 +20,8 @@ django-debug-toolbar==3.2
 newspaper3k==0.2.8
 
 git+git://github.com/DemocracyClub/dc_base_theme.git@0.3.11
-git+https://github.com/DemocracyClub/dc_signup_form.git@2.1.0
+git+https://github.com/DemocracyClub/dc_signup_form.git@2.1.
```
michaeljcollinsuk, line 23: Think a trailing `0` was removed by mistake here which is causing the CI check to fail

requirements/base.txt
```diff
@@ -21,6 +21,7 @@ newspaper3k==0.2.8
 
 git+git://github.com/DemocracyClub/dc_base_theme.git@0.3.11
 git+https://github.com/DemocracyClub/dc_signup_form.git@2.1.0
+git+https://github.com/DemocracyClub/design-system.git@python-package
```
symroe, line 24: I've merged the PR now, so you can change this line to `git+https://github.com/DemocracyClub/design-system.git`.

This won't pin to a version though - I think that's fine while we're developing both systems side by side, but we'll want to pin it like this before merging:

`git+https://github.com/DemocracyClub/design-system.git@[tag-name]`

wcivf/apps/people/models.py
```diff
@@ -153,6 +154,70 @@ def get_max_facebook_ad_spend(self):
...
+        linkedin_url = self.linkedin_url
+        linkedin_split = list(filter(None, linkedin_url.split("/")))
+        linkedin_username = linkedin_split[-1]
+
+        return linkedin_username
+
+    @property
+    def youtube_username(self):
+        youtube_url = self.youtube_profile
+        if ("channel") in youtube_url:
```
michaeljcollinsuk, line 192: You don't strictly need the parenthesis around `"channel"` when you are checking if a subset exists within a string

wcivf/apps/people/models.py
```diff
@@ -153,6 +154,70 @@ def get_max_facebook_ad_spend(self):
...
+            return self.name + "'s Channel"
+        else:
```
michaeljcollinsuk, line 194: Since you return on line 193 if your conditional is true, you don't need the `else` which I think is more 'pythonic'. Means that you can bring your lines below in one indent

wcivf/apps/people/models.py
```diff
@@ -153,6 +154,70 @@ def get_max_facebook_ad_spend(self):
...
+            youtube_split = list(filter(None, youtube_url.split("/")))
+            youtube_username = youtube_split[-1]
+            return youtube_username
+
+    @property
+    def long_statement(self):
+        if self.statement_count > 10:
```
michaeljcollinsuk, line 201: If you can return the result of your conditional check directly here to save the need for the if/else:
```
return self.statment_count > 10
```

michaeljcollinsuk:
If you wanted a hand writing any unit tests for your model methods, I'd be happy to pair on it with you!

wcivf/apps/people/models.py
VirginiaDooley: Is this what you mean? 

`    @property
    def youtube_username(self):
        youtube_url = self.youtube_profile
        if "channel" in youtube_url:
            return self.name + "'s Channel"
        youtube_split = list(filter(None, youtube_url.split("/")))
        youtube_username = youtube_split[-1]
        return youtube_username`

wcivf/apps/people/models.py
```diff
@@ -154,6 +154,22 @@ def get_max_facebook_ad_spend(self):
             )
         )
 
+    @property
+    def facebook_personal_username(self):
+        facebook_personal_url = self.facebook_personal_url
+        facebook_split = list(filter(None, facebook_personal_url.split("/")))
```
symroe, line 160: With all these properties:

First, I wonder if there's a way to DRY up the methods a little - like pulling the main logic in to one place?


Second, parsing URLs is (or can be) hard - for example, we don't know that someone hadn't added a querystring to the end of a given URL that might look odd. 

Here's an example from the actual data: 

`https://www.facebook.com/sarahmurraylibdem?notif_id=1523864067803194&notif_t=page_invite_accept&ref=notif`. 

I think this code would reduce this down to:

`sarahmurraylibdem?notif_id=1523864067803194&notif_t=page_invite_accept&ref=notif`.

A better option might be to use Python's build in URL parsing:

```py
from urllib.parse import urlparse
parsed_url = urlparse('https://www.facebook.com/sarahmurraylibdem?notif_id=1523864067803194&notif_t=page_invite_accept&ref=notif')
```

Now `parsed_url.path` would be `sarahmurraylibdem` but might also contain more path elements, so we'd want to then do `url = parsed_url.path.split("/")[0].

Of course the _real_ solution here is to get the screen name from the 3rd party site itself, but that's a problem for another day!

wcivf/apps/people/templates/people/not_current_person_detail.html
```diff
@@ -18,6 +18,7 @@ <h2 class="person_name">{{ object.name }}
       (Deceased)
     {% endif %}
     </h2>
+    <p>{{ object.name }} stood for election {{ object.past_not_current_candidacies.count }} times.</p>
```
symroe, line 21: I think this is useful, but we need to think about how to explain that we've only been tracking local elections since 2016 and generals since 2010. (and by-elections since ~2018). In other words, this is actually the number of elections we've tracked this person for. Help from @pmk01 welcome.

wcivf/apps/people/models.py
```diff
@@ -189,12 +189,29 @@ def linkedin_username(self):
...
-            youtube_split = list(filter(None, youtube_url.split("/")))
-            youtube_username = youtube_split[-1]
-            return youtube_username
+        youtube_split = list(filter(None, youtube_url.split("/")))
+        youtube_username = youtube_split[-1]
+        return youtube_username
+
+    @property
+    def long_statement(self):
+        return self.statement_count > 10
```
symroe, line 200: 10 words is a fairly short statement I think - maybe this should be more like 100? Needs to be tested with some real content I expect.

wcivf/apps/people/models.py
```diff
@@ -189,12 +189,29 @@ def linkedin_username(self):
...
+    @property
+    def statement_count(self):
+        statement = self.statement_to_voters
+        statement_count = len(statement.split())
+        return statement_count
+
+    @property
+    def statement_intro(self):
+        statement_to_voters = self.statement_to_voters
+        statement_split = statement_to_voters.splitlines()
```
symroe, line 211: I think we should use the same method to split here as above - but I don't know if splitting on space in both places is better than splitting on newlines.

The reason for this is: Imagine someone has a profile longer than 10 (or whatever) words, but all on one line. This will trigger the "long statement" interface, but if there's no extra lines, it won't show any extra content.

Maybe we give everyone a single line always, and hope they don't just write a single really really long line?

Or we split the text after so many words or a newline, whatever comes first?

wcivf/assets/scss/main.scss
```diff
@@ -150,3 +150,21 @@ summary {
   cursor: pointer;
   color: $dc-bright-pink;
 }
+
+.ds-scope { 
+  .ds-candidate {
+    background-color: white;
```
symroe, line 164: Can we remove these line for cards and just make the whole background white for the whole site please? This is the direction we're heading in anyway, so may as well do now

wcivf/apps/people/models.py
```diff
@@ -223,6 +223,14 @@ def statement_remainder(self):
 
         return statement_remainder
 
+    @property
+    def deceased(self):
```
symroe: I think it's useful to abstract this, but It might need a slightly different name - for example, imagine we want to show a list of deceased candidates in future (let's not do this!), we might be tempted to iterate over candidates and do `if person.deceased`. That makes sense, but we might not realise that it's actually checking other things too.

Can we call this....er...`should_show_deceased_text`? Is that what it's for?

wcivf/apps/parties/models.py
```diff
@@ -87,6 +87,10 @@ def is_independent(self):
         """
         return self.party_id == "ynmp-party:2"
 
+    @property
+    def is_joint_party(self):
+        return self.party_id == "joint-party:53-119"
```
symroe, line 92: There are more joint parties, and the IDs [always take the format](https://github.com/DemocracyClub/yournextrepresentative/blob/bd0db75e46ef27efb2d2f459087f26f5e03cba21/ynr/apps/parties/importer.py#L41) `joint-party:{}-{}"`

symroe: So I think `self.party_id.startswith("joint-party:")` would do it?

wcivf/apps/people/templates/people/person_detail.html
```diff
@@ -26,23 +26,43 @@ <h2 class="ds-candidate-name ds-h3">
...
-          {% endif %}
-          {% for candidacy in object.current_or_future_candidacies %}
-          <ul>
-            <li>
-              {{ candidacy.election.name }} for {{ candidacy.post }}
-            </li>
-          </ul>
-          {% endfor %}
+          
+            {% if object.personpost.party.is_joint_party %}
```
symroe, line 30: I might be reading this wrong - but I don't think join parties are quite what you think they are here?

A candidate can stand for one or more elections for a joint party - basically agreeing to campaign/work under the manifestos of more than one party.

That's not the same as that person standing for more than one party in _different candidacies_.

Joint parties are by agreement with the parties in advance, but the case I think you're thinking about in this conditional is where a person has >1 current candidacy and one is for e.g Labour and the other is e.g independent.

wcivf/assets/scss/main.scss
```diff
@@ -3,6 +3,63 @@
...
+// // Main files
+@import 'partials/_fonts.scss';
+@import 'partials/_variables.scss';
+@import 'partials/_type.scss';
+@import 'partials/_forms.scss';
+@import 'partials/_button.scss';
+@import 'partials/_call-to-action.scss';
+@import 'partials/_stack.scss';
+@import 'partials/_cluster.scss';
+// @import 'partials/_grid.scss';
```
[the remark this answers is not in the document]
symroe: I see you did this later - ignore!

wcivf/apps/people/models.py
symroe: Sorry you tidied this up later - glad you caught this!

wcivf/apps/people/templates/people/not_current_person_detail.html
pmk01: Yes, this is an interesting one. We also don't do parish elections, which some people might also be confused by.

We could say something like :

 [candidate] has stood in [n] elections collected in the Democracy Club database*.
or
According to our database*, [candidate] has stood in [n] elections since 2010

*perhaps 'Democracy Club database' could link to a separate page on WCIVF which explains the history of the project in a couple of sentences? Awkward perhaps, but it would be useful for us to have it all explained concisely somewhere.

wcivf/apps/people/models.py
```diff
@@ -134,6 +134,7 @@ def has_any_contact_info(self):
     @property
     def cta_example_details(self):
         attrs = (
+            ("candidacy", "candidacy"),
```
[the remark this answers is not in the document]
VirginiaDooley: I agree, it doesn't made sense. I added this early on before I fully understood candidacies. I will remove it.

wcivf/assets/scss/main.scss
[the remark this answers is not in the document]
VirginiaDooley: We aren't currently using grids so I can take it out and edit down to only those components that are currently in use in the new person page. 👍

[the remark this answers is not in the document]
VirginiaDooley: Actually, some of these components rely on others, so I'll need to review the underlying styles before making any changes.

wcivf/apps/people/models.py
VirginiaDooley: Absolutely. I'll put this on my 'things to refactor'. 👍

wcivf/assets/scss/main.scss
VirginiaDooley: And keep the card backgrounds grey?

wcivf/apps/people/models.py
VirginiaDooley: In a later version of this file, we call this method `display_deceased`. Ok?

wcivf/apps/parties/models.py
VirginiaDooley: Yes! Makes sense. Change made. 👍

wcivf/apps/people/templates/people/person_detail.html
VirginiaDooley: As discussed in Slack, I had misunderstood joint_parties so thanks for clarifying. I've removed the conditional - making the intro text the same for multiple candidacies regardless of party (although slight edit for 'a' vs 'an' in the case of Independents) and updated the method to account for different ids as described in the previous comment.

wcivf/assets/scss/main.scss
symroe: The card backgrounds are just transparent - so will end up white when the background is

wcivf/apps/people/models.py
symroe: Absolutely - sorry this is one of the things I missed the later change to as I was working though the commits

wcivf/apps/people/models.py
VirginiaDooley: Good point. I tried a few ideas based on your suggestions and I thought the first sentence intro followed by full statement (less first line) worked well.

wcivf/apps/people/templates/people/not_current_person_detail.html
VirginiaDooley: All good ideas. For simplicity, I suggest:

[candidate] has stood in [n] elections since 2010*

*link to explainer as suggested by @pmk01 or the summary given by @symroe

wcivf/apps/people/templates/people/person_detail.html
```diff
@@ -15,39 +15,7 @@
...
-          {% endif %}
-        </div>
-
-        {% if object.photo_url %}
-        <div class="ds-candidate-image">
-          <img src="{{ object.photo_url }}" alt="profile photo of {{ object.name }}">
-        </div>
-        {% endif %}
-      </section>
+      {% include "people/includes/_person_intro_card.html" %}
```
symroe, line 18: This file is _so_ much better now!

wcivf/apps/people/models.py
```diff
@@ -206,19 +206,16 @@ def statement_count(self):
...
-        statement_to_voters = self.statement_to_voters
-        statement_split = statement_to_voters.splitlines()
-        statement_intro = statement_split[0]
-        statement_remainder = statement_split.remove(statement_intro)
-        statement_remainder = "\n".join(statement_split)
+        statement_intro = self.statement_intro
+        # statement_to_voters = self.statement_to_voters
+        # statement_split = statement_to_voters.split("\n")
+        statement_split = self.statement_to_voters.split(".")
+        statement_remainder = "".join(statement_split[1:])
```
symroe, line 218: I think this is a good improvement! We'll have to see how it works in the wild and might have to revisit the logic (I don't know how much I trust candidates to use English properly!). The good thing about the implementation is we can change it quickly if we need to!

symroe:
Happy with where this is now - feel free to punt the URL parsing to an issue - I'll leave that one up to you to decide :+1:

pmk01:
Just jumping in to say these look brilliant! My only thought is on the photo cropping, have you tested this with candidates who have a particularly low-res picture, or headshots in profile? This is a small number of images and will probably encourage us to find better photos, but thought I'd mention it.

VirginiaDooley:
> Just jumping in to say these look brilliant! My only thought is on the photo cropping, have you tested this with candidates who have a particularly low-res picture, or headshots in profile? This is a small number of images and will probably encourage us to find better photos, but thought I'd mention it.

Do you have a lo-res candidate image in mind so I can test it?

wcivf/apps/people/models.py
VirginiaDooley: Punting this to an issue as suggested so we can get this reviewed and rolled out faster.

pmk01:
Yes, sorry, I couldn't come up with any immediately. Here are a couple of awkward examples from upcoming elections:
https://candidates.democracyclub.org.uk/person/72906/neil-wilson
https://candidates.democracyclub.org.uk/person/73218/gordon-shanks
https://candidates.democracyclub.org.uk/person/73540/michelle-frampton

VirginiaDooley:
> https://candidates.democracyclub.org.uk/person/73540/michelle-frampton



> Yes, sorry, I couldn't come up with any immediately. Here are a couple of awkward examples from upcoming elections:
> https://candidates.democracyclub.org.uk/person/72906/neil-wilson
> https://candidates.democracyclub.org.uk/person/73218/gordon-shanks
> https://candidates.democracyclub.org.uk/person/73540/michelle-frampton

What do you think? Next question I have is why the logic is rendering with 0 elections. 🤔 

Screen Shot 2021-02-25 at 8 49 34 AM

Screen Shot 2021-02-25 at 8 50 56 AM

Screen Shot 2021-02-25 at 8 49 34 AM

pmk01:
Good stuff! No significant difference from now, then (in terms of picture quality).

VirginiaDooley (commented):
@pmk01 are you happy with this language? 
Screen Shot 2021-02-25 at 9 20 53 AM

pmk01:
Yes - how does it work with Mayors & PCCs?

pmk01:
Here's a good candidate to try it on:
https://candidates.democracyclub.org.uk/person/6023/george-jabbour

VirginiaDooley:
> Yes - how does it work with Mayors & PCCs?

Screen Shot 2021-02-25 at 9 53 21 AM

Happy to send the entire page; not sure if you just wanted to see Previous elections for https://candidates.democracyclub.org.uk/person/6023/george-jabbour

symroe:
Sorry to jump in again, but for the "vote count not available" message on the table: we might have details about if they were elected or not, even if we don't have votes cast. For example, we'll have all general election results in terms of "elected: y/n".

Another thing to punt to an issue, but we should show that if we have it

pmk01:
Further on results, do we want to add (not elected)? What do you guys think?

That table also reveals a number of problems with the names of our historic elections, especially Mayor of Watford, but I don't think any of that relates to this change. Something to return to after May, I think.

VirginiaDooley:
> Sorry to jump in again, but for the "vote count not available" message on the table: we might have details about if they were elected or not, even if we don't have votes cast. For example, we'll have all general election results in terms of "elected: y/n".
> 
> Another thing to punt to an issue, but we should show that if we have it.
> 
> I think `elected` is a NullBoolean, so the values are one of `None, True, False`. e.g, if it's not `None` then we have asserted that they either were or weren't elected.

I think this is what you're referring to @symroe ? I've included elected regardless if votes cast is available https://github.com/DemocracyClub/WhoCanIVoteFor/pull/660#pullrequestreview-598616823

wcivf/apps/people/templates/people/includes/_person_previous_elections_card.html
```diff
@@ -21,10 +21,10 @@ <h3 class="ds-padded">Previous Elections</h3>
         <td>{{ person_post.election.election_date|date:"Y" }}</td>
         <td>{{ person_post.post.label }}: {{ person_post.election }}</td>
         <td>{{ person_post.party.party_name }}</td>
-        {% if person_post.elected %}
-        <td>{{ person_post.votes_cast|intcomma }} votes (elected)</td>
+        {% if person_post.votes_cast %}
+        <td>{{ person_post.votes_cast|intcomma }} votes{% if person_post.elected%} (elected){% endif %}</td>
         {% else %}
-        <td>{{ person_post.votes_cast|intcomma }} votes</td>
+        <td>Vote count not available{% if person_post.elected%} (elected){% endif %}</td>
```
symroe, line 27: Yep, this is the line I was talking about!

I think the logic is slightly different, here in pseudo code:
```
if person_post.elected == True:
    (elected)
elif if person_post.elected == False:
  (not elected)
else:
    Vote count not available

**Comment by Heydon** [comment]
@VirginiaDooley The photo cropping issue is due to `object-fit: cover` which is used to fit the image into the available space by cropping rather than stretching. For some reason, the croppable area is two narrow in the pictured configuration, which I’ll need to tweak.

**Comment by Heydon** [comment]
@VirginiaDooley I'd have to see the markup, but we should be able to avoid this weird styling if the `<details>` element truncates the blockquote’s content, rather than splits it into two blockquotes? Is that what’s happening here? 

blockquote split in two, breaking the border style

File: wcivf/apps/people/templates/people/includes/_person_previous_elections_card.html (hunk 18 above)
Comment (VirginiaDooley) [reply wcivf/apps/people/templates/people/includes/_person_previous_elections_card.html:27]: Ah makes sense, I'll make that edit, thanks!

**Comment by VirginiaDooley** [comment]
> @VirginiaDooley I'd have to see the markup, but we should be able to avoid this weird styling if the `<details>` element truncates the blockquote’s content, rather than splits it into two blockquotes? Is that what’s happening here?
> 
> blockquote split in two, breaking the border style

https://github.com/DemocracyClub/WhoCanIVoteFor/pull/660/files#diff-ae321f98dbcc359e064a2026d2dbbbefa7fc149669e954b138a7aee7ab749005

**Comment by Heydon** [comment]
@VirginiaDooley Okay, it seems you have a `<blockquote>` inside a `<blockquote>` there. I’ve added an example to the `release/0.1.0` branch on how to do this successfully. https://github.com/DemocracyClub/design-system/blob/release/0.1.0/src-site/components/details.md

**Comment by VirginiaDooley** [comment]
> Further on results, do we want to add (not elected)? What do you guys think?
> 
> That table also reveals a number of problems with the names of our historic elections, especially Mayor of Watford, but I don't think any of that relates to this change. Something to return to after May, I think.

Just following up, have you captured this somewhere already? If not I can make an issue.

File: wcivf/apps/people/templates/people/includes/_person_previous_elections_card.html [hunk 19]
```diff
@@ -0,0 +1,34 @@
...
+    <tr>
+        {% for person_post in object.past_not_current_candidacies %}
+        {{ person_post.post_election.short_cancelled_message_html }}
+        {{ person_post.post_election.short_cancelled_message_html }}
+        {{ person_post.post_election.short_cancelled_message_html }}
+        <td>{{ person_post.election.election_date|date:"Y" }}</td>
+        <td>{{ person_post.post.label }}: {{ person_post.election }}</td>
+        <td>{{ person_post.party.party_name }}</td>
+        {% if person_post.elected %}
+            <td>{% if person_post.votes_cast %}{{ person_post.votes_cast|intcomma }} votes (elected){% else %}Vote count not available{% endif %}</td>
```
michaeljcollinsuk, line 25: So I think in the `else` here, we would want to say "Elected" rather than "Vote count not available"? Because we know they were elected (from the check on line 24) we just don't know how many votes they got

wcivf/apps/people/models.py
```diff
@@ -153,6 +153,76 @@ def get_max_facebook_ad_spend(self):
...
+    @property
+    def statement_remainder(self):
+        statement_split = self.statement_to_voters.split(".")
+        statement_remainder = ".".join(statement_split[1:])
+
+        return statement_remainder
+
+    @property
+    def display_deceased(self):
+        if self.death_date and self.current_or_future_candidacies:
```
michaeljcollinsuk, line 221: With these sort of conditionals, you can return the result of your conditional directly:
```
return self.death_date and self.current_or_future_candidacies
```
So If both statements are True, True is returned, otherwise will return False

wcivf/apps/people/models.py
```diff
@@ -153,6 +153,76 @@ def get_max_facebook_ad_spend(self):
...
+        facebook_split = list(filter(None, facebook_personal_url.split("/")))
+        facebook_personal_username = facebook_split[-1]
+
+        return facebook_personal_username
+
+    @property
+    def facebook_username(self):
+        facebook_page_url = self.facebook_page_url
+        facebook_split = list(filter(None, facebook_page_url.split("/")))
+        facebook_username = facebook_split[-1]
```
michaeljcollinsuk, line 168: Minor thing, but when you set a variable and then return it immediately without doing anything else, you can simply return instead:
```
@property
def facebook_username(self):
    facebook_page_url = self.facebook_page_url
    facebook_split = list(filter(None, facebook_page_url.split("/")))
    return facebook_split[-1]
```

wcivf/apps/people/templates/people/includes/_person_about_card.html
```diff
@@ -0,0 +1,40 @@
+{% if object.wikipedia_bio or object.associatedcompany_set.exists or object.cv %}
+<section class="ds-card">
+  <div class="ds-card-body">
+    <h2 class="ds-candidate-name ds-h3">
+      About {{ object.name }}
+    </h2>
+    <p>
+      {% if object.cv %}
```
michaeljcollinsuk, line 8: I think this `if` should be before the opening `<p>`? Otherwise if there is no CV you will have an empty `<p>` element

michaeljcollinsuk: Or perhaps that `<p>` just isn't needed?

wcivf/apps/people/templates/people/includes/_person_about_card.html
```diff
@@ -0,0 +1,40 @@
...
+    <p>{{object.name}} is associated with the following companies:</p>
+    <ul>
+      {% for company in object.associatedcompany_set.all %}
+      <li>
+        <strong>{{ company.company_name }}</strong>
+
+        {{ company.role }}{% if company.role_status %}({{ company.role_status }}){% endif %}
+        {{ company.role_appointed_date | date:"Y" }} &ndash; {{ company.role_resigned_date | date:"Y"}}
+
+        <a href="https://beta.companieshouse.gov.uk/company/{{ company.company_number }}/">#}
```
michaeljcollinsuk, line 31: Looks like a typo at end of this link, and one at the end of line 32

wcivf/apps/people/templates/people/person_detail.html
```diff
@@ -10,356 +10,63 @@
...
-  {% if object.should_show_email_cta %}
-    <p>You can also email {{ object.name }} directly to ask them to add information to this page.
-  <p><a href="{% url 'email_person_view' pk=object.pk ignored_slug=person.name|slugify %}" class="link-button">Ask the candidate for more information &raquo;</a></p>
-{% endif %}
 
-{#  <h4>Upload your leaflets</h4>#}
-{#  <p>If you've received election leaflets from {{ object.name }}, please take a photo#}
-{#  of them and upload them to ElectionLeaflets.org</p>#}
-{#  <p><a class="link-button" href="https://electionleaflets.org/">Add leaflets</a> </p>#}
+      {#{% include "feedback/feedback_form.html" %}#}
```
michaeljcollinsuk, line 42: I reckon we could just delete this and as there has been some talk of changing what sort of feedback we ask for anyway

wcivf/apps/people/templates/people/includes/_person_previous_elections_card.html
VirginiaDooley: Yes, now that I think about it, maybe we want to say both where the vote count is not available.

PR #664, opened 2021-02-12
Feature/election enhancements

### LINKED ISSUE #183: 2015 constituency pages should link to 2017 constituency pages
User request from slack.

[Example](https://whocanivotefor.co.uk/elections/2015/post-65685/member-of-parliament-for-brentwood-and-ongar)

michaeljcollinsuk:
Closes #183 

Display the next ballot for the same post on a previous ballot page. Open to suggestions about better wording/placing on the page, screenshot below of how it looks currently

Screenshot 2021-02-12 at 17 21 00

Also updates the Election page to remove the hardcoded "Posts" and return the same work that is used on the ballot page e.g. "wards", "divisions" etc.

Screenshot 2021-02-12 at 17 21 09

Screenshot 2021-02-12 at 17 21 14

Screenshot 2021-02-12 at 17 21 19

### DIFF
--- a/wcivf/apps/elections/templates/elections/includes/_single_ballot.html
+++ b/wcivf/apps/elections/templates/elections/includes/_single_ballot.html
@@ -2,8 +2,8 @@
 {% load humanize %}
 {% load static %}
 
-<section id="election_{{ postelection.election.slug }}">
-  <div class="card">
+<section>
+  <div class="card" id="election_{{ postelection.election.slug }}">
     {% if postelection.cancelled %}
       {% include "elections/includes/_cancelled_election.html" with object=postelection only %}
     {% else %}
--- a/wcivf/apps/elections/templates/elections/includes/_single_ballot.html
+++ b/wcivf/apps/elections/templates/elections/includes/_single_ballot.html
@@ -12,8 +12,8 @@
         {{ postelection.election.name_without_brackets }}
         <span aria-hidden="true">🗳️</span>
       </h1>
-      {% if postelection.friendly_name != postelection.election.nice_election_name %}
-        <h2>{{ postelection.friendly_name }}</h2>
+      {% if not postelection.is_pcc and not postelection.is_mayoral %}
+        <h2>{% if postelection.is_london_assembly_additional %}Additional members{% else %}{{ postelection.friendly_name }}{% endif %}</h2>
       {% endif %}
       {% if postelection.metadata.coronavirus_message %}
         <div style="border:1px solid red;margin:1em 0;padding:1em">
@@ -38,7 +38,7 @@
 
       <p>
         {% if postelection.election.in_past %}
-          <strong>{{ postelection.people|length }} candidates</strong> stood in {{ postelection.post.full_label }}.
+          <strong>{{ postelection.people|length }} candidates</strong> stood in the {{ postelection.friendly_name }}.
         {% else %}
           {# Display different messages depending on the number of candidates #}
           {# Case: No candidates for a contested election #}
@@ -53,12 +53,7 @@
                 You will have one vote, and can vote for a single party list or independent candidate.
               {% else %}
                 You can choose from <strong>{{ postelection.people|length }} candidates</strong>
-                in the
-                {% if 'mayor' in postelection.election.slug %}
-                  {{ postelection.election.nice_election_name }}.
-                {% else %}
-                  {{ postelection.post.full_label }}.
-                {% endif %}
+                in the {{ postelection.friendly_name }}.
 
                 {% if postelection.winner_count and postelection.election.voting_system_id == 'FPTP' %}
                   You will have {{ postelection.winner_count|intword }} vote{{ postelection.winner_count|pluralize }}.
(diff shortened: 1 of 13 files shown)


wcivf/apps/elections/templates/elections/includes/_single_ballot.html
```diff
@@ -28,6 +28,9 @@ <h2>{{ postelection.friendly_name }}</h2>
         {% else %}
           {{ postelection.election.in_past|yesno:"was,will be" }} held
           <strong>{{ postelection.election.election_date|naturalday:"\o\n l j F Y" }}</strong>.
+          {% if postelection.next_ballot %}
+            <a href="{{ postelection.next_ballot.get_absolute_url }} ">Click here to view the upcoming election for this post.</a>
```
symroe, line 32: Couple of points about this:

1. using "click here" is [generally not a good idea for links](https://webaccess.berkeley.edu/ask-pecan/click-here) - I know you've included the descriptive text after "click here", but I don't think that adds a lot. 

2. Related: it might be nicer state that we know an election is happening slightly differently, like "{{ place name }} has an election on the {{ date }}" - could check in with @pmk01 about what he thinks on this.

3. Can we use the same logic to get the nice area name, rather than "post"? Now we have it, I'm keen to use it everywhere :)

Sorry for the bike-shed about a single line!

wcivf/apps/elections/templates/elections/includes/_single_ballot.html
```diff
@@ -28,6 +28,9 @@ <h2>{{ postelection.friendly_name }}</h2>
         {% else %}
           {{ postelection.election.in_past|yesno:"was,will be" }} held
           <strong>{{ postelection.election.election_date|naturalday:"\o\n l j F Y" }}</strong>.
+          {% if postelection.next_ballot %}
```
symroe, line 31: Being in the `_single_ballot` partial, this will be called for all current elections too. I don't think that's helpful, as we know that a current ballot wont have a newer election.

Can you move this call to the `post_view.html` template directly, or if we want it to show in the card, add a conditional flag if we want to make this call or not?

In general, I try to be careful about adding any extra queries to the pages that get the most traffic - every little helps!

wcivf/apps/elections/templates/elections/includes/_single_ballot.html
pmk01: I agree with Sym! 

perhaps:
The next election for {{position}} is due to take place on {{date}}.

If we wanted to be really clever, we could link to the previous election as well? I don't know if that's useful for current elections, but may be for past ones. Happy to have more of a think and chat about this.

Relatedly, I really like the way Wikipedia election infoboxes provide links to the previous election with a little arrow and year:

Screenshot 2021-02-15 122542

michaeljcollinsuk: I've moved the link to a new card that is only displayed on the `post_view.html`:

Screenshot 2021-02-15 at 16 35 11

This will only display if the election is not the 'current' election and if there is a future election to display. But not sure if you think that it is overkill?

@pmk01 from a technical point, we can link to the previous election. But I'm not sure how/where the arrow example from Wikipedia would best work, so yes lets have a chat about it if its something you'd like to get working

wcivf/apps/elections/templates/elections/includes/_single_ballot.html
michaeljcollinsuk: Thought it best to do both - remove from the partial, but also add in the `current` check in case it does get moved there or elsewhere

symroe: Yeah, good idea :+1:

wcivf/apps/elections/templates/elections/includes/_single_ballot.html
michaeljcollinsuk: In fact now I am starting to wonder if I haven't potentially misinterpreted the original issue entirely. I have been thinking about it in terms of wanting to display a link to the next election that is taking place. So a past election would always link to the next one taking place (if it exists). But should it be the next election chronologically? So 2015 would link to 2017, then 2017 to 2019 etc?

pmk01: Oh, that is a good question. I like the text in your example and would be very happy to deploy that now. I suppose we could show a list of previous elections, though at some point that might get a bit busy... @symroe do you know which seat has had the most elections since we started tracking by-elections in 2016?

pmk01: I am actually very happy with either 'most recent' or 'next election'. Maybe we could ask the volunteers. 

It's not a bad thing if we don't make a complete decision right now, as after May I want to have a conversation about how we've become an archive and how we manage information across the board anyway.

michaeljcollinsuk: Perhaps that is good compromise then - we go with the above for now, where all past elections link to the upcoming election. As that would seem the most useful feature for right now anyway. And then after May we could revisit this to decide if we want to implement wider changes, discussing with the volunteers.

PR #673, opened 2021-02-22
Try to fix dependabot ignore syntax

michaeljcollinsuk:
Dependabot PR for Django 3.0 was opened to this tried to update the config to ignore updates for version 3 matching syntax in the docs https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#specifying-dependencies-and-versions-to-ignore

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,5 @@ updates:
     interval: daily
   open-pull-requests-limit: 10
   ignore:
-  - dependency-name: django
-    versions:
-    - 2.x
+    - dependency-name: django
+      versions: [">=3.0", "<3.2"]


chris48s:
YMMV but the way I do this on a django site where I track LTS releases is I set:

```yml
ignore:
  - dependency-name: "django"
    versions: [">=3.0", "<3.2"]  # don't offer to bump django until 3.2 LTS released
```

That way you still get the 2.2.x patches, ignore the 3.0.x and 3.1.x version bumps but when the next LTS comes out in ~April you'll start getting reminders to migrate. (That syntax is confirmed working).

michaeljcollinsuk:
> YMMV but the way I do this on a django site where I track LTS releases is I set:
> 
> ```yaml
> ignore:
>   - dependency-name: "django"
>     versions: [">=3.0", "<3.2"]  # don't offer to bump django until 3.2 LTS released
> ```
> 
> That way you still get the 2.2.x patches, ignore the 3.0.x and 3.1.x version bumps but when the next LTS comes out in ~April you'll start getting reminders to migrate. (That syntax is confirmed working).

Good tip thanks!

PR #678, opened 2021-02-24
Feature/local parties 2021

michaeljcollinsuk:
Update the local parties import management command so that it can be run without passing a file, reading directly from urls. This will allow us to clean up the cron jobs in the deployment repo when merged. This has been done by refactoring to remove the bulk of the import process in to a new importer class.

Have left partial commit history to try to help with the review, but intend to rebase before merging

### DIFF
--- a/wcivf/apps/core/helpers.py
+++ b/wcivf/apps/core/helpers.py
@@ -1,6 +1,7 @@
 import re
 from contextlib import contextmanager
 from datetime import datetime, timedelta
+from urllib.parse import urlparse
 import json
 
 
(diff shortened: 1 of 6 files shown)


symroe (commented):
Couple of comments, but I think this is looking good.

One thing on the error reporting that might be useful to know about is how we manage this in YNR with a context manager: https://github.com/DemocracyClub/yournextrepresentative/blob/master/ynr/apps/candidates/management/commands/candidates_import_from_live_site.py#L27

It's possible we don't need that here, but it's a useful pattern to know about I think.

symroe (approved):

PR #680, opened 2021-02-25
Feature/referendums

michaeljcollinsuk:


michaeljcollinsuk:
Some screenshots @pmk01 may be interested in 👀. Worth noting that once the ballot cards have been updated to use the design system, these will look prettier as it will include the new blockquote styles.

Results for Sheffield postcode are displayed as:
Screenshot 2021-03-08 at 13 03 58
Screenshot 2021-03-08 at 13 04 02

Postcodes that return a ballot for City and East
Screenshot 2021-03-08 at 13 04 14
Screenshot 2021-03-08 at 13 05 10

Results for postcodes without a referendum are unchanged:
Screenshot 2021-03-08 at 13 05 43

pmk01:
These are excellent, I have no other comments!

wcivf/apps/elections/views/mixins.py
```diff
@@ -69,6 +69,11 @@ def postcode_to_ballots(self, postcode, compact=False):
         pes = pes.select_related("post")
         pes = pes.select_related("election")
         pes = pes.select_related("election__voting_system")
+
+        # only retrieve referendums until 6th May
+        if date.today() <= date(2021, 5, 6):
```
symroe, line 74: This is ok, but will result in the tests failing after the election. Let's keep this `prefetch_related` in all cases, so the number of queries stays the same, but keep the logic in `get_referendums`

wcivf/apps/elections/views/postcode_view.py
```diff
@@ -82,6 +83,21 @@ def get_todays_ballots(self):
             ballot for ballot in self.ballots if ballot.election.is_election_day
         ]
 
+    def get_referendums(self):
+        """
+        Yield all referendums associated with the ballots for this postcode.
+        After 6th May return an empty list to avoid displaying unwanted
```
symroe, line 89: This makes me think of two things:

1. We should show these for some weeks after election day - people come back after to see what happened.
2. We should consider telling people what happened!

michaeljcollinsuk: Good points - I feel like there is a few different things to think about with this, from what we display to how we display it, and more basic things like the wording in the card. So I think it probably merits a separate issue/PR? 

So for now if you are happy I thought I'd leave in the hard 'stop displaying after 6th May' so that this can be merged? And I'll add a new issue for further changes.

symroe: Yep, that words for me! Are you ok to make those issues?

symroe (approved):

PR #693, opened 2021-03-23
Frontend snags

### LINKED ISSUE #691: CVs
Some candidates are haunted by the ghosts of curriculum vitae past, eg: https://whocanivotefor.co.uk/person/1491/stephen-williams

Screenshot 2021-03-16 214022

I think we should remove CVs as an option entirely because **a)** every CV we hold is years out of date, and **b)** http://cv.democracyclub.org.uk/ ain't gonna be up and running anytime soon. We had loads of complaints about it when we tried to resurrect it in 2019.

symroe:
I was about to write up some issues about these, but in doing so realised it would be quicker to do the work!

## Ballot page cards no longer clash

Screenshot from 2021-03-23 11-44-59

## CVs are gone
We don't have new ones, or a quick way to remove them, so let's ignore them at the moment

Fixes #691

## Companies are gone
As per CVs, more problem than they're worth

## Don't link to the leaflets site, it's down
We can add back in if we get the site up again, so just commented out

## Link and sort a person's upcoming ballots

### DIFF
--- a/wcivf/apps/elections/templates/elections/includes/_single_ballot.html
+++ b/wcivf/apps/elections/templates/elections/includes/_single_ballot.html
@@ -2,141 +2,140 @@
 {% load humanize %}
 {% load static %}
 
-<section>
-  <div class="card" id="election_{{ postelection.election.slug }}">
-    {% if postelection.cancelled %}
-      {% include "elections/includes/_cancelled_election.html" with object=postelection only %}
-    {% else %}
-      <h1>
-        <span aria-hidden="true">🗳️</span>
-        {{ postelection.election.name_without_brackets }}
-        <span aria-hidden="true">🗳️</span>
-      </h1>
-      {% if not postelection.is_pcc and not postelection.is_mayoral %}
-        <h2>{% if postelection.is_london_assembly_additional %}Additional members{% else %}{{ postelection.friendly_name }}{% endif %}</h2>
+<div class="card" id="election_{{ postelection.election.slug }}">
+  {% if postelection.cancelled %}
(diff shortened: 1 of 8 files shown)


wcivf/apps/elections/templates/elections/includes/_single_ballot.html
```diff
@@ -2,141 +2,140 @@
...
-      {% if postelection.metadata.coronavirus_message %}
-        <div style="border:1px solid red;margin:1em 0;padding:1em">
-          <strong>{{ postelection.metadata.coronavirus_message|safe }}</strong>
-        </div>
+<div class="card" id="election_{{ postelection.election.slug }}">
+  {% if postelection.cancelled %}
+    {% include "elections/includes/_cancelled_election.html" with object=postelection only %}
+  {% else %}
+    <h1>
+      <span aria-hidden="true">🗳️</span>
```
VirginiaDooley, line 10: How does this work?

VirginiaDooley (approved):

wcivf/apps/elections/templates/elections/includes/_single_ballot.html
symroe: It's just the unicode emoji "[Ballot Box with Ballot](https://emojipedia.org/ballot-box-with-ballot/)" - I like using emoji as a free way to add icons - no icon font or images required!

The `aria-hidden="true"` is to stop screen readers reading the emoji out - there's no value in the icon being there as it's just visual decoration, so we don't want each heading to be read as `"Ballot box with ballot Stroud local election ballot box with ballot"`

PR #696, opened 2021-03-23
Open external links in new tab

### LINKED ISSUE #658: Should links to external sites open in a new tab?
At present, all links on WCIVF open in the same tab - this might be annoying when we are linking to external sites. Should we change this?

VirginiaDooley:
Closes #658


symroe:
Sorry, there might have been some confusion about this issue - I don't think I added it to backlog? If I did, I think it's very much a open question - the quote in the issue gives good reason _not_ to do this, as it's generally considered less good for unsuitability.

I'd like to park this until we have good evidence that it's a good idea - if the W3C suggest not doing it then I think we need a good reason to go against that!

VirginiaDooley:
Sorry! Completely misinterpreted the guidance.

symroe:
No I think it's my fault in not just closing the issue or making it more clear in my reply! Sorry you spent time on it!

Issue #122, opened 2017-05-19
Improve hustings display


symroe:
* [ ] Add video / post event URL
* [ ] Hide past events, unless they have post event links
* [ ] Deal with deleting events somehow
* [ ] Page listing all hustings, linked to in footer (For Google)
* [ ] Page per event, with schema.org markup in the hopes that Google pick up on it

annapowellsmith:
I actually think we should show past events even if no video URL, though others may disagree.

Does "deal with deleting events" mean that rows may vanish from the spreadsheet sometimes? I guess we just start the importer by deleting all events, if so.

chris48s:
> Does "deal with deleting events" mean that rows may vanish from the spreadsheet sometimes? I guess we just start the importer by deleting all events, if so.

We already delete them all and re-import:
https://github.com/DemocracyClub/WhoCanIVoteFor/blob/master/wcivf/apps/hustings/management/commands/import_hustings.py#L134
This handles situations where the row has just been deleted, which is fine.

I think what @symroe was hoping to do is be able to mark Hustings events as being cancelled (ie.: if there was one scheduled and it is now cancelled this is valuable information to people who were planning to attend).
At the moment it either just disappears because someone deleted the row (best-case scenario) or someone well-meaning indicates this in the spreadsheet by using formatting (e.g: strikethrough) which is just ignored when we pull it through onto WhoCIVF and we present it as if it were still happening (worst-case scenario). If I spot these I delete the row, but I only ever look at it if the import is failing.

michaeljcollinsuk:
@symroe Pretty sure that the last two on the checklist have not been implemented. Would the idea of the Hustings page be similar to https://whocanivotefor.co.uk/elections/? With a detail page for each Husting event also?

symroe:
In the past we've included the hustings for a ballot on the postcode page:

https://github.com/DemocracyClub/WhoCanIVoteFor/commit/7cfbb702b62509f173397f7b12b60bcb357fd604#diff-ba8a90759360c60a22821bad430a05de0da3522971ea17534c5c135d0244bdbbL142

pmk01:
Good Q... if it's an easy thing to do I am always in favour of more lists of things, you never know when you might need them. But otherwise I don't know if it's useful for users... I think it's worth asking some questions in people in Slack?

PR #698, opened 2021-03-23
Update hustings import management command for 2021

michaeljcollinsuk:
- Import hustings via urls by default
- Tried to keep backwards compatible so old hustings could be imported (is this necessary?)
- Optionally import from file

### DIFF
--- a/wcivf/apps/elections/templates/elections/includes/_hustings_list.html
+++ b/wcivf/apps/elections/templates/elections/includes/_hustings_list.html
@@ -3,6 +3,6 @@
 {% for husting in hustings %}
 <p>
 {% if husting.in_past %}(Past event) {% endif %}
-<strong>{{ husting.starts|naturalday:"l j F Y"|title }} {{ husting.starts|date:"H:i"}}{% if husting.ends %}&ndash;{{ husting.ends | date:"H:i"}}{% endif %}:</strong> {% if husting.url %}<a href="{{ husting.url }}">{% endif %}{{ husting.title}}{% if husting.url %}</a>{% endif %}, {{ husting.location }} {{ husting.postcode }}. {% if husting.postevent_url %}<a href="{{ husting.postevent_url }}"><i class="icon-videocam" aria-hidden="true"></i> See video or other info from this event.</a>{% endif %}
+<strong>{{ husting.starts|naturalday:"l j F Y"|title }} {{ husting.starts|date:"H:i"}}{% if husting.ends %}&ndash;{{ husting.ends | date:"H:i"}}{% endif %}:</strong> {% if husting.url %}<a href="{{ husting.url }}">{% endif %}{{ husting.title}}{% if husting.url %}</a>{% endif %}{% if husting.location %}, {{ husting.location }} {{ husting.postcode }}{% endif %}{% if husting.postevent_url %}<br><a href="{{ husting.postevent_url }}"><i class="icon-videocam" aria-hidden="true"></i> See video or other info from this event.</a>{% endif %}
 </p>
 {% endfor %}
--- a/wcivf/apps/elections/templates/elections/includes/_hustings_list.html
+++ b/wcivf/apps/elections/templates/elections/includes/_hustings_list.html
@@ -1,8 +1,0 @@
-{% load humanize %}
-
-{% for husting in hustings %}
-<p>
-{% if husting.in_past %}(Past event) {% endif %}
-<strong>{{ husting.starts|naturalday:"l j F Y"|title }} {{ husting.starts|date:"H:i"}}{% if husting.ends %}&ndash;{{ husting.ends | date:"H:i"}}{% endif %}:</strong> {% if husting.url %}<a href="{{ husting.url }}">{% endif %}{{ husting.title}}{% if husting.url %}</a>{% endif %}{% if husting.location %}, {{ husting.location }} {{ husting.postcode }}{% endif %}{% if husting.postevent_url %}<br><a href="{{ husting.postevent_url }}"><i class="icon-videocam" aria-hidden="true"></i> See video or other info from this event.</a>{% endif %}
-</p>
-{% endfor %}
--- a/wcivf/apps/elections/templates/elections/includes/_single_ballot.html
+++ b/wcivf/apps/elections/templates/elections/includes/_single_ballot.html
@@ -141,13 +141,14 @@
     <div class="card">
       <h3>
         <span aria-hidden="true">📅</span>
-        Election events near you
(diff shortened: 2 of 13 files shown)


michaeljcollinsuk:
Some screenshots of updated wording of those that we have currently. Below are from the ballot view e.g. [Mayor of Bristol](https://whocanivotefor.co.uk/elections/mayor.bristol.2021-05-06/)

Screenshot 2021-03-23 at 18 07 32

In the above you can see how past info would be displayed where it says "See video or other info from this event" (as someone seems to have entered this info already incorrectly on the [spreadsheet](https://docs.google.com/spreadsheets/d/1Krf0icsp-bExEKBpnQi_WZ6pcsT10MJ3p1NDBquf_ik/edit#gid=908428564)

Screenshot 2021-03-23 at 18 07 44
Screenshot 2021-03-23 at 18 14 54

And from a candidate/person page e.g. [https://whocanivotefor.co.uk/person/13545/sean-driscoll](https://whocanivotefor.co.uk/person/13545/sean-driscoll)
Screenshot 2021-03-23 at 18 15 58

michaeljcollinsuk:
Are the last two checkbox items on #122 still wanted? I have also noticed that there is a "Cancelled Hustings" spreadsheet that has not been utilised - should it be?

symroe:
> Are the last two checkbox items on #122 still wanted? I have also noticed that there is a "Cancelled Hustings" spreadsheet that has not been utilised - should it be?

I think @pmk01 had some ideas around a page of all hustings? Also, we wouldn't need a page per husting as schema.org can have more than one event on a page - so the work would just be to wrap each row in the event mark up (or use `ld-json`).

I'll leave it up to Peter to decide on this though - my instinct is that not enough people would find it useful without it being limited to a ballot / area, but if Google were to pick them up then it would be amazing.

I _do_ have a sneaky feature request though: can you time box to ~1 hour take a stab at adding these events to the calendar response for a postcode? If it's too hard then don't worry about it, but it would be a lovely quick win.

Also, can you explain what happens when the hustings have past?

michaeljcollinsuk:
> Also, can you explain what happens when the hustings have past?

I kept the changes to a minimum so its not changed how they [display in past years](https://github.com/DemocracyClub/WhoCanIVoteFor/pull/698/files#diff-992d4fe7721687984d932a417ae02439eb7f05a8d7a5cc0e4931af6db9eaf7a0R5) which is just to add "(Past event)" before the event title.

It might be better to split these up in to "Upcoming events" and "Past events" sections though? Also there is no clean up of events, I could have this automated to delete and stop importing after x days after the ballot if its useful?

symroe:
My view is that this is fine for now - happy for Peter disagree, but also it's better to have something imperfect up sooner that we can change later than it is to try to finish it all now.

wcivf/apps/elections/templates/elections/includes/_single_ballot.html
```diff
@@ -142,13 +142,14 @@ <h5>Citizenship</h5>
     <div class="card">
       <h3>
         <span aria-hidden="true">📅</span>
-        Election events near you
+        Husting events
```
symroe, line 145: In the past we've avoided using "hustings" as it's a word that I think most people understands. I would be nice to use language that explains what happens, and then include the jargon word alongside it.

pmk01:
How about something like:

You can meet candidates and question them at online events (often known as 'hustings'). Here are some events where [name] may be appearing:

michaeljcollinsuk:
@symroe I've got iCal events working - as you can see I will now be attending four husting events for the Mayor of Bristol election
Screenshot 2021-03-24 at 13 04 39

I've also added ld-json snippets for each event as Peter agreed not worth doing a full hustings page currently. Refactored the hustings includes as I did this.

And have updated the wording as per @pmk01 suggestion

wcivf/apps/hustings/management/commands/import_hustings.py
```diff
@@ -119,22 +143,14 @@ def handle(self, **options):
...
-                except:
-                    husting = None
-                if husting:
-                    hustings_counter += 1
-                    self.stdout.write(
-                        "Created husting {0} <{1}>".format(
-                            hustings_counter, husting
-                        )
-                    )
+        count, _ = Husting.objects.all().delete()
```
symroe, line 146: I think it's good to do this, but this wouldn't work well if we tried to use the old method of throwing a file at this command - it would remove all hustings, add in the ones from the file, but then delete them again next time it was run with no arguments.

This might be really useful for local development, but I think it should come with a warning

wcivf/apps/elections/templates/elections/includes/_hustings_list.html
```diff
@@ -3,6 +3,6 @@
 {% for husting in hustings %}
 <p>
 {% if husting.in_past %}(Past event) {% endif %}
-<strong>{{ husting.starts|naturalday:"l j F Y"|title }} {{ husting.starts|date:"H:i"}}{% if husting.ends %}&ndash;{{ husting.ends | date:"H:i"}}{% endif %}:</strong> {% if husting.url %}<a href="{{ husting.url }}">{% endif %}{{ husting.title}}{% if husting.url %}</a>{% endif %}, {{ husting.location }} {{ husting.postcode }}. {% if husting.postevent_url %}<a href="{{ husting.postevent_url }}"><i class="icon-videocam" aria-hidden="true"></i> See video or other info from this event.</a>{% endif %}
+<strong>{{ husting.starts|naturalday:"l j F Y"|title }} {{ husting.starts|date:"H:i"}}{% if husting.ends %}&ndash;{{ husting.ends | date:"H:i"}}{% endif %}:</strong> {% if husting.url %}<a href="{{ husting.url }}">{% endif %}{{ husting.title}}{% if husting.url %}</a>{% endif %}{% if husting.location %}, {{ husting.location }} {{ husting.postcode }}{% endif %}{% if husting.postevent_url %}<br><a href="{{ husting.postevent_url }}"><i class="icon-videocam" aria-hidden="true"></i> See video or other info from this event.</a>{% endif %}
```
symroe: I know this is how it was previously, but can we have some line breaks here, for legibility?

wcivf/apps/hustings/management/commands/import_hustings.py
michaeljcollinsuk: Yeah this is true - it was one of the complications of trying to keep it backwards compatible. But I wasn't really sure when we would want to import from a physical file other than local dev now (I might be wrong). I've added a user prompt to confirm when using a file, do you think that is reasonable?

symroe (approved):

PR #700, opened 2021-03-24
Stop breadcrumbs messing up Person cards

symroe:
Fixes https://github.com/DemocracyClub/design-system/issues/12

I thought this was an issue with the design system, but it was actually our old theme interacting with the design system.

This change converts the dc theme implementation of floating elements to using `inline-block`. This prevents the floated elements from messing with the block level flow in the rest of the page.


VirginiaDooley (approved):
⭐

PR #702, opened 2021-03-24
Feature/import parish councils

michaeljcollinsuk:
Import parish council elections from [Wards sheet in this google spreadsheet](https://docs.google.com/spreadsheets/d/10bm9x97r0-c1PLc09qebMIiXlwIiRN3ZgJ2p-0LFTK0/edit#gid=756401424).

In the process I started a refactor of the importers which altered the referendums importer - this could also be expanded for the hustings/local parties when there is time.

The card for the parish council elections are then displayed at the bottom of the results for a postcode page. Some screenshots below.

Example of a card where we know it is contested
Screenshot 2021-03-25 at 14 57 14

Example of a card where we don't know if it is contested or not
Screenshot 2021-03-25 at 14 39 41

Example of a card where we know it is not contested
Screenshot 2021-03-25 at 15 02 50

After 6th May the card will continue to be displayed for as long as the parent ballot is displayed, but the wording will change very slightly to read "You were also able to vote..." (assuming that it was contested - otherwise wording is unchanged).


wcivf/apps/parishes/models.py
```diff
@@ -0,0 +1,21 @@
+from django.db import models
+
+
+class ParishCouncil(models.Model):
```
symroe, line 5: Small niggle with this that might be tricky to change now - this isn't modelling a council but a council's election, right?

I think `ParishCouncilElection` might be a better name, but if it's tricky to change then we can leave as is for now (this is the MVP after all)

wcivf/apps/core/mixins.py
```diff
@@ -14,3 +23,38 @@ def read_from_file(self, path_to_file):
         with open(path_to_file, "r") as fh:
             reader = csv.DictReader(fh)
             yield from reader
+
+
+class ImportAdditionalElectionMixin(ReadFromUrlMixin):
```
symroe, line 28: :+1:

wcivf/apps/parishes/templates/parishes/includes/_card.html
```diff
@@ -0,0 +1,38 @@
...
+    {% if parish_council.is_contested %}
+    You will also be able to vote for {{ parish_council.ward_seats }} councillors to sit
+    {% else %}
+    You are also represented by {{ parish_council.ward_seats }} councillors
+    {% endif %}
+    on <a href="{{ parish_council.website }}">{{ parish_council.council_name }}</a>{% if parish_council.parish_ward_name %} {{ parish_council.parish_ward_name}}{% endif %}. 
+</p>
+
+<p>
+    {{ parish_council.council_name }} is a {{ parish_council.council_type }} Council.
```
symroe, line 22: This line looks a bit odd in the screenshots, as the data contains the council type in the name.

I suggest we either drop this line, or change the data so that `council_name` is "Foo" and the `council_type` is "Town Council". Then we can change below to:

```
In 2020-2021, {{ parish_council.council_name }} {{ parish_council.council_type }} raised
```

etc

pmk01:
I think these look brilliant. Two things:

 - can we make the text 'town/parish council' in the second line link to https://www.localgov.co.uk/What-are-parish-councils/35176
 - not all parish councils have wards. If we ever leave the 'Parish Ward Name' column blank, how will this be handled?

wcivf/apps/parishes/models.py
michaeljcollinsuk: Yeah you are right - I thought this right at the start and then never went back to it. Will sort

wcivf/apps/parishes/templates/parishes/includes/_card.html
pmk01: What about Salisbury City council tho

symroe: If it's tricky then I suggest we just drop the line. "Foo Town Council is a Town Council" just isn't useful :)

michaeljcollinsuk:
> I think these look brilliant. Two things:
> 
> * can we make the text 'town/parish council' in the second line link to https://www.localgov.co.uk/What-are-parish-councils/35176
> * not all parish councils have wards. If we ever leave the 'Parish Ward Name' column blank, how will this be handled?

1 - easy to add the link assuming we keep that line
2 - if we don't have a ward name the line will end with a full stop at the name of the council e.g. "You will also be able to vote for 3 councillors to sit on Chippenham Town Council"

wcivf/apps/parishes/templates/parishes/includes/_card.html
michaeljcollinsuk: If we drop the line, I could add the link you wanted to the first line of the next paragraph @pmk01? "Town/Parish Councils are the first tier of local government"

pmk01: Yes, that sounds good!

michaeljcollinsuk:
- @symroe Updated the model name to ParishCouncilElection and rebased
- Removed the problematic line from the card and added the URL to the next paragraph instead.
- @pmk01 here is an updated screenshot that also shows how it would look if we didn't have the Parish ward name:
Screenshot 2021-03-25 at 16 42 56

symroe (approved):
Looking really good :+1:

pmk01:
I think that's fine! Quite excited to see it in the wild.

Issue #440, opened 2019-08-07
Hide polling stations card in postcode view if all elections are cancelled


chris48s:
If **all** of the relevant elections on the postcode page are countermanded or uncontested we should hide the polling stations card.

PR #706, opened 2021-03-25
Feature/hide polling card

VirginiaDooley:
closes https://github.com/DemocracyClub/WhoCanIVoteFor/issues/440

Hides the polling station card on the postcode page if any of the post elections are cancelled or uncontested.

### DIFF
--- a/wcivf/apps/elections/templates/elections/postcode_view.html
+++ b/wcivf/apps/elections/templates/elections/postcode_view.html
@@ -32,7 +32,9 @@
 {% endif %}
 
 {# Add this at the top of the page if it's known, or at the bottom if it's not #}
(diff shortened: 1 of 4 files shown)


chris48s:
Shouldn't this be all rather than any?
If >=1 ballots are still going ahead, polling stations still open and electors can still vote

symroe:
This is looking good!

Couple of points before merge:

1. The diff in `test_postcode_views.py` is a little confusing - have some tests been deleted there, and is this the intent? Just confused, so no worries if it's all fine and I'm just reading it wrong!
2. Can you rebase this in to a single commit? I don't think we need the flakes or first commit.

VirginiaDooley:
I must have messed up a merge conflict; will fix that 👍

symroe (approved):
Nice :+1:

PR #708, opened 2021-03-26
Hotfix/parishes hustings

michaeljcollinsuk:
Some small(ish) updates following feedback from @pmk01 

We will account for parish council elections when we display the number of ballots a user will have to fill in - but only if it is contested. We also display an anchor link to the parish council card as we do for other ballots (again only if we know it is contested):
Screenshot 2021-03-26 at 13 56 16

Making sure the grouping of elections was more problematic than it first seemed, but managed to ensure would not be duplicates if you had elections occurring on different days, but also making sure that the parish council elections are displayed after all others. Some screenshots of mocking so that the elections appear in the past:
Screenshot 2021-03-26 at 15 41 25
Screenshot 2021-03-26 at 15 47 00
Screenshot 2021-03-26 at 15 36 29

This also updates hustings so that we only display them after they have passed when we have a post event URL to display - this will help cards from becoming clogged up when there are a lot of hustings.


wcivf/apps/hustings/models.py
```diff
@@ -1,14 +1,31 @@
...
+class HustingQueryset(models.QuerySet):
+    def future(self):
+        """
+        Returns QuerySet of objects in the future or today
+        """
+        return self.filter(starts__date__gte=timezone.now().date())
+
+    def displayable(self):
+        """
+        Excludes objects in the past unless we have a postevent url for them
```
symroe, line 21: I don't think this does what the comment says? This will always exclude events in the past, I think? Might need a `Q() | Q()` query?

michaeljcollinsuk:
Screenshot 2021-03-26 at 17 15 36
Example of past hustings when we have a `postevent_url`:

PR #720, opened 2021-03-30
Hotfix/manifestos

VirginiaDooley:
This work includes various manifesto enhancements and person page fixes. Extracting the manifestos into it's own card reduces clutter on the policy card. 

Screen Shot 2021-03-30 at 12 43 32 PM

### DIFF
--- a/wcivf/apps/elections/tests/test_models.py
+++ b/wcivf/apps/elections/tests/test_models.py
@@ -159,11 +159,9 @@ class TestPostElectionModel:
         )
 
     @pytest.mark.django_db
+    @pytest.mark.freeze_time("2021-04-06")
     def test_past_registration_deadline(self, post_election):
-        # election = ElectionWithPostFactory()
-        # election.past()
         post = PostFactory(territory="ENG")
-        # territory = post.territory
         oldest = PostElectionFactory(
             ballot_paper_id="parl.cities-of-london-and-westminster.2019-05-06",
             post=post,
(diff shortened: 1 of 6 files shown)


pmk01:
One thing I should have noticed earlier: can we remove 'UK'?

wcivf/apps/elections/tests/test_models.py
```diff
@@ -158,6 +158,30 @@ def test_friendly_name_mayoral(self, post_election):
...
+        future = PostElectionFactory(
+            ballot_paper_id="parl.cities-of-london-and-westminster.2021-05-06",
+            post=post,
+            election=ElectionFactoryLazySlug(
+                election_date="2021-5-6", current=True, election_type="parl"
+            ),
+        )
+
+        assert oldest.past_registration_deadline is True
+        assert future.past_registration_deadline is False
```
michaeljcollinsuk, line 183: When we get past the registration deadline for this election (Is it 19th April?) this test will fail. We can get around this by using Freezegun (docs for reference https://pypi.org/project/pytest-freezegun/) which we already have installed and are using in other tests, to fix the date at the time the test runs to before the deadline.

wcivf/apps/elections/tests/test_models.py
```diff
@@ -158,6 +158,30 @@ def test_friendly_name_mayoral(self, post_election):
             == "Greater London Authority mayoral election"
         )
 
+    @pytest.mark.django_db
+    def test_past_registration_deadline(self, post_election):
+        # election = ElectionWithPostFactory()
```
michaeljcollinsuk, line 163: Can the commented out lines be deleted?

wcivf/apps/elections/tests/test_models.py
michaeljcollinsuk: Example of other tests where we do this https://github.com/DemocracyClub/WhoCanIVoteFor/blob/64347d36f05df59fd9808eeb885932b4994a1912/wcivf/apps/parishes/tests/test_models.py#L10

Issue #726, opened 2021-04-07
Regional ballot text.


pmk01:
Affects: sp.r, senedd.r, gla.a

eg. https://whocanivotefor.co.uk/elections/sp.r.lothian.2021-05-06/lothian/

We're currently telling people they 'can choose from n candidates'

This is wrong, they pick parties or independent candidates. Ideally the text should read 'n ballot options'.

So the text from the Lothian list should be: 

"You can choose from 19 ballot options in the Lothian region."

VirginiaDooley:
Should this also be the case when the 'candidate list has not been published'? @pmk01

PR #728, opened 2021-04-07
Hotfix/Party ballot count

VirginiaDooley:
Closes https://github.com/DemocracyClub/WhoCanIVoteFor/issues/726

Renders ballot count for a given `postelection` rather than candidate count.

### DIFF
--- a/wcivf/apps/elections/models.py
+++ b/wcivf/apps/elections/models.py
@@ -1,12 +1,13 @@
 import datetime
-from django.utils.functional import cached_property
-import pytz
 import re
 
+import pytz
 from django.conf import settings
 from django.contrib.postgres.fields import JSONField
-from django.urls import reverse
 from django.db import models
+from django.template.defaultfilters import pluralize
+from django.urls import reverse
+from django.utils.functional import cached_property
 from django.utils.html import mark_safe
 from django.utils.text import slugify
 
@@ -467,6 +468,21 @@ class PostElection(models.Model):
             ).latest("election__election_date")
         except PostElection.DoesNotExist:
             return None
+
+    @property
+    def party_ballot_count(self):
+        if self.personpost_set.exists():
+            people = self.personpost_set
+            if self.election.uses_lists:
+                ind_candidates = people.filter(party_id="ynmp-party:2").count()
+                num_other_parties = (
+                    people.values("party_id").distinct().count()
+                    - ind_candidates
+                )
+                total_parties = ind_candidates + num_other_parties
+                return f"{total_parties} ballot options"
+            else:
+                return f"{people.count()} candidate{pluralize(people.count())}"
 
 
 class VotingSystem(models.Model):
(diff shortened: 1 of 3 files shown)


symroe:
Message from the department of "it's a little more complex", I'm afraid!

So a few points:

1. This case only applies to list based elections. This is stored in `Election.uses_lists`. This means we need a new conditional to display different numbers based on this
2. In list elections, parties stand candidates in list order (Person A: 1, person B: 2, etc), but *also* independent candidates can stand and they don't have a list number (or, they're always number 1). so what we need to do here is count all the parties on the ballot and then add all the independent candidates to this number. 

To deal with these two, I expect we should add a method to `PostElection` that returns the number of options that can be selected by a voter. The wording isn't easy either, but I think Peter's suggestion is reasonable.

A small point on the `|length` template filter: This internally is calling `len()` on the object. For a `QuerySet` this in turn calls `count()`, getting the number of rows returned on the database. For a string `len()` is the length of the string.

VirginiaDooley:
@symroe Could I have some feedback on this before I get too deep into writing a test?

wcivf/apps/elections/models.py
```diff
@@ -468,6 +468,16 @@ def next_ballot(self):
...
 
+    @property
+    def party_ballot_count(self):
+        people = self.people
+        independent_candidates = people.filter(party_id="ynmp-party:2").count()
+        num_all_parties = people.values("party_id").distinct().count()
+        if self.election.uses_lists:
+            return len(independent_candidates) + num_all_parties
+        else:
+            return num_all_parties
```
symroe, line 479: This is looking good with some small changes:

1. In the `else` we actually want to just return the number of people (`people.count()`) rather than the number of parties
1. Because of this, `num_all_parties` and `independent_candidates` can be moved inside the if condition

This means that if we have lists in use we count the distinct parties and add the independents, if lists not in use we just count the people.

One untested thing: I think `ynmp-party:2` will be included in `num_all_parties`, so you'll want to exclude that to save getting one more than we should.

wcivf/apps/elections/models.py
```diff
@@ -468,6 +469,23 @@ def next_ballot(self):
...
+                num_other_parties = (
+                    people.values("party_id").distinct().count()
+                    - ind_candidates
+                )
+                total_parties = ind_candidates + num_other_parties
+                return f"{total_parties} ballot options"
+            else:
+                return f"{people.count()} candidate{pluralize(people.count())}"
+        else:
+            return f"{self.election.name}: No candidates known yet."
```
symroe, line 487: Just to make sure - when no candidates are known will this show as `You can choose from {election.name}: No candidates known yet in the {{ postelection.friendly_name }}`?

I don't really think this conditional is useful here - maybe we should just `return None` if we can't tell someone how many choices they will have? I expect this logic is already in the template anyway, so let's reproduce it here.

If we should return something, then we should make the sentence structure the same as the other returned strings

VirginiaDooley:
@symroe You're right, `party_ballot_count` doesn't come into play in the case of no known candidates. 

Earlier on in the template, we have: 

` {% if not postelection.people and postelection.contested %}
          We don't know of any candidates standing yet.
          You can help improve this page: <a href="{{ postelection.ynr_link }}">
          add information about candidates to our database</a>.`

symroe:
Great — I think this is good to go then. Can you squash some of the commits up please?

Issue #719, opened 2021-03-30
Enhance How you vote information for all elections


VirginiaDooley:
### Issue and Steps to Reproduce
https://whocanivotefor.co.uk/elections/local.monmouthshire.st-kingsmark.by.2021-04-01/st-kingsmark/ 
How you vote is currently only enabled for by-elections in (country) when the ballot is locked

### Screenshots

<img width="494" alt="Screen Shot 2021-03-30 at 10 39 15 AM" src="https://user-images.githubusercontent.com/7017118/112968517-48cb0380-9144-11eb-872d-4ac2943ed69c.png">

#### Expected

- [x] Enable How you vote for all elections
- [x] Add link for first past the post
- [x] Customise how to for each election - @pmk01 will advise

pmk01:
There are **three** electoral systems in use on 6 May. We want to modify all of them to some extent.

**First Past the Post** 
In the text 'This election uses First-Past-the-Post', I'd like 'First-past-the-pose' to link to a page on WCIVF [containing this text.](https://docs.google.com/document/d/1njJxHolV3_kp4nnjWltWdyZP1i1zs9FB0H3Sz2WIDMs/edit)

If number of seats = greater than one, we should display the text "Mark an X in the box for your preferred candidates. You can vote for up to [n] candidates in this election".

**Supplementary Vote:**
In the text 'This election uses Supplementary Vote', I'd like 'Supplementary Vote' to link to a page on WCIVF [containing this text.](https://docs.google.com/document/d/1Pqm_xncbSYC-Dl3L0pD43qMmcFO_4UTN9SarlLnmA-8/edit)

I am otherwise happy with the SV text.

**Additional Member System:** This is the most complex and at the moment we don't have any text for it. In the text 'This election uses Additional Member System', I'd like 'Additional Member System' to link to a page on WCIVF [containing this text.](https://docs.google.com/document/d/1luzhcifh1TjNyYeim9PtSlWuxfbskLta7MunHLOT5A4/edit)

After that header, I think we should split it up to show different text for each ballot.

Constituency seats (sp.c, senedd.c, gla.c): 'Mark an X in the box for your preferred candidate.'

Regional seats (sp.r, senedd.r, gla.a): 'Mark an X in the box next to your preferred party or independent candidate.'


Hope this is all clear, it may not be!

VirginiaDooley:
@pmk01 is it important for each explainer text to be located on separate pages or can it go on one page with anchors?

[...]

PR #731, opened 2021-04-08
Feature/how to vote

VirginiaDooley:
Ref https://github.com/DemocracyClub/WhoCanIVoteFor/issues/719

This work enhances how to vote information for each voting system and includes: 

- [x] Text changes to FPTP
- [x] Link to further info on FPTP
- [x] Style FPTP
- [x] Text changes to SV
- [x] Link to further info on SV
- [x] Style SV
- [x] Text changes to AMS
- [x] Link to further info on AMS
- [x] Style AMS
- [x] Regional and Constituency methods
- [x] Regional and Constituency text

Not included in this PR which will be done in another to help expedite roll out: 

- Improved presentation of AMS formula
- Add sample ballots
- Jest test for correct links and content

Links to further info for each voting system, as instructed in the issue
Screen Shot 2021-04-12 at 11 28 51 AM

Each have the same breadcrumbs, DC header and footer
FPTP
Screen Shot 2021-04-12 at 4 15 01 PM

SV
Screen Shot 2021-04-12 at 4 14 50 PM

AMS
Screen Shot 2021-04-12 at 4 14 35 PM

### DIFF
--- a/wcivf/apps/elections/models.py
+++ b/wcivf/apps/elections/models.py
@@ -377,6 +377,14 @@ class PostElection(models.Model):
         return self.ballot_paper_id.startswith("pcc")
 
     @property
+    def is_constituency(self):
+        return self.ballot_paper_id.startswith(("gla.c", "senedd.c", "sp.c"))
+
+    @property
+    def is_regional(self):
+        return self.ballot_paper_id.startswith(("gla.r", "senedd.r", "sp.r"))
+
+    @property
     def friendly_name(self):
         """
         Helper property used in templates to build a 'friendly' name using
@@ -481,3 +489,14 @@ class VotingSystem(models.Model):
     @property
     def uses_party_lists(self):
         return self.slug in ["PR-CL", "AMS"]
+
+    @property
+    def get_absolute_url(self):
+        if self.slug == "FPTP":
+            return reverse("fptp_voting_system_view")
+        elif self.slug == "AMS":
+            return reverse("ams_voting_system_view")
+        elif self.slug == "sv":
+            return reverse("sv_voting_system_view")
+        else:
+            raise ValueError(f"{self.slug} has no URL defined")
--- a/wcivf/apps/elections/templates/elections/includes/_single_ballot.html
+++ b/wcivf/apps/elections/templates/elections/includes/_single_ballot.html
@@ -59,9 +59,8 @@
               {% endif %}
             {% endif %}
 
-            {% if postelection.get_voting_system %}
-              {% include "elections/includes/_how-to-vote.html" with voting_system=postelection.get_voting_system %}
-            {% endif %}
+            {% include "elections/includes/_how-to-vote.html" with voting_system=postelection.get_voting_system %}
+          
           {% else %}
             {# Case: Candidates and the post is NOT locked (add CTA) #}
             The official candidate list has not yet been published.
--- a/wcivf/apps/elections/templates/elections/ams.html
+++ b/wcivf/apps/elections/templates/elections/ams.html
@@ -1,0 +1,196 @@
+{% extends "base.html" %}
+{% load markdown_deux_tags %}
+{% load breadcrumb_tags %}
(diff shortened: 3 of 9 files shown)


wcivf/apps/elections/templates/elections/includes/_single_ballot.html
```diff
@@ -59,9 +59,8 @@ <h2>{% if postelection.is_london_assembly_additional %}Additional members{% else
               {% endif %}
             {% endif %}
 
-            {% if postelection.get_voting_system %}
-              {% include "elections/includes/_how-to-vote.html" with voting_system=postelection.get_voting_system %}
-            {% endif %}
+            {% include "elections/includes/_how-to-vote.html" with voting_system=postelection.get_voting_system %}
```
symroe, line 62: If we're going to rely on this existing in the code, we should also make voting system required at the database level.

I don't think that's for this PR, but could you make an issue to change [this line](https://github.com/DemocracyClub/WhoCanIVoteFor/blob/master/wcivf/apps/elections/models.py#L36) to make it not null?

We'll also need to make sure the importer always adds the voting system at import time. It looks like this is a good chance to improve performance of the importer, as it's [looking up the voting system for each ballot](https://github.com/DemocracyClub/WhoCanIVoteFor/blob/e0d842ac3365fa68702afa4173365fef49372b5b/wcivf/apps/elections/import_helpers.py#L76) at the moment.

wcivf/apps/elections/models.py
```diff
@@ -376,6 +376,14 @@ def is_pcc(self):
         """
         return self.ballot_paper_id.startswith("pcc")
 
+    @property
+    def is_consituency(self):
+        return self.ballot_paper_id.startswith("gla.c", "senedd.c", "sp.c")
```
symroe, line 381: `startswith` needs to be passed a tuple, so `self.ballot_paper_id.startswith(("gla.c", "senedd.c", "sp.c"))` (double braces).

I suspect this silently failing in the template layer at the moment - can you add a test for these properties please? (same for `is_regional` below)

wcivf/apps/elections/models.py
```diff
@@ -489,3 +489,14 @@ def __str__(self):
     @property
     def uses_party_lists(self):
         return self.slug in ["PR-CL", "AMS"]
+
+    @property
+    def explainer_url(self):
```
symroe, line 494: :+1: for being on the model, but given we don't have a `get_absolute_url` for this model, we could just use that? I think this would mean the admin interface links to it (it will add the "view on site" button).

Also, is returning nothing the best thing to do if we have no match for the voting system? It feels like this is something we should enforce at the DB level (see previous comment), and that having no valid voting system should raise an exception

VirginiaDooley: @symroe Are you suggesting to add a get_absolute_url method to this model in place of `explainer_url`? 

👍  on exception

symroe: Yes I think so - it seems to fit the concept of a unique resource for the model, anyway. I think this is just a case of renaming the method :+1:

wcivf/apps/elections/models.py
```diff
@@ -499,4 +499,4 @@ def explainer_url(self):
         elif self.slug == "sv":
             return reverse("sv_voting_system_view")
         else:
-            return ""
+            raise ValueError
```
symroe, line 502: Small point, but it's a good idea to give a message when raising - this ends up getting sent through to sentry / emails / tests / etc and helps debugging.

```raise ValueError(f"{self.slug} has no URL defined")```

VirginiaDooley: Thanks! I was wondering about this.

symroe:
The reason I didn't merge this last night is that `explainer_url` is still used in the template - can you update it to use the new `get_absolute_url` please?

(Also, I wonder why tests didn't fail this?)

symroe (approved):

PR #738, opened 2021-04-12
Remove "not elected" if we don't know the vote count

h2g2bob:
If we don't have a vote count, we don't know if they were elected

eg: https://whocanivotefor.co.uk/person/4412/ian-gilbert was elected in 2015, but shows as "Vote count not available (not elected)"


symroe:
Thanks!

PR #744, opened 2021-04-15
Hotfix/How to vote text edits

VirginiaDooley:
Closes https://github.com/DemocracyClub/WhoCanIVoteFor/issues/748 and https://github.com/DemocracyClub/WhoCanIVoteFor/issues/749
Tidy up following https://github.com/DemocracyClub/WhoCanIVoteFor/pull/731

Screen Shot 2021-04-19 at 12 47 03 PM
Screen Shot 2021-04-19 at 12 46 24 PM
Screen Shot 2021-04-19 at 12 45 55 PM

### DIFF
--- a/wcivf/apps/elections/models.py
+++ b/wcivf/apps/elections/models.py
@@ -494,7 +494,9 @@ class PostElection(models.Model):
                 total_parties = ind_candidates + num_other_parties
                 return f"{total_parties} ballot options"
             else:
-                return f"{people.count()} candidate{pluralize(people.count())}"
+                winner_count = self.winner_count
+                winner_count_pluralized = pluralize(self.winner_count)
+                return f"{winner_count} ballot option{winner_count_pluralized}"
 
     @property
     def should_display_sopn_info(self):
@@ -539,4 +541,4 @@ class VotingSystem(models.Model):
         elif self.slug == "sv":
             return reverse("sv_voting_system_view")
         else:
-            raise ValueError(f"{self.slug} has no URL defined")
+            None
--- a/wcivf/apps/elections/templates/elections/includes/_how-to-vote.html
+++ b/wcivf/apps/elections/templates/elections/includes/_how-to-vote.html
@@ -11,7 +11,7 @@ PR-CL:	Closed List
 
 <details style="padding-top:1em; clear: both;">
   <summary>How you vote</summary>
-  {% if voting_system.slug == "FPTP" or "sv" or "AMS" %}
+  {% if voting_system.get_absolute_url %}
     This election uses <a href="{{ voting_system.get_absolute_url }}">{{ voting_system.name }}</a>.
   {% else %} 
     This election uses {{ voting_system.name }}.
@@ -20,7 +20,7 @@ PR-CL:	Closed List
   {% if voting_system.slug == "FPTP" %}
     <p>Mark an X in the box for your preferred candidate{{ postelection.winner_count|pluralize }}.</p>
     {% if postelection.people|length > 1 %} 
-      <p>You can vote for up to {{ postelection.people|length }} candidates in this election.</p>
+      <p>You can vote for up to {{ postelection.winner_count }} candidate{{ postelection.winner_count|pluralize }} in this election.</p>
     {% endif %}
   {% endif %}
 
--- a/wcivf/apps/elections/templates/elections/includes/_single_ballot.html
+++ b/wcivf/apps/elections/templates/elections/includes/_single_ballot.html
@@ -51,11 +51,11 @@
             {% if postelection.voting_system.slug == "PR-CL" %}
               You will have one vote, and can vote for a single party list or independent candidate.
             {% else %}
+              {% if postelection.winner_count and postelection.election.voting_system_id == 'FPTP' %}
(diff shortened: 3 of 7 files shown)


symroe (approved):
Happy with this, but I understand there might be a couple of other commits coming.

wcivf/apps/elections/templates/elections/includes/_how-to-vote.html
```diff
@@ -12,7 +12,7 @@
 <details style="padding-top:1em; clear: both;">
   <summary>How you vote</summary>
   {% if voting_system.slug == "FPTP" or "sv" or "AMS" %}
-    This election uses <a href="{{ voting_system.get_absolute_url }}">{{ voting_system.name }}</a>.
+    This election uses <a href="{{ voting_system.get_voting_system_url }}">{{ voting_system.name }}</a>.
```
michaeljcollinsuk, line 15: Was this change in relation to https://github.com/DemocracyClub/WhoCanIVoteFor/issues/749? I can't see a `get_voting_system_url` method - so although you won't get a 500 error anymore, that will be because `get_voting_system_url` will silently fail, but the link to the explainer pages will no longer work either.

The cause of the error is the `if` statement on line 14. At the moment the `if` statement will always resolve to True, so we always try to return a URL for an explainer, but for some voting systems where we don't have a page to link to (like STV) the `ValueError` ends up being raised. When you use `or` you need to use the first part of the comparison again each time so it becomes:
```
{% if voting_system.slug == "FPTP" or voting_system.slug == "sv" or voting_system.slug == "AMS" %}
```

But I would also change the `get_absolute_url` method on VotingSystem to return `None` or an empty string instead of raising a ValueError:
```
    @property
    def get_absolute_url(self):
        if self.slug == "FPTP":
            return reverse("fptp_voting_system_view")
        elif self.slug == "AMS":
            return reverse("ams_voting_system_view")
        elif self.slug == "sv":
            return reverse("sv_voting_system_view")
        return None
```
And then the if statement in the template can be changed to instead check if we have a URL to link to or not:
```
{% if voting_system.get_absolute_url %}
```

VirginiaDooley: Thanks, all that makes sense.

wcivf/apps/elections/models.py
```diff
@@ -494,7 +494,9 @@ def party_ballot_count(self):
                 total_parties = ind_candidates + num_other_parties
                 return f"{total_parties} ballot options"
             else:
-                return f"{people.count()} candidate{pluralize(people.count())}"
+                winner_count = self.winner_count
+                winner_count_pluralized = pluralize(self.winner_count)
+                return f"{winner_count} ballot option{winner_count_pluralized}"
```
michaeljcollinsuk, line 499: I thought in this case it should be `candidate` rather than `ballot option`, as it is only list elections that you have ballot options? Might be worth checking with @symroe

wcivf/apps/elections/templates/elections/includes/_single_ballot.html
```diff
@@ -51,11 +51,11 @@ <h2>{% if postelection.is_london_assembly_additional %}Additional members{% else
...
               You will have one vote, and can vote for a single party list or independent candidate.
             {% else %}
-              You can choose from <strong>{{ postelection.party_ballot_count }}</strong>
-              in the {{ postelection.friendly_name }}.
               {% if postelection.winner_count and postelection.election.voting_system_id == 'FPTP' %}
-                You will have {{ postelection.winner_count|intword }} vote{{ postelection.winner_count|pluralize }}.
+                You will have up to {{ postelection.winner_count|intword }} vote{{ postelection.winner_count|pluralize }}.
               {% endif %}
+              You can choose from <strong>{{ postelection.party_ballot_count }}</strong>
+              in the {{ postelection.friendly_name }}.
```
michaeljcollinsuk, line 58: Struggling to follow the nested if statements a little but it looks like this line is always displayed, is that intended? So if it was FTPT you would get both this and the lines on 55.

VirginiaDooley: Yes, approved by @pmk01

wcivf/apps/elections/models.py
VirginiaDooley: Checked w @pmk01, all good.

michaeljcollinsuk (approved):

wcivf/apps/elections/templates/elections/includes/_how-to-vote.html
symroe: I'm a little confused by this!

I think we can keep the name `get_absolute_url` if we're returning `None` now. Also, if we're returning `None then we can simplify the conditional in the template, and simply do `{% if voting_system.get_absolute_url %}` - that way when we add a new explainer we don't need to change the template in order to link to it.

VirginiaDooley: We already have that name in the election model.

symroe: Also, sorry about suggesting the `ValueError` in the previous version - I had assumed we had an explainer for every known voting system, so an exception would be useful to us!

Issue #521, opened 2020-02-10
Ensure list based elections are displayed properly on WCIVF


symroe:
e.g., currently https://whocanivotefor.co.uk/elections/SW1A%201AA/  is wrong.

This is because the `uses_lists` flag isn't set properly, I expect.

symroe:
I think this might be done, but should double check that it's all ok!

PR #747, opened 2021-04-16
Hotfix/person intro

michaeljcollinsuk:
This started out as a quick fix to change the intro text on candidates running in Mayoral elections as per https://docs.google.com/document/d/1xRgw441CIBvYo3YdwNIqejxaRhl_BUseZ4ucSGB-tUw/edit#... but turned in to something a bit bigger.

- Removes `get_intro` method and its logic from the PersonView
- Replaced with property method on the Person model which uses template files and `render_to_string`
- This should make it much easier to change the intro text on a case by case basis - as each type of election can have their own HTML include that the intro string can be built from
- Adds lots of tests to attempt to make sure we havent broken anything in the process. I still think that this should be deployed to test site initially for a good click through

### DIFF
--- a/wcivf/apps/elections/import_helpers.py
+++ b/wcivf/apps/elections/import_helpers.py
@@ -32,7 +32,8 @@ class YNRElectionImporter:
             election_type = slug.split(".")[0]
 
             election_weight = 10
-            if ballot_dict["election"]["party_lists_in_use"]:
+            uses_lists = ballot_dict["election"]["party_lists_in_use"]
+            if uses_lists:
                 election_weight = 20
             if election_type == "mayor":
                 election_weight = 5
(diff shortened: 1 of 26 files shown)


michaeljcollinsuk:
I have now improved intros for other election types and this should finally be ready for final review and merge  - some screenshots below that @pmk01 may want to double check. Also closes #521 

Screenshot 2021-04-20 at 12 56 23
Screenshot 2021-04-20 at 12 55 45
Screenshot 2021-04-20 at 12 55 27
Screenshot 2021-04-20 at 12 54 34
Screenshot 2021-04-20 at 12 28 07

pmk01:
RECOVER THRIVE SANDY

I'm going to go and fix that, hahaha.

Looks good! All read well.

symroe:
This looks good, I'll merge and chuck on a stage site for looking at in a browser :+1:

PR #751, opened 2021-04-19
Bugfix/remove cvs

michaeljcollinsuk:
- We no longer display CV's or import then, so remove final links left in the code base that [are causing errors](https://sentry.io/organizations/democracy-club-gp/issues/2298915585/?project=1426221&query=is%3Aunresolved) and adds data migration to delete any existing CV objects. 
- Also remove old commented out leaflet links


symroe:
This PR is good, but once it's merged and deployed I think we should do another to remove the app completely from the project. We can't do it before, as it won't clean up the database if we just delete it.

Issue #753, opened 2021-04-19
Reinstate feedback form


VirginiaDooley:
We want to capture feedback from our users to improve WCIVF. 

1. Get the form working properly. The main thing is that we need to show a simple yes/no question first with an expanding form after.
2. The answer to the first question is always submitted, then it's updated with any follow up. See here: https://github.com/DemocracyClub/WhoCanIVoteFor/blob/master/wcivf/apps/feedback/static/feedback/js/feedback_form.js#L
3. Make sure the model is storing all the questions
4. Set up the crontab lines on the who_deploy repo: https://github.com/DemocracyClub/WhoCanIVoteFor/blob/master/wcivf/apps/feedback/static/feedback/js/feedback_form.js#L5

PR #754, opened 2021-04-19
Hotfix/Feedback form

VirginiaDooley:
Closes https://github.com/DemocracyClub/WhoCanIVoteFor/issues/753
Co-authored with @michaeljcollinsuk 
Ref https://trello.com/c/5zm6BsBK/2478-put-some-good-research-questions-on-wdiv-wcivf-to-replace-did-you-find-what-you-were-looking-for

This work includes: 
- [x] Reinstated the form
- [x] Added a new question 
- [x] Updated the model to store new question
- [x] Set up the crontab lines on the who_deploy repo https://github.com/DemocracyClub/who_deploy/pull/88
- [x] Added new question to admin

Screen Shot 2021-04-21 at 11 56 44 AM
Screen Shot 2021-04-21 at 12 10 57 PM


wcivf/apps/feedback/templates/feedback/feedback_form.html
```diff
@@ -13,6 +13,14 @@ <h2>
       {{ choice.tag }}
       <label class="link-button" for="{{ choice.id_for_label }}">{{ choice.choice_label }}</label>
     {% endfor %}
+    
+    <div class="comments_container">
+      <p>
+        What other websites/sources have you used to 
+        research the election?
+      </p> 
+        {{ feedback_form.comments }}
```
michaeljcollinsuk, line 22: I think this should be `feedback_form.sources`

wcivf/apps/feedback/forms.py
```diff
@@ -12,11 +12,12 @@ def __init__(self, *args, **kwargs):
...
     class Meta:
         model = Feedback
-        fields = ["found_useful", "comments", "source_url"]
+        fields = ["found_useful", "sources", "comments", "source_url"]
 
     found_useful = forms.ChoiceField(
         choices=FOUND_USEFUL_CHOICES,
         widget=forms.RadioSelect(attrs={"data-toggle": "button"}),
     )
+    sources = forms.CharField(widget=forms.HiddenInput())
```
michaeljcollinsuk, line 21: Not sure you want this to be a hidden field? In fact, as you have added `sources` to the `fields` list in the Meta, there is no need for this line at all - Django will render know to render the field as a TextInput as it is defined on the model as a `TextField`

wcivf/apps/feedback/templates/feedback/feedback_form.html
VirginiaDooley: That's what I assumed too, but the text box didn't appear. And that's because I had it hidden as you mention below. Thanks!

symroe:
This looks good to me!

The only thing I changed was to comment out the form on the [elections and post page](https://github.com/DemocracyClub/WhoCanIVoteFor/commit/093116548e3d99cf5cc8ce698ad35bbb2085c293).

This is because we used to get loads of spam from those forms, and it's less common from a page where you have to enter a postcode before seeing it.

Issue #763, opened 2021-04-22
Parties or independent candidate[s]


VirginiaDooley:
In [`party_ballot_count`](https://github.com/DemocracyClub/WhoCanIVoteFor/blob/ea4b0cc747e3058aaf1e4c86e9b56a87042fa3a2/wcivf/apps/elections/models.py#L494), for `uses_lists`, change the response to read: 

"**n** Parties or independent candidate[s]"

Issue #764, opened 2021-04-22
Audit unused ds-scope divs after the design system is implemented site wide


VirginiaDooley:

VirginiaDooley:
Closed n https://github.com/DemocracyClub/WhoCanIVoteFor/pull/823 and https://github.com/DemocracyClub/WhoCanIVoteFor/pull/852

Issue #762, opened 2021-04-21
Implement ds-sidebar for Independent candidates on people lists 


VirginiaDooley:
https://test.whocanivotefor.co.uk/elections/SA1%205HF/

![Screen Shot 2021-04-21 at 10 13 06 PM](https://user-images.githubusercontent.com/7017118/115621722-d38bc200-a2ee-11eb-9a0e-672e938d8558.png)

VirginiaDooley:
Fixed Feature/Implement design system #772

Issue #761, opened 2021-04-21
"You can choose from N ballot options" number is wrong?


chris48s:
I'm not really sure what this number is trying to tell me. I guess it is either supposed to be either:

- The total number of things (people or parties) that will be on the ballot paper or
- The number of those things (people or parties) that I can vote for in some way

but the number doesn't correspond to either of those things. For example:

This is a Supplementary Vote ballot with 6 people standing. I can vote for 2 of them (first and second preference):

![Screenshot at 2021-04-21 17-25-05](https://user-images.githubusercontent.com/6025893/115588738-f5714e80-a2c6-11eb-9c1b-06e35ac0a28a.png)

---

This is a PR ballot with 14 total things standing (13 parties plus one independent candidate). I can vote for one of them:

![Screenshot at 2021-04-21 17-25-25](https://user-images.githubusercontent.com/6025893/115588779-01f5a700-a2c7-11eb-816e-51f7dcc007ea.png)

Not sure if the feature is broken or working and I don't understand what it is trying to tell me. If its the second one, I suspect it is unclear to others too and this could be phrased more clearly.

symroe:
I think we might have fixed this already, but not deployed it yet. Can you check on https://test.whocanivotefor.co.uk and confirm if it's improved?

chris48s:
Yes - that does appear to be fixed now and more clearly phrased :+1: 

This definitely shouldn't be a blocker to deploying that because it is a big improvement but...

The one further minor suggestion I would have is: On party list elections the caption is something like "You will have one vote, and can choose from 10 **candidates** ", whereas actually you've got way more candidates standing than that. That's the number of [parties or people]. Obviously this is difficult because parties can field a list and independents can also stand too so you've got 10.. things there isn't a single nice word for, but it would probably be clearer to say "You will have one vote, and can choose from 10 parties" (accepting that in some case that means 10 parties and in some cases it means 8 parties and 2 independents, or whatever). I think that would be a better explanation and a fairly simple change to make.

The really nice solution would be to actually present that as  "You will have one vote, and can choose from 8 parties and 2 independent candidates" or whatever where that is the case (which you do have the data for), but y'know.. priorities :)

sjorford:
This may be a stupid question, but what wording does the ballot paper use?

chris48s:
Definitely not a stupid question - I had not thought to look until you mentioned it, although having done so I don't think it provides a neat answer :laughing: 

[...]

PR #765, opened 2021-04-22
Hotfix/voting edits

VirginiaDooley:
Closes https://github.com/DemocracyClub/WhoCanIVoteFor/issues/763 and https://github.com/DemocracyClub/WhoCanIVoteFor/issues/762
ref https://github.com/DemocracyClub/WhoCanIVoteFor/issues/761

I did not remove extra `ds-scope` tags because there are a few templates that have yet to receive the design system classes that would be affected. [New issue](https://github.com/DemocracyClub/WhoCanIVoteFor/issues/764) to audit after these tags after full implementation. 

Screen Shot 2021-04-22 at 11 17 09 AM

### DIFF
--- a/wcivf/apps/elections/models.py
+++ b/wcivf/apps/elections/models.py
@@ -500,10 +500,10 @@ class PostElection(models.Model):
                     people.values("party_id").distinct().count()
                     - ind_candidates
                 )
-                total_parties = ind_candidates + num_other_parties
-                total_parties_apnumber = apnumber(total_parties)
-                options_pluralized = pluralize(total_parties)
-                return f"{total_parties_apnumber} option{options_pluralized}"
+                ind_and_parties = ind_candidates + num_other_parties
+                ind_and_parties_apnumber = apnumber(ind_and_parties)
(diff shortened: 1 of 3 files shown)


symroe:
This is great, thanks!

I did a couple of things before merging:

1. [I put "independent candidates" in a conditional](https://github.com/DemocracyClub/WhoCanIVoteFor/commit/578241949595284869b8331df392eb0fa0972450), as we know when we have independents or not

2. I noticed that we were running _loads_ of queries on the postcode view. A London postcode was performing over 4,000 queries and running very slowly. This was because we were performing new queries in the templates for objects we had already. It's all too easy to do this in Django, and it [didn't take a lot to fix it](https://github.com/DemocracyClub/WhoCanIVoteFor/commit/78730fed49bbeabf084f85f85317b89b1c25d920). This is something that's worth keeping an eye on in development.

BTW, I spotted the queries thing using Django Debug Toolbar. It's enabled in the project, but I think you need to add the following to `local.py` to see it:

```py
INTERNAL_IPS = [
    '127.0.0.1',
]
```

Issue #766, opened 2021-04-22
Implement Design System 


VirginiaDooley:
- [ ] Home page
- [ ] Header
- [ ] Nav & Breadcrumbs https://github.com/DemocracyClub/WhoCanIVoteFor/issues/697
- [ ] Footer
- [x] Elections
- [ ] Feedback
- [ ] Hustings
- [ ] Leaflets
- [ ] News
- [ ] Parishes
- [x] Parties 
- [x] People
- [ ] Templates linked from home page footer

PR #772, opened 2021-04-27
Feature/Implement design system

VirginiaDooley:
Ref https://github.com/DemocracyClub/WhoCanIVoteFor/issues/766

Closes
https://github.com/DemocracyClub/WhoCanIVoteFor/issues/762
https://github.com/DemocracyClub/WhoCanIVoteFor/issues/780

This work is part of a multi-step process that converts classes to the design system and makes general UX improvements users would find in the postcode search journey 
It also:
- Temporarily hides upcoming elections on the home page
- Adds instructions for users who want to manually subscribe to our the election calendar. This is a temp fix until we can switch ical library. 
- Links to post in Person intro

_Not included in this PR

- Convert forms
- Convert markdown pages
- Remove scope
- Implement filter for all Elections_

Screen Shot 2021-04-28 at 9 55 50 AM

Screen Shot 2021-04-28 at 12 38 17 PM

Screen Shot 2021-04-28 at 12 56 14 PM


symroe:
I think this is looking great!

One small thing, and I'm not sure this is introduced in this PR anyway: can you make sure that `ds-card-link` is added as a class on person cards? This turns the [whole card in to a link](https://democracyclub.github.io/design-system/components/card/), rather than just the text.  

Otherwise I think this is ready to deploy :tada:

VirginiaDooley:
> I think this is looking great!
> 
> One small thing, and I'm not sure this is introduced in this PR anyway: can you make sure that `ds-card-link` is added as a class on person cards? This turns the [whole card in to a link](https://democracyclub.github.io/design-system/components/card/), rather than just the text.
> 
> Otherwise I think this is ready to deploy 🎉

Added! RTG

PR #824, opened 2021-06-04
Feature/party name

michaeljcollinsuk:
We were already storing the `party_name` and `party_description_text` against a candidacy (Membership) in YNR - this applies the same changes, on the `PersonPost` model. This came up as @pmk01 noticed that we were still displaying Reform UK for Brexit Party candidates before the name change this year.

For example, with these changes [this ballot from 2019 with a Brexit Party candidate](https://whocanivotefor.co.uk/elections/parl.stalybridge-and-hyde.2019-12-12/) will no longer display Reform UK as the party:
Screenshot 2021-06-04 at 15 46 02

- Adds new fields
- Data migration to populate the `party_name` so that we don't need to run a fill import
- Update the ballot importer to capture these in future fields when importing from YNR
- Update templates to read party name from the candidacy
- Make the intro templates more readable by keeping them to a 70 char word length (after finally understanding when/why a new line in a template is rendered with whitespace, [second answer down on this SO post](https://stackoverflow.com/questions/588356/why-does-the-browser-renders-a-newline-as-space))


symroe (approved):
This looks good to me!

PR #846, opened 2021-07-14
Feature/Validate html

VirginiaDooley:
Ref DemocracyClub/dc_django_utils#6

This work finds and corrects html errors found through `tidy` available through `dc_utils`.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -20,8 +20,8 @@ django-debug-toolbar==3.2.1
 newspaper3k==0.2.8
 
 git+https://github.com/DemocracyClub/dc_signup_form.git@2.1.0
-git+https://github.com/DemocracyClub/design-system.git@0.1.5
-git+https://github.com/DemocracyClub/dc_django_utils.git@0.0.5
+git+https://github.com/DemocracyClub/design-system.git@0.1.4
+git+https://github.com/DemocracyClub/dc_django_utils.git@0.0.7
 djangorestframework-jsonp==1.0.2
 feedparser==6.0.8
 
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -20,7 +20,7 @@ django-debug-toolbar==3.2.1
 newspaper3k==0.2.8
 
 git+https://github.com/DemocracyClub/dc_signup_form.git@2.1.0
-git+https://github.com/DemocracyClub/design-system.git@0.1.4
+git+https://github.com/DemocracyClub/design-system.git@0.1.5
 git+https://github.com/DemocracyClub/dc_django_utils.git@0.0.7
 djangorestframework-jsonp==1.0.2
 feedparser==6.0.8
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,11 @@ jobs:
             - ./.venv
             - /home/circleci/repo/.tox/
           key: v2-dependencies-{{ checksum "requirements/base.txt" }}-{{ checksum "requirements/testing.txt" }}
-
+      
+      - run:
+          name: Install HTML Tidy
+          command: wget https://github.com/htacg/tidy-html5/releases/download/5.8.0/tidy-5.8.0-Linux-64bit.deb && sudo dpkg -i tidy-5.8.0-Linux-64bit.deb
+      
       # run tests!
       - run:
           name: run tests
--- a/wcivf/apps/core/tests/test_frontend.py
+++ b/wcivf/apps/core/tests/test_frontend.py
@@ -5,12 +5,19 @@ Used for making sure meta tags and important information is actually
 shown before and after template changes.
 """
 
-
 from django.test import TestCase
 from django.test.utils import override_settings
 from django.urls import reverse
+from dc_utils.tests.helpers import validate_html
+from people.tests.factories import PersonFactory, PersonPostFactory
+from parties.tests.factories import PartyFactory
+from elections.tests.factories import (
+    ElectionFactory,
+    PostElectionFactory,
+)
 
 import vcr
+import pytest
 
 
 @override_settings(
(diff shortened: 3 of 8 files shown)


michaeljcollinsuk:
@VirginiaDooley re: the current build failure - if you had `pre commit` configured to run Black that would have been caught before the push. I'm guessing the install was removed at some point when we were deleting/rebuilding env, but it should just be a case of running `pre-commit install` https://github.com/DemocracyClub/WhoCanIVoteFor/blob/master/INSTALL.md#code-formatting

wcivf/tests/test_urls.py
```diff
@@ -0,0 +1,69 @@
...
+            reverse(
+                "email_person_view",
+                kwargs={"pk": PersonFactory().pk, "ignored_slug": "sadiq-khan"},
+            ),
+            reverse("parties_view"),
+            PartyFactory().get_absolute_url(),
+            reverse("elections_view"),
+            ElectionFactory().get_absolute_url(),
+            reverse(
+                "redirect_election_view",
```
michaeljcollinsuk, line 35: Not sure if you have seen but the tests are failing attempting to build this URL https://app.circleci.com/pipelines/github/DemocracyClub/WhoCanIVoteFor/1026/workflows/7da6c9e9-407d-45a9-aa08-963d6e132b70/jobs/1849. Do you have `pre-commit` setup (check the readme) as if you did I would have expected the test to fail locally before pushing?

Although I dont think you need to test this url as its purpose seems to be to return a redirect to the ElectionView, which is already tested on line 33. One that is missing however is a test for `PostElectionView` - you could test this by using the `PostElectionFactory().get_absolute_url()`

wcivf/tests/test_urls.py
```diff
@@ -0,0 +1,69 @@
...
+                kwargs={
+                    "election": ElectionFactory(),
+                    "ignored_slug": ElectionFactory().slug,
+                },
+            ),
+            reverse("postcode_view", kwargs={"postcode": "TE11ST"}),
+            reverse(
+                "redirect_post_view",
```
michaeljcollinsuk, line 43: Similar to the above, I dont think we should test this view as it will not return HTML, it will return a response with HTML content, I think it will always return a redirect to the [PostElectionView](https://github.com/DemocracyClub/WhoCanIVoteFor/blob/911562510a3bc9363b46afbfe46680c392f66b5b/wcivf/apps/elections/views/election_views.py#L67). So as long as we have a set to cover the `PostElectionView` directly (mentioned in previous comment) we dont need this.

wcivf/tests/test_urls.py
```diff
@@ -0,0 +1,69 @@
...
+            reverse("ams_voting_system_view"),
+        ]
+
+    @pytest.mark.django_db
+    def test_html_valid(self, client, subtests, urls):
+        for url in urls:
+            with subtests.test(msg=url):
+                import pdb
+
+                pdb.set_trace()
```
michaeljcollinsuk, line 65: Needs to be deleted

michaeljcollinsuk (changes requested):

wcivf/tests/test_urls.py
```diff
@@ -0,0 +1,51 @@
...
+            reverse("sv_voting_system_view"),
+            reverse("fptp_voting_system_view"),
+            reverse("ams_voting_system_view"),
+        ]
+
+    @pytest.mark.django_db
+    def test_html_valid(self, client, subtests, urls):
+        for url in urls:
+            with subtests.test(msg=url):
+                assert client.get(url).status_code == 200 or 301
```
michaeljcollinsuk, line 48: Are some of these urls still returning a 301 status code? If so, then those should be removed rather than changing the assert here. As remember the aim of this test is to run checks against the HTML  - if a response is a 301, it won't contain any HTML so we shouldn't test it. 

Unless we are getting a 301 where we expect a 200 status code. In that case we may be getting redirected because something is going wrong with the objects that we are setting up.

michaeljcollinsuk:
One final comment is about the structure/location of these tests. I dont think it is right to add a new `tests` directory with this file in. As the tests are not specific to a single app, I think adding them to the `core` app would be best - there is already a [test_frontend.py](https://github.com/DemocracyClub/WhoCanIVoteFor/blob/master/wcivf/apps/core/tests/test_frontend.py) file that you could add your `TestHTML` class to, rather than a new file.

wcivf/tests/test_urls.py
VirginiaDooley: `/privacy` is redirecting to the DC website. I assumed in that case a 301 is expected, but now that I look again, `/about` (which is also redirecting) isn't generating a 301.

VirginiaDooley: When I print out the content for `/privacy`, this is what I get: 
`b''`

michaeljcollinsuk: Oh sorry either a 301 or 302 (both redirect codes) we shouldn't test as the response won't include any HTML to validate

requirements/base.txt
```diff
@@ -20,8 +20,8 @@ django-debug-toolbar==3.2.1
 newspaper3k==0.2.8
 
 git+https://github.com/DemocracyClub/dc_signup_form.git@2.1.0
-git+https://github.com/DemocracyClub/design-system.git@0.1.5
-git+https://github.com/DemocracyClub/dc_django_utils.git@0.0.5
+git+https://github.com/DemocracyClub/design-system.git@0.1.4
```
michaeljcollinsuk, line 23: Is this version drop to 0.1.4 intended?

VirginiaDooley: No, but I didn't think it should go here.

michaeljcollinsuk: [The increase to 0.1.5 already got merged to master ](https://github.com/DemocracyClub/WhoCanIVoteFor/blob/master/requirements/base.txt#L23) so this is actually a change down a version, which would bring back the logo error. Changing back to 0.1.5 here will stop the line showing up as a change.

michaeljcollinsuk (approved):

PR #848, opened 2021-07-22
Hotfix/social media preview

VirginiaDooley:
This work updates the Open Graph image used in social media post previews as well as adds the favicons lost with the removal `dc_base_theme`.


symroe (approved):
Yep, this works. I think we should add a png version of the logo to the design system repo, so we can use that single file. Otherwise we'll have to copy this image file to all repos.

VirginiaDooley:
> Yep, this works. I think we should add a png version of the logo to the design system repo, so we can use that single file. Otherwise we'll have to copy this image file to all repos.

That makes sense, will do.

PR #851, opened 2021-07-29
WIP Feature/aws lambda

michaeljcollinsuk:


wcivf/aws_lambda/handlers.py
```diff
@@ -24,7 +24,12 @@ def lambda_handler(event, context):
 
         Return doc: https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html
     """
-
+    print("EVENT BELOW")
+    print(event)
```
symroe: By the way, did you know about `print(f"{event=}")`?

wcivf/settings/base.py
```diff
@@ -121,7 +121,7 @@
         "ENGINE": "django.db.backends.postgresql_psycopg2",
         "NAME": "wcivf",
         "USER": os.environ.get("DB_USER", "wcivf"),
-        "PASSWORD": os.environ.get("DB_PASSWORD", ""),
+        "PASSWORD": os.environ.get("DB_PASSWORD", "wcivf"),
```
symroe: We might want this in a Lambda settings file that we can import from - this will mean local installs will need a password too

wcivf/aws_lambda/handlers.py
michaeljcollinsuk: Ah yes this would be a smarter way of doing it!

symroe:
:tada:

Issue #773, opened 2021-04-27
Improve UX for all Elections page


VirginiaDooley:
Implement filter component for elections list
https://democracyclub.github.io/design-system/components/filter/ for https://whocanivotefor.co.uk/elections/

PR #858, opened 2021-08-17
Feature/filter elections

VirginiaDooley:
Ref https://github.com/DemocracyClub/WhoCanIVoteFor/issues/773

This work

- [x] implements the design system filter buttons
- [x] filters by election type
- [x] hides dates when no elections exist for that date
- [x] fixes filter highlighting; toggles aria-attribute
https://dev.wcivf.club/elections/


wcivf/apps/elections/templates/elections/elections_view.html
```diff
@@ -128,7 +140,20 @@ <h2>Past Elections</h2>
             }
         });
 
-
+        function checkHideDate() {
+            for (var i = 0; i < date_grouper.length; i++) {
+                all_hidden = true;
+                date_grouper_urls = date_grouper[i].querySelectorAll('li a');
+                for (var j = 0; j < date_grouper_urls.length; j++) {
+                    if (date_grouper_urls[j].style.display == '' && date_grouper_urls[j] in local) {
```
VirginiaDooley, line 148: Stuck. When I click on a filter button (ie "local") and there are no local elections on a given date, I want to hide that date. And how to do this dynamically based on the filter that is clicked. 🤔

wcivf/apps/elections/templates/elections/elections_view.html
```diff
@@ -4,9 +4,26 @@
 {% block og_title %}{% trans "UK Elections" %}{% endblock og_title %}
 {% block og_description %}{% trans "All Elections in the UK" %}{% endblock og_description %}
 
+
 {% block content %}
 
     {% include 'elections/includes/_elections_breadcrumbs.html' %}
+    {% comment %} {% if javascript_enabled %}   {% endcomment %}
```
michaeljcollinsuk, line 11: Is this meant to be commented out?

wcivf/apps/elections/templates/elections/includes/_election_list.html
```diff
@@ -1,14 +1,16 @@
...
         {{ election_group.grouper }}
     </h3>
-    {% for election in election_group.list %}
-        <ul style="list-style: none;">
-            <li>
-                <a href="{{ election.get_absolute_url }}">{{ election.nice_election_name }}</a>
-            </li>
-        </ul>
-    {% endfor %}
+    <div id="election-urls" >
```
michaeljcollinsuk, line 7: Should this be using `class` as I thought that an HTML `id` was meant to be unique?

wcivf/apps/elections/templates/elections/elections_view.html
michaeljcollinsuk: Hmm not sure if this is possible, but could you loop through every H3 with the `grouper-date` class, and look for the next `election-urls` div and check if there is a `<li>` element with the display set, and if there are not any, set the display for that `h3` element to none? Not sure how difficult that conditional is to do in JS but feels like something along those lines might work

michaeljcollinsuk:
One general comment, when I click one of the filter buttons, it doesn't highlight it as selected, the focus stays on the "all" button

wcivf/apps/elections/templates/elections/includes/_election_list.html
VirginiaDooley: You're right, thanks for catching.

wcivf/apps/elections/templates/elections/elections_view.html
VirginiaDooley: Removed

wcivf/apps/elections/templates/elections/elections_view.html
```diff
@@ -27,4 +42,237 @@ <h2>{% trans "Past Elections" %}</h2>
...
+            checkHideDate()
+        });
+
+        document.getElementById("mayor").addEventListener("click", function(e){
+            e.preventDefault();
+            setAriaAttribute(e);
+
+            for (var i = 0; i < all.length; i++) {
+                if (all[i].href.includes('mayor.')) {
+                    all[i].style.display = "block";
```
michaeljcollinsuk, line 204: Can you DRY up this repeat logic? e.g. use a `hideOrDisplay` function?

symroe:
I think this is a decent enhancement to this page!

We're going to want to think broadly across DC about an election browsing interface (applies to YNR too), but this will let us explore what we want from that interface, I think.

Ideally longer term this would be implemented in the backend maybe as well as the front end. That is, it should work without JavaScript using djang-filter, but with a JS layer on top.

Let's keep it in JS at the moment though.

Given that, I have some feedback:

1. Can you add the filter elements using JavaScript? As things stand the buttons show up but do nothing if JS isn't enabled. It would be better to only add the feature if it can be used.
2. The query strings created should be prefixed ideally. That is, rather than `?gla` they should be `?election_type=gla`. This would allow us to add more filters in the future that could be combined. It would also mean when we add a backend filter we can maintain the same URLs
3. I think as Michael said, can we DRY up the logic? We should be able to add a loop over all the buttons and apply a handler to them in a single function, rather than a new function per filter.

Thinking more about these together, are you able to define the filters in an object and then use code to create the buttons? Something like:

```javascript
election_type_filters = {
  gla: "London Assembly",
  pcc: "Police &amp; Crime Commissioner"
}

for election_type, label in election_type_filters { # Is this even how you JavaScript?!
  # add element
  # add handler for the element
}

```

I don't know exactly how to implement the single method for each election type, but I expect you could data-attributes to get the type from the element (or maybe the ID is enough?) and then show/hide based on that?

michaeljcollinsuk:
> I don't know exactly how to implement the single method for each election type, but I expect you could data-attributes to get the type from the element (or maybe the ID is enough?) and then show/hide based on that?

I had a similar thought about this, and I think that if you move the `id` to the `a` element rather than the `<li>` then you can get the ID of the clicked button from your listener e.g.
```
var clickedId = e.target.id;
```
Then you can use that ID with `document.getElementById` and your conditional where you check `href.includes`

VirginiaDooley:
> > I don't know exactly how to implement the single method for each election type, but I expect you could data-attributes to get the type from the element (or maybe the ID is enough?) and then show/hide based on that?
> 
> I had a similar thought about this, and I think that if you move the `id` to the `a` element rather than the `<li>` then you can get the ID of the clicked button from your listener e.g.
> 
> ```
> var clickedId = e.target.id;
> ```
> 
> Then you can use that ID with `document.getElementById` and your conditional where you check `href.includes`

Thanks for the suggestion. The only case that I'm not sure how to handle with this approach is when the senned filter should include ballot ids with  "naw."

VirginiaDooley:
> > > I don't know exactly how to implement the single method for each election type, but I expect you could data-attributes to get the type from the element (or maybe the ID is enough?) and then show/hide based on that?
> > 
> > 
> > I had a similar thought about this, and I think that if you move the `id` to the `a` element rather than the `<li>` then you can get the ID of the clicked button from your listener e.g.
> > ```
> > var clickedId = e.target.id;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then you can use that ID with `document.getElementById` and your conditional where you check `href.includes`
> 
> Thanks for the suggestion. The only case that I'm not sure how to handle with this approach is when the senned filter should include ballot ids with "naw."

Resolved :)

VirginiaDooley:
> I think this is a decent enhancement to this page!
> 
> We're going to want to think broadly across DC about an election browsing interface (applies to YNR too), but this will let us explore what we want from that interface, I think.
> 
> Ideally longer term this would be implemented in the backend maybe as well as the front end. That is, it should work without JavaScript using djang-filter, but with a JS layer on top.
> 
> Let's keep it in JS at the moment though.
> 
> Given that, I have some feedback:
> 
> 1. Can you add the filter elements using JavaScript? As things stand the buttons show up but do nothing if JS isn't enabled. It would be better to only add the feature if it can be used.
> 2. The query strings created should be prefixed ideally. That is, rather than `?gla` they should be `?election_type=gla`. This would allow us to add more filters in the future that could be combined. It would also mean when we add a backend filter we can maintain the same URLs
> 3. I think as Michael said, can we DRY up the logic? We should be able to add a loop over all the buttons and apply a handler to them in a single function, rather than a new function per filter.
> 
> Thinking more about these together, are you able to define the filters in an object and then use code to create the buttons? Something like:
> 
> ```js
> election_type_filters = {
>   gla: "London Assembly",
>   pcc: "Police &amp; Crime Commissioner"
> }
> 
> for election_type, label in election_type_filters { # Is this even how you JavaScript?!
>   # add element
>   # add handler for the element
> }
> ```
> 
> I don't know exactly how to implement the single method for each election type, but I expect you could data-attributes to get the type from the element (or maybe the ID is enough?) and then show/hide based on that?

Done. Although this was a good exercise, I can completely see the argument for doing this in the backend. I had to double check IE support on several occasions and often had to find an alternative solution. Not worth it. Happy to either refactor without JS now or deploy this and do it next.

VirginiaDooley:
Completed with https://github.com/DemocracyClub/WhoCanIVoteFor/pull/1236

PR #867, opened 2021-09-09
Skip to content text is missing in link

danjordan:
I noticed the skip to content text is missing on https://whocanivotefor.co.uk/ but is present on https://democracyclub.org.uk/ so have added some in.


symroe:
Thanks @danjordan!

PR #868, opened 2021-09-10
Feature/croydon referendum

michaeljcollinsuk:
Adds a management command to create a referendum. Creates an Election and PostElection object so that it is treated the same as other ballots when user enters a postcode, or views the election list etc. But I have also tried to utilise the existing Referendum model to use those fields to store data specific to the referendum, and then use that as a hook for how we display the information. Not sure if it is a bit hacky as it increases the logic in templates for example, but it seems to work better than the previous implementation we used for referendums.

One thing I am a little unsure about is if a `Post` is still relevant for a referendum, and if we are likely to have an existing object we can use or if they might need to be created? For now I have enforced adding one in the management command, as the data model forces a relationship between `PostElection` and `Post`.

Some screenshots:
Screenshot 2021-09-10 at 16 10 56

Screenshot 2021-09-10 at 16 11 22

Screenshot 2021-09-10 at 16 11 30

Screenshot 2021-09-10 at 16 11 34


symroe:
> One thing I am a little unsure about is if a Post is still relevant for a referendum, and if we are likely to have an existing object we can use or if they might need to be created?

In WCIVF, the `Post` model is really only used as a way to link elections in the same place. That is, "what elections have happened in this place before". I don't think there's another election type that uses the whole of Croydon as a `Post` anywhere (e.g, we'd see this if they had a directly elected mayor), so I think making one ad-hoc here is reasonable.

pmk01:
This all looks really good to me! Font style/scale is really clear. 

My only comment is grammatical, as the 'read more' link is a bit wordy. Could cut a couple for something like "Find (or 'Read'?) more information on the Croydon council website."

Also, why is it 'Croydon Central' in that last screenshot? is that pulling in a constituency name from somewhere?

michaeljcollinsuk:
@symroe updated now so that a new `Post` is created rather than relying on an existing one.

@pmk01 the "Croydon Central" label was related to what Sym commented about the Post - the breadcrumbs use the name of the Post (division) related to the ballot. I've changed this now, along with your suggested text alteration:
Screenshot 2021-09-13 at 11 42 33

pmk01:
Great stuff! 🎈

PR #895, opened 2021-09-29
Pin django above 3.2

michaeljcollinsuk:
Django 3.1 introduced their implementation of JSONField, which we are
now using in the application. So we need to pin the django version to
at least 3.1 otherwise migrations will not run. Pinning to >=3.2 as
this is the LTS version.


VirginiaDooley (approved):
👍

PR #939, opened 2021-11-02
Remove import_votes_cast

michaeljcollinsuk:
Results will be updated by the ballot importer looking for recent
changes. This is because when a result is added/changed in YNR, this
triggers will cause the ballot to appear as 'updated'.


michaeljcollinsuk:
@symroe I've had a look at the `import_results_atom` management command, and I dont think it should be enabled at the moment as I think the whole results app it needs a rework if we did want to use it ([the results list page is certainly quite broken](https://whocanivotefor.co.uk/results/)). I think there had been some talk of removing the app entirely in the past, but I could leave in place and just remove the URL if we think we might want to reuse it later?

symroe (approved):
Given this new system is importing the `elected` boolean and the votes cast, in near real time, I think it's reasonable to bin `import_results_atom` too.

I can see that we might want really fast importing of results overnight for a `parl` election in the future, but that might be as simple as running the ballots importer every minute or something. Either way, I don't think that command is helping us.

Same for the results page: happy for you to bin that. Can be in this PR if you like, or a new one.

The work in the PR is good to go: up to you as to if you want to carry on or merge.

PR #948, opened 2021-11-05
Update Party to store url of emblem

michaeljcollinsuk:
- Avoid writing the emblem file to disk so that the import_parties
management command can be run in AWS lambda
- Stores the URL of the emblem which is stored by YNR is s3 bucket
- Updates templates and other code to use the url directly rather than
the emblem object


michaeljcollinsuk:
Tested by deploying these changes to the stage environment and running the importer - completed in under 30 seconds, image displays like:
https://wcivf.club/parties/party:210/21st-century-conservative-democrat
vs
Party emblems broken currently: https://whocanivotefor.co.uk/parties/party:210/21st-century-conservative-democrat

symroe (approved):

Issue #931, opened 2021-10-29
Add text when a ballot is uncontested


michaeljcollinsuk:
### Issue and Steps to Reproduce
If a ballot is uncontested, on the ballot page ([example of uncontested ballot here](https://whocanivotefor.co.uk/elections/local.ceredigion.llandyfriog.2017-06-08/llandyfriog/)) we should add text to make this clear. Suggested text from @pmk01 :

> Uncontested Election
>
> This election was uncontested because only one candidate came forward to stand for each available seat.
>
>No votes will be cast, and the candidate/s below has/have been automatically declared the winner/s.

Then on the person page, we should update the results table to say "Elected unopposed" instead of "Vote count not available (elected)" ([example here](https://whocanivotefor.co.uk/person/26670/james-wyn-reynolds-thomas))

sjorford:
It is possible for there to be fewer candidates than seats e.g. [local.north-kesteven.billinghay-martin-and-north-kyme.2019-05-02](https://candidates.democracyclub.org.uk/elections/local.north-kesteven.billinghay-martin-and-north-kyme.2019-05-02/) where only one candidate ran for a two-seat ward. So the wording could read:

>Uncontested Election
>This election was uncontested because the number of candidates was less than or equal to the number of seats.
>No votes will be cast, and the candidate/s below has/have been automatically declared the winner/s.

chris48s:
In the case where one candidate stands for a two seat ward, presumably there is some extra complexity there because the candidate who stood gets one of the seats but then there's an immediate vacancy (and hence by-election) for the other seat - right? So I guess in the example there https://candidates.democracyclub.org.uk/elections/local.north-kesteven.billinghay-martin-and-north-kyme.by.2019-06-13/ is the corresponding by-election for the second vacancy

sjorford:
Yes, that's right - the by-election has to be held within 35 working days of the original election date ([RoPA 1983 39(1)](https://www.legislation.gov.uk/ukpga/1983/2/section/39))

pmk01:
Hmm, thanks for pointing this out. I think we'd want separate text to cover these eventualities (I'm concerned that "less than or equal to" would create unnecessary confusion). I also think we'd want to tell people that the election will be rescheduled.

Perhaps:

[...]

Issue #884, opened 2021-04-10
Wording for contested parish elections 


sjorford:
If the election is contested and we have a link to the SOPN PDF, the top card still says "There _may_ also be parish, town or community council elections in some areas." - this can be removed in this case

e.g. https://whocanivotefor.co.uk/elections/SY3%205HA/
![image](https://user-images.githubusercontent.com/10633357/114266765-edc0c880-99ef-11eb-9b1f-d38d1e57671f.png)

PR #970, opened 2021-11-23
Feature/Update uncontested explanation

VirginiaDooley:
Closes https://github.com/DemocracyClub/WhoCanIVoteFor/issues/931 and https://github.com/DemocracyClub/WhoCanIVoteFor/issues/884

This PR includes text and logic updates for greater clarity on uncontested, cancelled, & rescheduled elections. 

In the case of fewer candidates than seats but where number of candidates is one or more  (when no metadata in EE):
Screen Shot 2022-01-18 at 12 24 16 PM

When there are equal candidates to seats  (and no metadata in EE) 
Screen Shot 2022-01-17 at 12 07 50 PM

If metadata exists, it overrides the above text:
Screen Shot 2022-01-20 at 11 28 18 AM

If no one shows up:
Screen Shot 2022-01-20 at 11 27 00 AM

Cancelled for any other reasons
Screen Shot 2022-01-18 at 12 27 11 PM

### DIFF
--- a/wcivf/apps/elections/import_helpers.py
+++ b/wcivf/apps/elections/import_helpers.py
@@ -315,8 +315,9 @@ class YNRBallotImporter:
                 "locked": ballot_dict["candidates_locked"],
             }
 
-            if ballot_dict["candidates_locked"] and ballot_dict["winner_count"]:
-                defaults["contested"] = not ballot_dict["uncontested"]
+            if ballot_dict["candidates_locked"] or ballot_dict["cancelled"]:
+                if ballot_dict["winner_count"]:
+                    defaults["contested"] = not ballot_dict["uncontested"]
 
             # only update this when using the recently_updated flag as otherwise
             # the timestamp will only be the modifed timestamp on the ballot
--- a/wcivf/apps/elections/import_helpers.py
+++ b/wcivf/apps/elections/import_helpers.py
@@ -315,7 +315,7 @@ class YNRBallotImporter:
                 "locked": ballot_dict["candidates_locked"],
             }
 
-            if ballot_dict["candidates_locked"]:
+            if ballot_dict["candidates_locked"] and ballot_dict["winner_count"]:
                 defaults["contested"] = (
                     len(ballot_dict["candidacies"])
                     > ballot_dict["winner_count"]
--- a/wcivf/apps/elections/import_helpers.py
+++ b/wcivf/apps/elections/import_helpers.py
@@ -315,6 +315,12 @@ class YNRBallotImporter:
                 "locked": ballot_dict["candidates_locked"],
             }
 
+            if ballot_dict["candidates_locked"]:
+                defaults["contested"] = (
+                    len(ballot_dict["candidacies"])
+                    > ballot_dict["winner_count"]
+                )
+
             # only update this when using the recently_updated flag as otherwise
             # the timestamp will only be the modifed timestamp on the ballot
             # see BallotSerializer.get_last_updated in YNR
--- a/wcivf/apps/elections/import_helpers.py
+++ b/wcivf/apps/elections/import_helpers.py
@@ -316,10 +316,7 @@ class YNRBallotImporter:
             }
 
             if ballot_dict["candidates_locked"] and ballot_dict["winner_count"]:
-                defaults["contested"] = (
-                    len(ballot_dict["candidacies"])
-                    > ballot_dict["winner_count"]
-                )
+                defaults["contested"] = not ballot_dict["uncontested"]
 
             # only update this when using the recently_updated flag as otherwise
             # the timestamp will only be the modifed timestamp on the ballot
--- a/wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
+++ b/wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
@@ -27,9 +27,6 @@
             {{ object.replaced_by.election.election_date|date:"j F Y" }}
         </a>
     {% endif %}
-    {% if object.metadata.cancelled_election.detail %}
-        {{ object.metadata.cancelled_election.detail }}
-    {% endif %}
     {% if object.metadata.cancelled_election.url %}
         <a href="{{ object.metadata.cancelled_election.url }}">Read more</a>
     {% endif %}
--- a/wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
+++ b/wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
@@ -10,7 +10,7 @@
             <b>Uncontested Election</b>
 
             <p>This election was uncontested because the number of candidates who stood was equal to the number of available seats.
-                There is{{ object.winner_count|pluralize:"are" }} {{ object.winner_count|apnumber }}
+                There {{ object.winner_count|pluralize:"is,are" }} {{ object.winner_count|apnumber }}
                 seat{{ object.winner_count|pluralize }} in this ward/division, and only
                 {{ object.people.count|apnumber }} candidate{{ object.people|pluralize }}.</p>
 
@@ -21,7 +21,7 @@
                 <b>Uncontested and Rescheduled Election</b>
 
                 <p>This election was uncontested because the number of candidates who stood was fewer than the number of available seats.
-                    There is{{ object.winner_count|pluralize:"are"}} {{ object.winner_count|apnumber }}
+                    There {{ object.winner_count|pluralize:"is,are"}} {{ object.winner_count|apnumber }}
                     seat{{ object.winner_count|pluralize }} in this ward/division, and only
                     {{ object.people.count|apnumber }} candidate{{ object.people|pluralize }}.</p>
 
--- a/wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
+++ b/wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
@@ -15,7 +15,16 @@
                 {{ object.people.count|apnumber }} candidate{{ object.people|pluralize }}.</p>
 
             <p>No votes will be cast, and the candidate{{ object.people|pluralize }} below have been automatically declared
-                the winner{{ object.people|pluralize }}.
+                the winner{{ object.people|pluralize }}.</p>
+
+            {% if postelection.people %}
+                {% if postelection.display_as_party_list %}
+                    {% include "elections/includes/_people_list_with_lists.html" with people=postelection.people %}
+                {% else %}
+                    {% include "elections/includes/_people_list.html" with people=postelection.people %}
+                {% endif %}
+            {% endif %}
+
         {% else %}
             {% if object.people.count >= 1 %}
                 <b>Uncontested and Rescheduled Election</b>
@@ -30,6 +39,13 @@
                     winner{{ object.people|pluralize }}.
                     A new election to fill the unclaimed seat{{ object.people|pluralize }}
                     will be held within 35 working days of the original election date.</p>
+                {% if postelection.people %}
+                    {% if postelection.display_as_party_list %}
+                        {% include "elections/includes/_people_list_with_lists.html" with people=postelection.people %}
+                    {% else %}
+                        {% include "elections/includes/_people_list.html" with people=postelection.people %}
+                    {% endif %}
+                {% endif %}
             {% else %}
                 <b>Rescheduled Election</b>
 
--- a/wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
+++ b/wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
@@ -2,10 +2,12 @@
 {% load humanize %}
 {% load static %}
 
+{% comment %}Case 1: Cancelled election and Metadata is set in EE {% endcomment %}
 {% if object.metadata.cancelled_election.title %}
     <h4>{{ object.metadata.cancelled_election.title }}</h4>
 {% else %}
     {% if not object.contested %}
+        {% comment %} Case 2: Election cancelled, uncontested, number of candidates equal seats, no metadata{% endcomment %}
         {% if object.winner_count == object.people.count %}
             <h4>Uncontested Election</h4>
 
@@ -22,19 +24,21 @@
             {% else %}
                 {% include "elections/includes/_people_list.html" with people=object.people %}
             {% endif %}
-
+            {% comment %} Case 3: Election cancelled, uncontested and there are fewer candidates than seats, no metadata {% endcomment %}
         {% elif object.winner_count > object.people.count %}
+
+            <h4>Uncontested and Rescheduled Election</h4>
+
+            <p>This election was uncontested because the number of candidates who stood was fewer than the number of available seats.
+                There {{ object.winner_count|pluralize:"is,are"}} {{ object.winner_count|apnumber }}
+                seat{{ object.winner_count|pluralize }} in {{ object.post.full_label }}, and
+                {{ object.people.count|apnumber }} candidate{{ object.people|pluralize }}.</p>
+            {% comment %} Case 4: Election cancelled, uncontested and zero candidates, no metadata {% endcomment %}
             {% if object.people %}
-                <h4>Uncontested and Rescheduled Election</h4>
-
-                <p>This election was uncontested because the number of candidates who stood was fewer than the number of available seats.
-                    There {{ object.winner_count|pluralize:"is,are"}} {{ object.winner_count|apnumber }}
-                    seat{{ object.winner_count|pluralize }} in {{ object.post.full_label }}, and only
-                    {{ object.people.count|apnumber }} candidate{{ object.people|pluralize }}.</p>
-
                 <p>No votes will be cast, and the candidate{{ object.people|pluralize }} below
                     ha{{ object.people|pluralize:"s,ve" }} been automatically declared the
                     winner{{ object.people|pluralize }}.
+
                     A new election to fill the unclaimed seat{{ object.people|pluralize }}
                     will be held within 35 working days of the original election date.</p>
 
@@ -43,11 +47,9 @@
                 {% else %}
                     {% include "elections/includes/_people_list.html" with people=object.people %}
                 {% endif %}
-            {% elif object.people.count > object.winner_count %}
-                <h4>Cancelled Election</h4>
-                <p>This election was cancelled.</p>
             {% endif %}
         {% endif %}
+        {% comment %} Case 5: Contested but cancelled for other reasons {% endcomment %}
     {% else %}
         <h4>Cancelled Election</h4>
         <p>This election was cancelled.</p>
@@ -70,7 +72,6 @@
     {% if object.metadata.cancelled_election.url %}
         <a href="{{ object.metadata.cancelled_election.url }}">Read more</a>
     {% endif %}
-
 </p>
 
 <p>
--- a/wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
+++ b/wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
@@ -1,22 +1,46 @@
 {% load markdown_deux_tags %}
+{% load humanize %}
+{% load static %}
+
 {% if object.metadata.cancelled_election.title %}
     <h2>{{ object.metadata.cancelled_election.title }}</h2>
 {% else %}
-    <h2>Cancelled Election</h2>
+    <div>
+        {% if object.winner_count == object.people.count %}
+            <b>Uncontested Election</b>
+
+            <p>This election was uncontested because the number of candidates who stood was equal to the number of available seats.
+                There is{{ object.winner_count|pluralize:"are" }} {{ object.winner_count|apnumber }}
+                seat{{ object.winner_count|pluralize }} in this ward/division, and only
+                {{ object.people.count|apnumber }} candidate{{ object.people|pluralize }}.</p>
+
+            <p>No votes will be cast, and the candidate{{ object.people|pluralize }} below have been automatically declared
+                the winner{{ object.people|pluralize }}.
+        {% else %}
+            {% if object.people.count >= 1 %}
+                <b>Uncontested and Rescheduled Election</b>
+
+                <p>This election was uncontested because the number of candidates who stood was fewer than the number of available seats.
+                    There is{{ object.winner_count|pluralize:"are"}} {{ object.winner_count|apnumber }}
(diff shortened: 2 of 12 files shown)


VirginiaDooley:
May be worth adding https://github.com/DemocracyClub/WhoCanIVoteFor/issues/884 with this PR?

michaeljcollinsuk:
Do these changes display locally? As if so they may not have deployed correctly, because on dev I can't see any difference https://dev.wcivf.club/elections/local.ceredigion.llandyfriog.2017-06-08/llandyfriog/

michaeljcollinsuk:
> Do these changes display locally? As if so they may not have deployed correctly, because on dev I can't see any difference https://dev.wcivf.club/elections/local.ceredigion.llandyfriog.2017-06-08/llandyfriog/ EDIT I think this is actually because the `contested` field on the PostElection isn't being updated. We may be better changing this to a property method that counts the number of seats up and the number of elected candidates

@VirginiaDooley pushed a couple of commits to help with this, which in turn will help with checking the changes in on the development site https://dev.wcivf.club/elections/local.ceredigion.llandyfriog.2017-06-08/llandyfriog/

Locally the easiest thing to do to test is to open a shell, lookup a ballot, and change its `contested` field to `False`

wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
```diff
@@ -8,6 +8,15 @@ <h2>Cancelled Election</h2>
     The poll for {{ object.post.label }} in the {{ object.election.name }}
     {% if object.election.in_past %}was{% else %}has been{% endif %}
     {% if object.replaced_by %}rescheduled{% else %}cancelled{% endif %}.
+
```
VirginiaDooley, line 11: I'm not quite sure about "Rescheduled Election" as a header - whether I can use the logic `{% if object.metadata.cancelled_election.title %}` or if not, the condition that I need to set.

pmk01:
Yes, rescheduled replaces current text. there are three possible pieces of text we should display depending on circumstances: 

Uncontested
Uncontested and Rescheduled Election*
Rescheduled election

*reading back over my text I realise there's a typo/unnecessary words in this one. Opening sentence should read:

**This election was uncontested because the number of candidates is fewer than the number of available seats.**

(sorry!)

VirginiaDooley:
@pmk01 Currently we have uncontested logic & text for when an election is cancelled or not in two separate templates. Should the text in each situation you've outline apply whether an election is cancelled or not?

pmk01:
Yes, all the scenarios will result in a rescheduled poll. 'Cancelled' is just a button we press... In an ideal world we wouldn't have to press the button, the system would identify these automatically.

wcivf/apps/elections/tests/test_election_views.py
```diff
@@ -141,6 +141,107 @@ def test_num_candidates(self):
...
+            "This election was uncontested because the number of candidates who stood was equal to the number of available seats.",
+        )
+        self.assertContains(
+            response,
+            "No votes will be cast, and the candidate below has been automatically declared the winner.",
+        )
+        self.assertNotContains(response, "This election was cancelled.")
+
+    # In the case where the number of candidates is equal to seats but then the candidate dies
+    def test_cancelled_candidate_died(self):
```
VirginiaDooley, line 218: @symroe I'm still not clear on this. Do we want a specific message in this case?

wcivf/apps/elections/tests/test_election_views.py
```diff
@@ -141,6 +141,146 @@ def test_num_candidates(self):
             response, f"See all 5 candidates in the {self.post.label}"
         )
 
+    def test_cancelled_election_with_candidates(self):
+        self.post_election.winner_count = 4
```
VirginiaDooley, line 145: 😕 What am I missing in the setup that would cause the second assert to fail?

wcivf/apps/elections/tests/test_election_views.py
```diff
@@ -141,6 +141,117 @@ def test_num_candidates(self):
...
+            response, "elections/includes/_post_meta_title.html"
+        )
+        self.assertTemplateUsed(
+            response, "elections/includes/_post_meta_description.html"
+        )
+        self.assertContains(response, "Cancelled Election")
+        self.assertNotContains(response, "No candidates known yet.")
+
+    # In the case of fewer candidates than seats but where number of candidates is one or more
+    def test_uncontested_and_rescheduled(self):
```
VirginiaDooley, line 189: And here?

wcivf/apps/elections/tests/test_election_views.py
```diff
@@ -141,6 +141,117 @@ def test_num_candidates(self):
...
+        self.post_election.winner_count = 5
+        people = [PersonFactory() for p in range(4)]
+        for person in people:
+            PersonPostFactory(
+                post_election=self.post_election,
+                election=self.election,
+                post=self.post,
+                person=person,
+            )
+        self.post_election.uncontested = True
```
michaeljcollinsuk, line 199: There is no `uncontested` field on `PostElection` think this should be `self.post_election.contested = False`

michaeljcollinsuk: https://github.com/DemocracyClub/WhoCanIVoteFor/blob/master/wcivf/apps/elections/models.py#L342

wcivf/apps/elections/templates/elections/includes/_post_meta_description.html
```diff
@@ -1,10 +1,14 @@
...
-        See all {{ postelection.people|length }} candidates in the {{ object.election.name }} on {{ object.election.election_date }}:
-        {% for pp in postelection.people %}
-            {{ pp.person.name|safe }} ({{ pp.party_name }})
-        {% endfor %}
-    {% endautoescape %}
+{% if postelection.cancelled %}
+    {{ object.election.name }}: This election has been cancelled.
 {% else %}
-    {{ object.election.name }}: No candidates known yet.
+    {% if postelection.people|length %}
```
michaeljcollinsuk, line 4: Could avoid the nested `if` statements by using `elif`
```
{% if postelection.cancelled %}
     {{ object.election.name }}: This election has been cancelled.
{% elif postelection.people|length %}
.....
{% else %}
.....
{% endif %}
```

wcivf/apps/elections/tests/test_election_views.py
```diff
@@ -141,6 +141,117 @@ def test_num_candidates(self):
...
+        self.assertTemplateUsed(response, "elections/post_view.html")
+        self.assertTemplateUsed(
+            response, "elections/includes/_post_meta_title.html"
+        )
+        self.assertTemplateUsed(
+            response, "elections/includes/_post_meta_description.html"
+        )
+        self.assertContains(response, "Cancelled Election")
+        self.assertNotContains(response, "No candidates known yet.")
+
```
michaeljcollinsuk, line 169: You can assert your meta changes worked correctly with:
```
self.assertContains(response, f"{self.post_election.election.name}: This election has been cancelled")

File: wcivf/apps/elections/templates/elections/includes/_single_ballot.html [hunk 8]
```diff
@@ -6,7 +6,7 @@
     <div class="ds-card" id="election_{{ postelection.election.slug }}">
         <div class="ds-card-body">
 
-            {% if postelection.cancelled %}
+            {% if postelection.cancelled and postelection.contested %}
```
VirginiaDooley, line 9: Not sure about this one.

wcivf/apps/elections/templates/elections/includes/_single_ballot.html
```diff
@@ -123,10 +123,10 @@ <h3>{% if postelection.is_london_assembly_additional %}Additional members{% else
                                     winner{{ postelection.people|pluralize }}.
                                     A new election to fill the unclaimed seat{{ postelection.people|pluralize }}
                                     will be held within 35 working days of the original election date.</p>
-                            {% else %}
+                            {% elif postelection.people.count == 0 %}
```
VirginiaDooley, line 126: I added this to be explicit, but recognise that an `elif` without an `else` doesn't work. Will refactor with fresh eyes. 👀

wcivf/apps/elections/tests/test_election_views.py
```diff
@@ -256,7 +256,8 @@ def test_no_candidates_turn_up(self):
         )
         self.assertContains(
             response,
-            "This election will not take place",
+            f"This election will not take place because <a href='{self.post_election.ynr_link}'>no candidates have been nominated to stand</a>.",
```
VirginiaDooley, line 259: I'm struggling to find an example of this type of assert which checks for the presence of a url. What am I missing here?

symroe (commented):
Just realised I did a review but didn't publish the comments :face_exhaling: Doing that now, in case it's useful!

VirginiaDooley:
Blocked until https://github.com/DemocracyClub/yournextrepresentative/pull/1684 is merged.

VirginiaDooley:
Blocked until https://github.com/DemocracyClub/yournextrepresentative/pull/1684 is merged.

wcivf/apps/elections/import_helpers.py
```diff
@@ -315,6 +315,12 @@ def add_ballots(self, results):
                 "locked": ballot_dict["candidates_locked"],
             }
 
+            if ballot_dict["candidates_locked"] and ballot_dict["winner_count"]:
```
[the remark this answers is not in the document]
VirginiaDooley: Do you mean `if defaults["contested"] != not ballot_dict["contested"]`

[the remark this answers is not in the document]
michaeljcollinsuk: @VirginiaDooley I've exposed `uncontested` on the API [in this PR which I will merge and deploy](https://github.com/DemocracyClub/yournextrepresentative/pull/1686)

Then in here, this line will need changing to (think it was a typo in original comment):
```
defaults["contested"] = not ballot_dict["uncontested"]
```

VirginiaDooley:
Remaining issue is the pluralisation with 'is' vs 'are' which I think can be addressed with `ngettext` https://docs.djangoproject.com/en/4.0/topics/i18n/translation/#pluralization

wcivf/apps/elections/templates/elections/includes/_single_ballot.html
```diff
@@ -6,43 +6,46 @@
     <div class="ds-card" id="election_{{ postelection.election.slug }}">
         <div class="ds-card-body">
 
-            {% if postelection.cancelled %}
```
michaeljcollinsuk, line 9: Unfortunately I think the change to remove this and move it further down to line 96 has had the knock on effect of breaking some historic elections e.g. those cancelled due to covid in 2020. See the difference between the [live site](https://whocanivotefor.co.uk/elections/gla.a.2020-05-07/london-assembly-elections-additional/) and [dev site](https://dev.wcivf.club/elections/gla.a.2020-05-07/london-assembly-elections-additional/)

I think the reason this change was made was so that the candidates were still listed when a ballot was uncontested? So I wonder if this should be reverted, and then instead we take

VirginiaDooley: I made this change as suggested, but then lost the date and summary line at the top: "X candidates stood in the ..." since it's now included in the `else` (not cancelled). I'll look again tomorrow with fresh eyes to avoid further duplication. 👀

wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
```diff
@@ -1,27 +1,81 @@
...
+{% load humanize %}
+{% load static %}
+
 {% if object.metadata.cancelled_election.title %}
     <h2>{{ object.metadata.cancelled_election.title }}</h2>
 {% else %}
-    <h2>Cancelled Election</h2>
+    <div>
+        {% if object.winner_count == object.people.count %}
+            <b>Uncontested Election</b>
```
symroe, line 10: A note on the `b` tag. It's got a bit of a complex history, where it was used for "bold" in the early days, then not used at all ("strong" replaced it) and now it's back in a different form to mean "bring to attention". 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b

I can see the case for using "bring to attention" here, but the above link suggests not using it for headings. 

Because of that, I thin this should be a heading under the heading level of the ballot title (`h4`?). If the size looks wrong, we can use the `.ds-h[n]` classes to change that.

wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
```diff
@@ -1,27 +1,81 @@
 {% load markdown_deux_tags %}
+{% load humanize %}
+{% load static %}
+
 {% if object.metadata.cancelled_election.title %}
     <h2>{{ object.metadata.cancelled_election.title }}</h2>
 {% else %}
-    <h2>Cancelled Election</h2>
+    <div>
```
symroe, line 8: Just out of interest, why the empty `div` here? Is it to do with spacing / stacking?

wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
```diff
@@ -1,27 +1,81 @@
...
+                the winner{{ object.people|pluralize }}.</p>
+
+            {% if object.display_as_party_list %}
+                {% include "elections/includes/_people_list_with_lists.html" with people=object.people %}
+            {% else %}
+                {% include "elections/includes/_people_list.html" with people=object.people %}
+            {% endif %}
+
+        {% else %}
+            {% if object.people.count >= 1 %}
```
symroe, line 27: I think this can just be `if object.people`? But just to make sure, is this only shown when we've set the cancelled flag?

wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
```diff
@@ -1,27 +1,81 @@
...
+                <b>Uncontested and Rescheduled Election</b>
+
+                <p>This election was uncontested because the number of candidates who stood was fewer than the number of available seats.
+                    There {{ object.winner_count|pluralize:"is,are"}} {{ object.winner_count|apnumber }}
+                    seat{{ object.winner_count|pluralize }} in this ward/division, and only
```
symroe, line 32: One for @michaeljcollinsuk, but I think we can use `object.post.division_suffix` to replace "ward/division" here.

wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
```diff
@@ -1,27 +1,81 @@
...
+                    will be held within 35 working days of the original election date.</p>
+
+                {% if object.display_as_party_list %}
+                    {% include "elections/includes/_people_list_with_lists.html" with people=object.people %}
+                {% else %}
+                    {% include "elections/includes/_people_list.html" with people=object.people %}
+                {% endif %}
+
+            {% else %}
+                <b>Rescheduled Election</b>
```
symroe, line 48: As above, this should be a heading not a `b`

wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
VirginiaDooley: It's not doing anything here, will remove.

wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
VirginiaDooley: Ballot name is `h2` but this also made me realise that I need to move the ballot name out of the conditional block.

pmk01:
One small point which might be addressed elsewhere: on the Ashford example here, the text should read 'has', not 'have'. But we'd want 'have' for more than one winner (and a plural on 'winner' in that case).

wcivf/apps/elections/templates/elections/includes/_single_ballot.html
```diff
@@ -97,22 +98,12 @@ <h3>{% if postelection.is_london_assembly_additional %}Additional members{% else
...
-                        This election isn't contested. That means that there are the same number of candidates nominated
-                        as there are seats up for election, so they all get positions without an election taking place.
-                    </div>
+                {% if object.display_as_party_list %}
+                    {% include "elections/includes/_people_list_with_lists.html" with people=object.people %}
+                {% else %}
+                    {% include "elections/includes/_people_list.html" with people=object.people %}
                 {% endif %}
 
-                {% if postelection.people %}
```
michaeljcollinsuk, line 107: I think this check has been deleted in error?

wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
VirginiaDooley: I used `full_label` instead to handle cases where a `division_type` has not been set.

VirginiaDooley:
> One small point which might be addressed elsewhere: on the Ashford example here, the text should read 'has', not 'have'. But we'd want 'have' for more than one winner (and a plural on 'winner' in that case).

@pmk01 Thanks for catching that one; screenshots are now up to date.

michaeljcollinsuk:
> Do these changes display locally? As if so they may not have deployed correctly, because on dev I can't see any difference https://dev.wcivf.club/elections/local.ceredigion.llandyfriog.2017-06-08/llandyfriog/ EDIT I think this is actually because the `contested` field on the PostElection isn't being updated. We may be better changing this to a property method that counts the number of seats up and the number of elected candidates

This is from ages ago, but I'm pretty sure this got resolved by this YNR PR https://github.com/DemocracyClub/yournextrepresentative/pull/1684?

VirginiaDooley:
This election is making me reconsider whether my logic is correct: [


This is a cancelled electiont: local.renfrewshire.paisley-southeast.by.2021-12-14](https://candidates.democracyclub.org.uk/elections/local.renfrewshire.paisley-southeast.by.2021-12-14/)
Screen Shot 2022-01-18 at 8 41 49 AM

VirginiaDooley:
> This election is making me reconsider whether my logic is correct: https://candidates.democracyclub.org.uk/elections/local.renfrewshire.paisley-southeast.by.2021-12-14/ Screen Shot 2022-01-18 at 8 41 49 AM

Bug found! Fixed in https://github.com/DemocracyClub/yournextrepresentative/pull/1708

wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
```diff
@@ -1,27 +1,81 @@
...
+            <p>No votes will be cast, and the candidate{{ object.people|pluralize }} below ha{{ object.winner_count|pluralize:"s,ve" }} been automatically declared
+                the winner{{ object.people|pluralize }}.</p>
+
+            {% if object.display_as_party_list %}
+                {% include "elections/includes/_people_list_with_lists.html" with people=object.people %}
+            {% else %}
+                {% include "elections/includes/_people_list.html" with people=object.people %}
+            {% endif %}
+
+        {% elif object.winner_count > object.people.count %}
```
VirginiaDooley, line 26: This logic won't work if we want the "Uncontested and Rescheduled Election" text to render in the case of a no candidates & uncontested.

michaeljcollinsuk: Could you move the `{% if object.people %}` check further down to where it is used? E.g.:
```
{% elif object.winner_count > object.people.count %}
    <h4>Uncontested and Rescheduled Election</h4>

     <p>This election was uncontested because the number of candidates who stood was fewer than the number of available seats.<p>
     {% if object.people %}
         ....
```

wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
```diff
@@ -1,27 +1,81 @@
...
+                    winner{{ object.people|pluralize }}.
+                    A new election to fill the unclaimed seat{{ object.people|pluralize }}
+                    will be held within 35 working days of the original election date.</p>
+
+                {% if object.display_as_party_list %}
+                    {% include "elections/includes/_people_list_with_lists.html" with people=object.people %}
+                {% else %}
+                    {% include "elections/includes/_people_list.html" with people=object.people %}
+                {% endif %}
+            {% elif object.people.count > object.winner_count %}
```
michaeljcollinsuk, line 46: As it stands I dont think these lines will ever be hit because of line 27, so think it can be removed?

VirginiaDooley: If there are 2 seats and only one person shows, they would be elected and displayed in a people list. Then line 46 is just an `else`?

symroe:
I think one thing that's making this hard to review (and write!) is that there are loads of cases to think about all at once.

Being kind to future us, I think we should do two things:

1. Add in _loads_ of comments to the conditionals, explaining how each case comes up. Give these cases numbers (Case 1, 2, 2.1, etc)
2. Make sure we have a test for each case that uses the same number system as the comments (in a doc string). We can then add the test that covers each case to the template comment to create a two way link.

At that point, we _should_ be able to review this by looking more at the tests than the template code, but also, the next time we have to deal with this it'll be a whole lot easier.

VirginiaDooley:
> I think one thing that's making this hard to review (and write!) is that there are loads of cases to think about all at once.
> 
> Being kind to future us, I think we should do two things:
> 
> 1. Add in _loads_ of comments to the conditionals, explaining how each case comes up. Give these cases numbers (Case 1, 2, 2.1, etc)
> 2. Make sure we have a test for each case that uses the same number system as the comments (in a doc string). We can then add the test that covers each case to the template comment to create a two way link.
> 
> At that point, we _should_ be able to review this by looking more at the tests than the template code, but also, the next time we have to deal with this it'll be a whole lot easier.

Great suggestion, this feels so much clearer now. Let me know what you think.

michaeljcollinsuk:
> > I think one thing that's making this hard to review (and write!) is that there are loads of cases to think about all at once.
> > Being kind to future us, I think we should do two things:
> > 
> > 1. Add in _loads_ of comments to the conditionals, explaining how each case comes up. Give these cases numbers (Case 1, 2, 2.1, etc)
> > 2. Make sure we have a test for each case that uses the same number system as the comments (in a doc string). We can then add the test that covers each case to the template comment to create a two way link.
> > 
> > At that point, we _should_ be able to review this by looking more at the tests than the template code, but also, the next time we have to deal with this it'll be a whole lot easier.
> 
> Great suggestion, this feels so much clearer now. Let me know what you think.

This helped! LGTM

michaeljcollinsuk (approved):

symroe (approved):

PR #1083, opened 2022-03-04
Welsh translations

symroe:
An initial implementation of Welsh translations. Basically just the bootstrapping.

Screenshot 2022-03-04 at 12-33-08 Pwy alla i bleidleisio ar gyfer

The idea is that none of this will be shown if `USE_I18N==False`. I would like to get this PR merged soon, so we can add `trans` tags through the code in stages from now on, rather than having a large PR hanging about that might cause problems with other PRs.


michaeljcollinsuk (approved):
🏴󠁧󠁢󠁷󠁬󠁳󠁿

PR #1099, opened 2022-03-17
Add leaflets and hustsings to API

symroe:
TODO on this: optimize the leaflets query. We're currently making one query per person on the returned. We want to only return the latest 4 leaflets for a person, but I couldn't figure out a good way to do this. 

At the moment the top priority is to get this deployed so people can use it, and we can do a later PR to optimise the queries.

### DIFF
--- a/wcivf/apps/api/serializers.py
+++ b/wcivf/apps/api/serializers.py
@@ -30,8 +30,7 @@ class PersonSerializer(serializers.ModelSerializer):
     leaflets = serializers.SerializerMethodField(allow_null=True)
 
     def get_leaflets(self, obj: Person):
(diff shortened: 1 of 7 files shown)


michaeljcollinsuk:
@symroe re optimisation of the leaflet query, this really confused me for a while, but see latest commit which has reduced the query number of queries when getting leaflets to 1 rather than 1 query per-person. Testing on :
Screenshot 2022-03-18 at 15 30 34

It seems that by using the `latest_four` manager method it creates a new query set, so that the benefit from the original prefetch is lost. This SO post was useful https://stackoverflow.com/questions/38487303/django-prefetch-with-custom-queryset-which-uses-managers-method

PR #1110, opened 2022-03-24
Welsh Translations

VirginiaDooley:


wcivf/apps/people/templates/people/includes/_person_about_card.html
```diff
@@ -1,17 +1,18 @@
+{% load i18n %}
 {% if object.wikipedia_bio or object.favourite_biscuit %}
     <div class="ds-card">
         <div class="ds-card-body">
             <h2 class="ds-candidate-name ds-h3">
-                About {{ object.name }}
+                {% blocktrans with name=object.name %}About {{ object.name }}{% endblocktrans %}
```
VirginiaDooley, line 6: I think a `blocktrans` tag is necessary to translate "About", but a name is not translatable is the `name=object.name` necessary?

wcivf/apps/people/templates/people/includes/_person_intro_card.html
```diff
@@ -1,42 +1,48 @@
...
 
-        {% else %}
-            <p>
-                {{ object.intro|safe }}
-            </p>
-        {% endif %}
-    </div>
+{% else %}
+    <p>
+        {% blocktrans %}{{ object.intro|safe }}{% endblocktrans %}
```
VirginiaDooley, line 38: `intro` is created in the model. Should the translation be done there instead?

symroe: Yes, I think so

wcivf/apps/people/templates/people/includes/_person_about_card.html
symroe: 1. We can't use `object.name`, because we can only translate simple variables. So this should be `name`.
2. When making the vars, it's useful to think about what the translators will see in their interface. Here all they will see is that they need to translate `"about [name]"`, but this isn't as useful as e.g `"about [person_name]"`. Or `"about [ward_name]"`. e.g, can you try to describe what the vars are in the name, just in case that's useful?

If 2. isn't easy in the var names, we can use [context markers](https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#contextual-markers) or [comments](https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#comments-for-translators-in-templates)

wcivf/apps/people/models.py
```diff
@@ -8,6 +8,7 @@
 from django.utils.functional import cached_property
 from elections.models import Election, Post
 from parties.models import Party
+from django.utils.translation import gettext_lazy as _
```
VirginiaDooley, line 11: These changes introduced some test failures so I don't think I've done this correctly.

wcivf/apps/people/templates/people/includes/_person_intro_card.html
```diff
@@ -1,32 +1,35 @@
...
+                    {% if candidacy.elected %}
+                        {% trans "(elected)" %}
+                    {% endif %}
+                </li>
                 </ul>
             {% endfor %}
-
         {% else %}
             <p>
                 {{ object.intro|safe }}
```
VirginiaDooley, line 35: I'm not quite sure the best way to handle this in the model.

wcivf/apps/people/templates/people/includes/intros/_constituency.html
```diff
@@ -1,13 +1,13 @@
...
-        {{ candidacy.post.label }}</a> constituency in the <a
-            href="{{ candidacy.election.get_absolute_url }}"
-            title="{{ candidacy.election.name }}">
-            {{ candidacy.election.name_without_brackets }}</a>.
+    {% blocktrans with name=person.name%}{{ name }} {{ verb }} the{% endblocktrans %}
+    <a href="{{ candidacy.party.get_absolute_url }}"title="{{ party }}">{{ party }}</a>
+    {% blocktrans with post_label=candidacy.post.label election=candidacy.election.name with election_name=candidacy.election.name_without_brackets %} candidate for
+        <a href="{{ candidacy.post_election.get_absolute_url }}">{{ post_label }}</a> constituency in the
+        <a href="{{ candidacy.election.get_absolute_url }}" title="{{ election }}">{{ election_name }}</a>
+    {% endblocktrans %}.
```
michaeljcollinsuk, line 9: I think you need to move the fullstop to after the closing `</a>` tag rather than after the `endblocktrans` as by having it on a new line it adds a whitespace (which is why one of tests is failing along with the above)

VirginiaDooley:
👏  One thing I don't see is the how `statement_to_voters` will be translated, if at all?

symroe:
> clap One thing I don't see is the how `statement_to_voters` will be translated, if at all?

Yeah, in general we can't translate content from the database like party names or statements. Hopefully Welsh candidates will just translate their own statements within the text box on YNR

PR #1120, opened 2022-03-31
Feature/previous party affiliations michael

michaeljcollinsuk:


wcivf/apps/people/migrations/0042_personpost_previous_party_affiliations.py
```diff
@@ -0,0 +1,23 @@
...
+        ("people", "0041_dummycandidacy_dummyperson"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="personpost",
+            name="previous_party_affiliations",
+            field=models.ManyToManyField(
+                blank=True,
+                related_name="alliated_memberships",
```
symroe, line 19: Is this a word?

wcivf/apps/elections/views/mixins.py
```diff
@@ -108,7 +108,13 @@ def people_for_ballot(self, postelection, compact=False):
             F("elected").desc(nulls_last=True), *order_by
         )
         people_for_post = people_for_post.select_related(
-            "post", "election", "person", "party"
+            "post",
+            "election",
+            "person",
+            "party",
+        )
+        people_for_post = people_for_post.prefetch_related(
```
symroe, line 116: Maybe I'm confused about how m2m works, but do we need to `select_related` on `Party` with a `Prefetch` here as well? If you add 10 previous parties, does this still create a single query?

wcivf/apps/people/migrations/0042_personpost_previous_party_affiliations.py
michaeljcollinsuk: Haha good catch

wcivf/apps/elections/views/mixins.py
michaeljcollinsuk: It is just the standard `prefetch_related` that is required for a many to many relationship. I've added a test in dae6362f3c55cf49f5bdde9381335376abb9708f to check the number of queries. I've tested the mixin directly because we dont actually display the previous parties in the template in this PR yet.

symroe: Ok, good to know, thanks!

Issue #1077, opened 2022-03-01
Import and present previous party affiliation from YNR


symroe:
Once DemocracyClub/yournextrepresentative#1770 is deployed we need to import the data in the WCIVF.

This should copy the model on YNR:

- [x] Add a column to the `PersonPost` model
- [ ] Present the information on the front end where it exists
- [x] Add the field to the API so that devs.DC can import it

VirginiaDooley:
Closed with https://github.com/DemocracyClub/WhoCanIVoteFor/pull/1126

PR #1126, opened 2022-04-01
Add previous party affiliation to templates

VirginiaDooley:
Ref https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1077 and closes https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1144

Display previous party affiliations relating to all elections on the (revised) elections table
Screen Shot 2022-04-21 at 10 06 33 AM

Display previous party affiliations in the person intro card 1 vs 2+ parties
Screen Shot 2022-04-14 at 7 08 29 PM
Screen Shot 2022-04-14 at 7 20 32 PM

### DIFF
--- a/wcivf/apps/elections/templates/elections/includes/_person_card.html
+++ b/wcivf/apps/elections/templates/elections/includes/_person_card.html
@@ -25,8 +25,7 @@
         {% endif %}
 
         {% if not person_post.list_position or not postelection.display_as_party_list %}
-            <p>
-                {{ person_post.party_name }}
+            <p>{{ person_post.party_name }}
                 {% if person_post.votes_cast %}
                     <br><strong>{% blocktrans with votes_cast=person_post.votes_cast|intcomma%}{{ votes_cast }} votes{% endblocktrans %}</strong>
                 {% endif %}
--- a/wcivf/apps/elections/templates/elections/includes/_previous_party_affiliations.html
+++ b/wcivf/apps/elections/templates/elections/includes/_previous_party_affiliations.html
@@ -1,0 +1,16 @@
+{% load i18n %}
+{% load static %}
+{% load humanize %}
+
+<p>
+    {% for candidacy in candidacies %}
+        {% blocktrans trimmed with person_name=person.name count counter=candidacy.previous_party_affiliations.count %}
+            {{ person_name }} has declared their affiliation with the following party in the past 12 months:
+        {% plural %}
+            {{ person_name}} has declared their affiliation with the following parties in the past 12 months:
+        {% endblocktrans %}
+        {% for previous_party in candidacy.previous_party_affiliations.all %}
+            {{ previous_party.party_name }}{% if forloop and not forloop.last %}, {% endif %}
+        {% endfor %}
+    {% endfor %}
+</p>
--- a/wcivf/apps/people/templates/people/includes/_person_intro_card.html
+++ b/wcivf/apps/people/templates/people/includes/_person_intro_card.html
@@ -19,25 +19,30 @@
 
             {% for candidacy in object.current_or_future_candidacies %}
                 <ul>
-                    <li>{% blocktrans with election=candidacy.election.name %}{{ election }} for{% endblocktrans %}
+                    <li>
+                        {% blocktrans with election=candidacy.election.name %}{{ election }} for{% endblocktrans %}
                         <a href="{{ candidacy.post_election.get_absolute_url }}">
                             {{ candidacy.post_election.friendly_name }}
                         </a>
+                        {% if candidacy.votes_cast %}
+                            {% blocktrans trimmed with num_votes=candidacy.votes_cast|intcomma %}
+                                {{ num_votes }} votes
+                            {% endblocktrans %}
+                        {% endif %}
+                        {% if candidacy.elected %}
+                            {% trans "(elected)" %}
+                        {% endif %}
                     </li>
-                    {% if candidacy.votes_cast %}
-                        {% blocktrans trimmed with num_votes=candidacy.votes_cast|intcomma %}
-                            {{ num_votes }} votes
-                        {% endblocktrans %}
-                    {% endif %}
-                    {% if candidacy.elected %}
-                        {% trans "(elected)" %}
-                    {% endif %}
                 </ul>
+
             {% endfor %}
         {% else %}
             <p>
                 {{ object.intro|safe }}
             </p>
+        {% endif %}
+        {% if object.previous_party_count %}
+            {% include "elections/includes/_previous_party_affiliations.html" with person=object candidacies=object.current_or_future_candidacies %}
         {% endif %}
     </div>
 
--- a/wcivf/apps/people/templates/people/includes/_person_previous_elections_card.html
+++ b/wcivf/apps/people/templates/people/includes/_person_previous_elections_card.html
@@ -1,21 +1,24 @@
 {% load humanize %}
 {% load i18n %}
 
-{% if object.past_not_current_candidacies %}
+{% if object.personpost_set.exists %}
     <section class="ds-card">
         <div class="ds-table">
             <table>
                 <caption style="font-style: normal;">
-                    <h3 class="ds-padded">{% trans "Previous Elections" %}</h3>
+                    <h3 class="ds-padded">{{ object.name }}'s {% trans "Elections" %}</h3>
                 </caption>
                 <tr>
-                    <th>{% trans "Date" %}</th>
-                    <th>{% trans "Election" %}</th>
-                    <th>{% trans "Party" %}</th>
-                    <th>{% trans "Results" %}</th>
+                    <th>Date</th>
+                    <th>Election</th>
+                    <th>Party</th>
+                    <th>Results</th>
+                    {% if object.previous_party_count %}
+                        <th>Other party affiliations in the past 12 months</th>
+                    {% endif %}
                 </tr>
                 <tr>
-                    {% for person_post in object.past_not_current_candidacies %}
+                    {% for person_post in object.personpost_set.all %}
                         {% blocktrans trimmed with message=person_post.post_election.short_cancelled_message_html %}
                             {{ message }}
                             {{ message }}
@@ -29,8 +32,16 @@
                         {% else %}
                             <td>{% if person_post.votes_cast %}{% blocktrans with num_votes=person_post.votes_cast|intcomma %}{{ num_votes }} votes (not elected){% endblocktrans %}{% else %}{% trans "Vote count not available" %}{% endif %}</td>
                         {% endif %}
+                        {% if object.previous_party_count %}
+                            <td>
+                                {% for previous_party_affiliation in person_post.previous_party_affiliations.all %}
+                                    {{ previous_party_affiliation.party_name }}{% if not forloop.last %}, {% endif %}
+                                {% endfor %}
+                            </td>
+                        {% endif %}
                         </tr>
                     {% endfor %}
+
                 </table>
             </div>
         </section>
--- a/wcivf/apps/people/tests/test_person_views.py
+++ b/wcivf/apps/people/tests/test_person_views.py
@@ -139,8 +139,8 @@ class PersonViewTests(TestCase):
             party_name=party.party_name,
         )
         response = self.client.get(self.person_url, follow=True)
-        expected = f"<h5>{self.person.name} is a Liberal Democrat candidate in the following elections:"
-        self.assertContains(response, expected, html=True)
+        expected = "is a Liberal Democrat candidate in the following elections:"
+        self.assertContains(response, expected)
 
     def test_multiple_independent_candidacies_intro(self):
         election_one = ElectionFactory()
@@ -159,8 +159,8 @@ class PersonViewTests(TestCase):
             party_name=party.party_name,
         )
         response = self.client.get(self.person_url, follow=True)
-        expected = f"<h5>{self.person.name} is an Independent candidate in the following elections:"
-        self.assertContains(response, expected, html=True)
+        expected = "is an Independent candidate in the following elections:"
+        self.assertContains(response, expected)
 
     def test_one_candidacy_intro(self):
         election = ElectionFactory()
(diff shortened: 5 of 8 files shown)


wcivf/apps/elections/templates/elections/includes/_person_card.html
```diff
@@ -30,6 +31,10 @@ <h6 class="ds-candidate-name ds-h5">
                 {% endif %}
             </p>
         {% endif %}
+
```
VirginiaDooley, line 34: Not sure why this isn't working

wcivf/apps/elections/templates/elections/includes/_person_card.html
```diff
@@ -30,6 +31,10 @@ <h6 class="ds-candidate-name ds-h5">
...
+        {% if person_post.previous_party_affiliations.all %}
```
VirginiaDooley, line 35: `previous_party_affiliations` is a field on the `PersonPost` model. So why isn't this working?

VirginiaDooley:
Input welcome on the best way to display `previous_party_affiliations`; I realise a list doesn't work well in every context. Another thought would be: "This candidacy was previously affiliated with the following parties: ..., ..., .... "

wcivf/apps/elections/templates/elections/includes/_person_card.html
VirginiaDooley: Yes, I have. Using that data, I have the screen shots above for ballot and person views, but this is for the post code view and it's not rendering the data like the others.

wcivf/apps/elections/templates/elections/includes/_previous_party_affiliations.html
```diff
@@ -0,0 +1,11 @@
+{% load static %}
+{% load i18n %}
+
+<i>
```
symroe, line 4: I don't think the `<i>` tag is right here. It's designed for [_Idiomatic_ text](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i) and I don't fully understand what that means, but I don't think the whole section meets the criteria. I think it's ok to just have this in the normal text style?

wcivf/apps/people/tests/test_person_views.py
```diff
@@ -180,6 +180,29 @@ def test_one_candidacy_intro(self):
...
+            party_name=party.party_name,
+        )
+        previous_party = PartyFactory(
+            party_name="Conservative", party_id="foo_id"
+        )
+        person_post.previous_party_affiliations.add(previous_party)
+        response = self.client.get(self.person_url, follow=True)
+        self.assertTemplateUsed("people/person_detail.html")
+        # assert that previous party affiliations are listed in the template
+        self.assertContains(response, "Previous Party Affiliations")
```
VirginiaDooley, line 201: Does this setup seem right? @michaeljcollinsuk

wcivf/apps/elections/templates/elections/includes/_previous_party_affiliations.html
```diff
@@ -0,0 +1,11 @@
...
+    <p>Previous Party Affiliations</p>
+    {% for previous_party in person_post.previous_party_affiliations.all %}
+        <ul>
```
symroe, line 7: Untesed, does it work to show this inline?

`Previous Party Affiliations: {% for previous_party in person_post.previous_party_affiliations.all %}{{ previous_party.party_name }}{% if not forloop.last %}, {% endif %}{% endfor %}`

symroe (commented):
Couple of inline comments, but the major one is making sure we're not making loads more queries. 

I think it would be good to make sure we're using `select_related` in the views, and maybe only try to show the data if we're on a Welsh run ballot (we can use the same method as we have in YNR for this)

wcivf/apps/elections/templates/elections/includes/_previous_party_affiliations.html
VirginiaDooley: TIL about "[Idiomatic text](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i)"! Will fix that.

VirginiaDooley:
It's worth noting that the candidate image on the post and postcode views narrows with multiple `person_party_affiliations` on the card.

VirginiaDooley:
> Couple of inline comments, but the major one is making sure we're not making loads more queries.
> 
> I think it would be good to make sure we're using `select_related` in the views, and maybe only try to show the data if we're on a Welsh run ballot (we can use the same method as we have in YNR for this)

I am not sure the best way to do this:

> I think it would be good to make sure we're using `select_related` in the views

wcivf/apps/people/models.py
```diff
@@ -51,6 +51,15 @@ def get_local_party(self):
         else:
             return None
 
+    @property
+    def is_welsh_run(self):
+        """
+        Return a boolean if this is a Welsh government run ballot
+        """
+
+        if self.post.nice_territory() == "Wales":
```
VirginiaDooley, line 60: Since (I don't think) we don't have access to `NUTS` keys in WCIVF,  I wasn't sure the best way to find the country. I am confused as to why 'local' seems to fall outside of country definitions rather than within.

wcivf/apps/people/templates/people/includes/_person_intro_card.html
```diff
@@ -23,8 +23,12 @@ <h5>{{ object.name }} is {% if object.current_or_future_candidacies.0.party.is_i
                         {% if candidacy.elected %}
                             (elected)
                         {% endif %}
+                        {% if candidacy.previous_party_affiliations.exists and candidacy.is_welsh_run %}
+                            {% include "elections/includes/_previous_party_affiliations.html" with person_post=candidacy %}
+                        {% endif %}
```
VirginiaDooley, line 28: This isn't working (it was at some point) and I can't figure out why 😕

wcivf/apps/people/models.py
michaeljcollinsuk: I dont think we actually need this, as we can instead just check if the `previous_party_affiliations` exist. The reason for the `is_welsh_run` helper in YNR is about stopped the relations being added on ballots that shouldn't be possible. But in WCIVF there is no update/creating happening, so it is not so important

wcivf/apps/people/templates/people/includes/_person_intro_card.html
michaeljcollinsuk: This does work, but will only display for a candidate that has more than 1 current candidacies (see the conditional on line 11).
If temp hardcode that conditional to `if True` then I see:
Screenshot 2022-04-07 at 16 40 21

michaeljcollinsuk: Having said that, I am unsure if we need to display it in the intro, or if just on the ballot view is enough? Maybe something for @pmk01 to think about?

wcivf/apps/people/models.py
symroe: Sorry, this was my suggestion because I wanted to prevent unneeded queries! As long as we're using `select_related` on them it should be ok, but I'd like to ensure that

wcivf/apps/people/templates/people/includes/_person_intro_card.html
VirginiaDooley: Discussed this PR with @pmk01: Since the language regarding previous parties on the Welsh SOPNs is a bit vague (party affiliations aren't always previous), we discussed the following options: 

1. A explanatory sentence on the person intro card which appears on the person detail view, something like: 

> PERSON NAME has declared their membership with the following political parties in the last 12 months: Party #1, Party #2

2. Add it to the previous elections table to make it less prominent (with no mention in the person list or person detail)

@symroe @michaeljcollinsuk thoughts?

michaeljcollinsuk: Option 2 is probably going to be easier to implement, just because of how we build the intro. Although I dont think it would be too difficult to add to the intro as well, so happy with what people think is better.

wcivf/apps/people/models.py
michaeljcollinsuk: I have added the `prefetch_related` on the method that [gets people for the postcode view](https://github.com/DemocracyClub/WhoCanIVoteFor/pull/1120/files#diff-0e232cd028b7a1c7c67eec353c08f2a2aa146ca5359037faf9fa1fb06da109b6R116) but we will need something similar on the person view also.

wcivf/apps/people/templates/people/includes/_person_intro_card.html
```diff
@@ -19,10 +19,18 @@ <h5>{{ object.name }}
 
             {% for candidacy in object.current_or_future_candidacies %}
                 <ul>
-                    <li>{% blocktrans with election=candidacy.election.name %}{{ election }} for{% endblocktrans %}
```
michaeljcollinsuk, line 22: Is this meant to be deleted?

VirginiaDooley: Thanks for catching.

michaeljcollinsuk:
> It's worth noting that the candidate image on the post and postcode views narrows with multiple `person_party_affiliations` on the card.

So was it decided that we would not show the previous party affiliations on either the post postcode view because of this?

wcivf/apps/people/templates/people/includes/_person_previous_elections_card.html
```diff
@@ -6,13 +6,16 @@
...
                 <tr>
-                    <th>{% trans "Date" %}</th>
-                    <th>{% trans "Election" %}</th>
-                    <th>{% trans "Party" %}</th>
-                    <th>{% trans "Results" %}</th>
+                    <th>Date</th>
+                    <th>Election</th>
+                    <th>Party</th>
+                    <th>Results</th>
+                    {% if object.past_not_current_candidacies.first.previous_party_affiliations %}
```
michaeljcollinsuk, line 16: Missing the `exists` here.

Although looking at this now, the issue with displaying the details in this table is that it will only appear to the user _after_ the election has happened (when the candidacy becomes a past ballot). Which means it loses its usefulness somewhat? But perhaps this was already something discussed with @pmk01?

wcivf/apps/people/models.py
michaeljcollinsuk: @VirginiaDooley in the `get_person` method you may want to try adding:
```
        queryset = queryset.prefetch_related(
            "personpost__previous_party_affiliations",
        )
```
around line 32, before the Try/Except block

wcivf/apps/people/tests/test_person_views.py
```diff
@@ -195,44 +201,57 @@ def test_previous_party_affiliations(self):
...
+            response,
+            "Other party affiliations in the past 12 months",
+            html=True,
         )
         self.assertContains(response, "Conservative, Labour", html=True)
+        self.assertContains(
+            response,
+            "has declared their affiliation with the following parties in the past 12 months",
+            html=True,
+        )
```
VirginiaDooley, line 254: This turns up when I inspect `response.content` 😕

wcivf/apps/people/tests/test_person_views.py
```diff
@@ -179,13 +179,19 @@ def test_one_candidacy_intro(self):
             f"{self.person.name} is a {person_post.party_name} candidate in {person_post.post.label} constituency in the {election.nice_election_name}.",
         )
 
-    def test_previous_party_affiliations(self):
+    def test_previous_party_affiliations_in_current_elections(self):
```
VirginiaDooley, line 182: This test is probably trying to do too much. Note to self to refactor into smaller chunks.

wcivf/apps/people/templates/people/includes/_person_previous_elections_card.html
```diff
@@ -29,8 +32,18 @@ <h3 class="ds-padded">{% trans "Previous Elections" %}</h3>
                         {% else %}
                             <td>{% if person_post.votes_cast %}{% blocktrans with num_votes=person_post.votes_cast|intcomma %}{{ num_votes }} votes (not elected){% endblocktrans %}{% else %}{% trans "Vote count not available" %}{% endif %}</td>
                         {% endif %}
+                        {% if object.previous_party_count %}
+                            <td>
+                                {% for candidacy in object.current_or_future_candidacies %}
```
VirginiaDooley, line 37: Just want to confirm, we are only interested in previous party affiliations for curren elections?

wcivf/apps/people/templates/people/includes/_person_previous_elections_card.html
```diff
@@ -29,8 +32,18 @@ <h3 class="ds-padded">{% trans "Previous Elections" %}</h3>
                         {% else %}
                             <td>{% if person_post.votes_cast %}{% blocktrans with num_votes=person_post.votes_cast|intcomma %}{{ num_votes }} votes (not elected){% endblocktrans %}{% else %}{% trans "Vote count not available" %}{% endif %}</td>
                         {% endif %}
+                        {% if object.previous_party_count %}
+                            <td>
```
michaeljcollinsuk, line 36: You dont want another loop through the candidacies here, as in this block it is already looping through `PersonPost` (candidacies) objects - see line 21. So we want to loop through the `previous_party_affiliations` for the `person_post`

wcivf/apps/people/views.py
```diff
@@ -30,6 +30,11 @@ def get_person(self, queryset=None):
                 # "leaflet_set",
             )
         )
+        queryset = queryset.annotate(
+            previous_party_count=Count(
+                "personpost__previous_party_affiliations"
+            )
+        )
```
michaeljcollinsuk, line 38: Will still need to use `prefetch_related` here re: https://github.com/DemocracyClub/WhoCanIVoteFor/pull/1126#discussion_r848635026

wcivf/apps/people/tests/test_person_views.py
michaeljcollinsuk: When you check the `response.content` does it have any extra whitespaces or line breaks etc?

VirginiaDooley: It has line breaks. In the past, setting `html=True` has handled this. 
`response.content` includes `<p>\n    \n        Candidate 29 has declared their affiliation with the following parties in the past 12 months:\n        \n            Conservative, \n        \n            Labour\n        \n    \n</p>` so I thought this would pass regardless.

wcivf/apps/people/views.py
VirginiaDooley: I forgot to write a comment about this: when I add this to the view, it throws an error: 
`Cannot find 'personpost' on Person object, 'personpost__previous_party_affiliations' is an invalid parameter to prefetch_related()`

wcivf/apps/people/tests/test_person_views.py
michaeljcollinsuk: That's really odd as you are right it looks like it would pass fine... could try reducing text?

wcivf/apps/people/views.py
michaeljcollinsuk: Oh in that case it may need to be `prefetch_related("personpost_set__previous_party_affiliations")`

wcivf/apps/people/tests/test_person_views.py
michaeljcollinsuk: Actually I just tried running this locally without the `html=True` and it passed. But dont ask me why 🤷

VirginiaDooley: Weird. Ok thanks for trying it. Will remove.

wcivf/apps/people/views.py
VirginiaDooley: Still not working 😕

VirginiaDooley: That worked and agree, much cleaner.

michaeljcollinsuk (commented):
This looks good - I think one last thing that is needed is running `django-admin makemessages -l cy --ignore='env*'` as the templates include new translation blocks.

Also this covers the person view, but not the ballot/postcode views. I can't remember if we decided not to show the previous party affiliations on there (I thought we said we would but might be wrong). But regardless perhaps that can be tackled separately if we are adding there

wcivf/apps/people/tests/test_person_views.py
```diff
@@ -179,6 +179,97 @@ def test_one_candidacy_intro(self):
...
+        self.assertTemplateUsed(
+            "elections/includes/_previous_party_affiliations.html"
+        )
+        self.assertContains(
+            response,
+            "Other party affiliations in the past 12 months",
+            html=True,
+        )
+        self.assertContains(response, "Conservative, Labour", html=True)
+
```
michaeljcollinsuk, line 272: One additional test that would be useful would be to check that when the person doesn't have any previous party affiliations, that the related text is **not** shown

VirginiaDooley:
> This looks good - I think one last thing that is needed is running `django-admin makemessages -l cy --ignore='env*'` as the templates include new translation blocks.
> 
> Also this covers the person view, but not the ballot/postcode views. I can't remember if we decided not to show the previous party affiliations on there (I thought we said we would but might be wrong). But regardless perhaps that can be tackled separately if we are adding there

Thanks for the reminder, will do. It was decided to only add the previous party affiliation to the previous election table and person intro given the lack of space on the ballot and postcode views as well as the importance of the information.

symroe:
> This looks good - I think one last thing that is needed is running `django-admin makemessages -l cy --ignore='env*'` as the templates include new translation blocks.

Just a note that we only need to `makemessages` for `en` not `cy`. This is because the translators pull in `en` and write to `cy`. If we only update `cy` then transifex doesn't register any changes

wcivf/apps/people/templates/people/includes/_person_previous_elections_card.html
```diff
@@ -29,8 +32,16 @@ <h3 class="ds-padded">{% trans "Previous Elections" %}</h3>
                         {% else %}
                             <td>{% if person_post.votes_cast %}{% blocktrans with num_votes=person_post.votes_cast|intcomma %}{{ num_votes }} votes (not elected){% endblocktrans %}{% else %}{% trans "Vote count not available" %}{% endif %}</td>
                         {% endif %}
+                        {% if object.previous_party_count %}
```
symroe, line 35: `object.previous_party_count` will resolve to True if the current `PersonPost` has a previous party, but this table will only show previous `PersonPosts`. That means you get the column showing with no data in for a person with a current previous affiliation, but no previous one. This is the only case we'll see!

It will also include the intro text when there is a non-current ballot, because we're using it to include `_previous_party_affiliations.html`.

We have two choices:

1. Change the annotation add a new `has_noncurrent_previous_parties` and `has_current_previous_parties`. Then include the intro text based on `has_current_previous_parties`.
2. Change the "Previous Elections" table to include current elections (changing the text), and add the column based on the current logic. We'll also need to check for current in the intro text.

I think 2 might be better. Looking at it now, it's a little odd that we don't include the current candidacies in that table.

PR #1134, opened 2022-04-05
Volunteer CTA

VirginiaDooley:
Link colour will be changed to white when this change in the https://github.com/DemocracyClub/design-system/pull/44 in design-system v2.0 is deployed.

Screen Shot 2022-04-05 at 6 30 12 PM


pmk01:
Looks good, but am not sure about the '10 minutes' claim... account creation implies longer. What about 'find out how'?

symroe:
I reckon keep the 10 minutes, it's a near CTA even it it might really take longer (or you might get addicted and spend the whole weekend on it)

symroe (approved):

PR #1148, opened 2022-04-08
Bugfix/translations

michaeljcollinsuk:


wcivf/apps/referendums/templates/referendums/includes/_card.html
```diff
@@ -1,4 +1,4 @@
-{% load humanize %}
+{% load humanize i18n %}
```
symroe, line 1: _Super_ minor point, but can you move this to it's own `load` line. This is just because it's helpful to search for `{% load i18n %}` sometimes, and this will get missed

PR #1162, opened 2022-04-11
Whole page translations

VirginiaDooley:
To test, change`LANGUAGE_CODE` to `cy` Screen Shot 2022-04-11 at 7 33 10 PM

### DIFF
--- a/wcivf/apps/elections/tests/test_urls.py
+++ b/wcivf/apps/elections/tests/test_urls.py
@@ -1,0 +1,20 @@
+from django.test import TestCase
+from django.shortcuts import reverse
+from django.conf import settings
+
+settings.USE_I18N = True
+settings.LANGUAGE_CODE = "cy"
+
+
+class TestTranslatedURL(TestCase):
+    def test_translated_ams_url(self):
+        response = self.client.get(reverse("ams_voting_system_view"))
+        self.assertContains(response, "<h2>Y System Aelodau Ychwanegol</h2>")
+
+    def test_fptp_page_url(self):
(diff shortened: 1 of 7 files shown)


wcivf/apps/elections/tests/test_urls.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+import os
+from django.test import TestCase
+from django.shortcuts import reverse
+
+
+@pytest.fixture(autouse=True)
+def env_setup(monkeypatch):
+    monkeypatch.setenv("LANGUAGE_CODE", "cy")
```
VirginiaDooley, line 9: I've tried to set the `LANGUAGE_CODE` for these tests without success.

symroe:
There's something going on with newlines here. First off the HTML isn't indented very nicely, but that's had the effect of converting some paragraphs to headers.

For example on the about page we have an h1 (`#`):

`Ynglŷn â WhoCaniVoteFor Offeryn syml yw WhoCanIVoteFor sy'n caniatáu i bleidleiswyr y DU ddefnyddio’u cod post i weld rhestrau o ymgeiswyr mewn etholiadau sydd ar y gorwel.`

This translates to:

`About WhoCaniVoteFor WhoCanIVoteFor is a simple tool that allows UK voters to use their postcode to view lists of candidates in upcoming elections.`

So we need a newline in there somewhere. 

I could go back to the translators if this is tricky for us to do? But I think there'll only be a few instances.

(I used Google translate on the headers)

symroe (approved):

PR #1164, opened 2022-04-12
Local parties

michaeljcollinsuk:
This appears to have worked in dev as a way to get the local parties stopping from timing out. Wondering if it worth merging for now as is, and then change to using SQS later?


symroe:
With the `delete_existing` change, how do we end up deleting things not in the sheet?

symroe:
Couple more thoughts:

1. I don't think you need to change the time in the jobs. Each one will spin up a new Lambda invocation, so they could all run at the same time in theory. The database might be a limiting factor, but I doubt it (it's really light work for the DB)
2. I wonder if we should add the URL to the model, so we can delete all models with that URL at the start of each run?

michaeljcollinsuk:
@symroe 

1. with the change to delete existing, we wouldn't delete items removed from the sheet. But I hadn't considered this would be an issue? It is using `update_or_create` so changes made to rows would be pulled in.
2. Happy to try with them all at the same time, was a bit of an experiment to stagger them all. I know it seems like the database work shouldn't take long but the loops themselves can be big (in the logs you can see it being killed mid loop).
3. Good idea about adding the url - would solve the issue of not deleting objects mentioned in point 1

PR #1168, opened 2022-04-14
Fix translation tags

VirginiaDooley:
This work fixes an error with translation tag definitions
Before 
Screen Shot 2022-04-14 at 11 01 03 AM

After
Screen Shot 2022-04-14 at 10 57 54 AM

### DIFF
--- a/locale/cy/LC_MESSAGES/django.po
+++ b/locale/cy/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-04-11 11:04+0000\n"
+"POT-Creation-Date: 2022-04-14 15:18+0100\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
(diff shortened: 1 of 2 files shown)


symroe:
Sorry to make this a more complex fix, but, as we're here...


As per https://docs.djangoproject.com/en/3,2/topics/i18n/translation/#pluralization and https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#blocktranslate-template-tag, we should be careful about pluralization.

Can I suggest splitting this block up by adding a `{% plural %}` block in the middle?

Like 

```
{% blocktrans trimmed with winner_count=object.winner_count|apnumber post_label=object.post.full_label num_people=object.people.count|apnumber count object.people.count %}
This election was uncontested because the number of candidates who stood was fewer than the number of available seats.
There is {{ winner_count }} seat in {{ post_label }}, and {{ num_people }} candidate.
{% plural %}
This election was uncontested because the number of candidates who stood was fewer than the number of available seats.
There are {{ winner_count }} seats in {{ post_label }}, and {{ num_people }} candidates.
{% endblocktrans %}
```
Untested, but you should get the idea

michaeljcollinsuk (approved):

symroe (approved):

symroe (approved):

PR #1181, opened 2022-04-20
Dc logging client

michaeljcollinsuk:
Log valid postcodes to the DCWidePostcodeLoggingClient.

To match the work in [WDIV](https://github.com/DemocracyClub/UK-Polling-Stations/pull/4676) it is only logging valid postcodes, and not postcodes via the API.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -21,6 +21,7 @@ newspaper3k==0.2.8
 https://github.com/DemocracyClub/dc_signup_form/archive/refs/tags/2.1.1.tar.gz
 https://github.com/DemocracyClub/design-system/archive/refs/tags/0.2.1.tar.gz
 https://github.com/DemocracyClub/dc_django_utils/archive/refs/tags/0.1.8.tar.gz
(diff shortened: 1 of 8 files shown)


requirements/base.txt
```diff
@@ -21,6 +21,7 @@ newspaper3k==0.2.8
 https://github.com/DemocracyClub/dc_signup_form/archive/refs/tags/2.1.1.tar.gz
 https://github.com/DemocracyClub/design-system/archive/refs/tags/0.2.1.tar.gz
 https://github.com/DemocracyClub/dc_django_utils/archive/refs/tags/0.1.8.tar.gz
+git+https://github.com/DemocracyClub/dc_logging.git@0.0.9
```
symroe, line 24: I think there was a reason we used the `tar.gz` archives above this line...was it to do with when we dump `pipenv` in to a `requirements.txt` file? If this fails, maybe change it to the other format (I'm interested to know!)

symroe (approved):

symroe (approved):

PR #1194, opened 2022-04-25
Reduce SPAM in feedback form

VirginiaDooley:
https://trello.com/c/Vxxj4MrY/2941-feedback-spam-filtering

This PR 
- implements Akismet on the feedback form and filters out any feedback marked as spam from Slack feedback mgmt command. 
- changes the cron job to post to slack hourly

Of 13 total Feedback objects, 1 marked as spam:
```
In [31]: Feedback.objects.all()[9].flagged_as_spam
Out[31]: False

In [32]: Feedback.objects.all()[9].flagged_as_spam = True

In [33]: Feedback.objects.all()[9].save()

In [34]: Feedback.objects.count()
Out[34]: 13
```

Only 12 are posted to slack
Screen Shot 2022-04-26 at 3 39 11 PM


symroe:
API key in email :+1:

wcivf/apps/feedback/views.py
```diff
@@ -11,6 +13,17 @@
...
+    akismet = Akismet(
+        key=settings.AKISMET_API_KEY, blog_url=settings.AKISMET_BLOG_URL
+    )
+
+    @property
+    def is_spam(self):
+        return Akismet.comment_check(
+            user_ip=self.request.META["REMOTE_ADDR"],
+            user_agent=self.request.META["HTTP_USER_AGENT"],
+            comment_type=self.request.POST.get("comments"),
```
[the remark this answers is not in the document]
VirginiaDooley: Thanks. Just realised I had combined two implementations of Akismet.

wcivf/apps/feedback/views.py
```diff
@@ -12,6 +14,18 @@ class FeedbackFormView(UpdateView):
     form_class = FeedbackForm
     template_name = "feedback/feedback_form_view.html"
 
+    @property
+    def is_spam(self):
+        akismet = Akismet(
+            settings.AKISMET_API_KEY, blog=settings.AKISMET_BLOG_URL
+        )
```
symroe, line 21: For safety, can you `return False` if `AKISMET_API_KEY` doesn't exist?

symroe:
I think this is good to go, but before merge we need to think about how to deploy.

This is making a change to the logger DB, so we need to make sure that migrations are run there as well as the main db. @michaeljcollinsuk can you advise?

michaeljcollinsuk:
> I think this is good to go, but before merge we need to think about how to deploy.
> 
> This is making a change to the logger DB, so we need to make sure that migrations are run there as well as the main db. @michaeljcollinsuk can you advise?

This is already handled in [CI](https://github.com/DemocracyClub/WhoCanIVoteFor/blob/master/.circleci/config.yml#L155) via the `make lambda-migrate` [command](https://github.com/DemocracyClub/WhoCanIVoteFor/blob/master/Makefile#L32) which runs migrate for both databases. So should be all good to go - but could test in dev first if preferred.

symroe:
Great, ready to go then! :shipit:

michaeljcollinsuk (approved):
There is conflict in `parameterstore.md` and fixup commit to squash but otherwise 👍

wcivf/apps/feedback/views.py
```diff
@@ -12,6 +14,20 @@ class FeedbackFormView(UpdateView):
...
+
+        if not akismet.api_key:
```
michaeljcollinsuk, line 23: Not a big deal, but if the implementation of `Akismet` changed so that an error is thrown on init with a bad API key (like we had seen previously in CI) this could then cause a 500 error if we didn't have `settings.AKISMET_API_KEY` set. So while not a big deal, it would be safer to check the `settings.AKISMET_API_KEY` directly and return false before trying to initialise `Akismet` e.g:
```
@property
def is_spam(self):
    if not settings.AKISMET_API_KEY:
        return False

    akismet = Akismet(
        settings.AKISMET_API_KEY, blog=settings.AKISMET_BLOG_URL
    )
    return akismet.check(
        self.request.META["REMOTE_ADDR"],
        self.request.META["HTTP_USER_AGENT"],
        comment_content=self.request.POST.get("comments"),
    )



```

michaeljcollinsuk (approved):

PR #1201, opened 2022-04-27
Update ami gevent

michaeljcollinsuk:


wcivf/settings/base.py
```diff
@@ -203,7 +203,7 @@
 
 YNR_BASE = "https://candidates.democracyclub.org.uk"
 YNR_UTM_QUERY_STRING = "utm_source=who&utm_campaign=ynr_cta"
-EE_BASE = "https://elections.democracyclub.org.uk"
+EE_BASE = os.environ.get("EE_BASE", "https://elections.democracyclub.org.uk")
 
 WDIV_BASE = "https://wheredoivote.co.uk"
```
symroe, line 208: One unrelated point to this PR that would be good to get in: can you change this to `http`? We serve the API on HTTPS and HTTP, but HTTP should be faster to return a response, thus helping speed things up.

symroe (approved):

PR #1202, opened 2022-04-27
Display cancelled election title from metadata

michaeljcollinsuk:
Fixes bug displaying cancelled election title from metadata

Screenshot 2022-04-27 at 16 55 23


pmk01:
Brill, thanks!

symroe (approved):

Issue #1223, opened 2022-05-06
ImproperlyConfigured: Could not find the GDAL library when setting up local environment


Samwalton9:
### Issue and Steps to Reproduce
I've been working on getting WCIVF set up locally, following INSTALL.md. That's mostly worked well except when I come to run `python manage.py migrate` for the first time. I get the following error:

> ...
>   File "C:\Users\samwa\WhoCanIVoteFor\venv\lib\site-packages\django\contrib\gis\gdal\libgdal.py", line 46, in <module>
    raise ImproperlyConfigured(
> django.core.exceptions.ImproperlyConfigured: Could not find the GDAL library (tried "gdal302", "gdal301", "gdal300", "gdal204", "gdal203", "gdal202", "gdal201", "gdal20"). Is GDAL installed? If it is, try setting GDAL_LIBRARY_PATH in your settings.

I'm currently following [Django's GDAL install instructions](https://docs.djangoproject.com/en/3.2/ref/contrib/gis/install/geolibs/#gdal) to see if this resolves the issue, but I'd love to know if this is an expected error, if I missed a step, etc.

Samwalton9:
I installed GDAL and set `GDAL_LIBRARY_PATH = r'C:\OSGeo4W\bin\gdal304.dll'`

I'm now getting a new GIS-related error:

>   File "C:\Users\samwa\WhoCanIVoteFor\venv-64\lib\site-packages\django\db\utils.py", line 122, in load_backend
    raise ImproperlyConfigured(
django.core.exceptions.ImproperlyConfigured: 'django.contrib.gis.db.backends.postgis' isn't an available database backend or couldn't be imported. Check the above exception. To use one of the built-in backends, use 'django.db.backends.XXX', where XXX is one of:
    'mysql', 'oracle', 'postgresql', 'sqlite3'

Samwalton9:
Setting `GEOS_LIBRARY_PATH = r'C:\OSGeo4W\bin\geos_c.dll'` solved the new error.

Samwalton9:
I'm now having more GDAL issues while trying to run `pytest` - I had to go and install the GDAL Python package and am getting stuck on:
>  File "c:\users\samwa\whocanivotefor\venv-64\lib\site-packages\django\contrib\gis\gdal\libgdal.py", line 53, in <module>
    lgdal = CDLL(lib_path)
[...]

PR #1225, opened 2022-05-06
Fix use of 's when person's name ends with s

Samwalton9:
I was browsing a local candidate's page when I noticed her name ended with **s**, and messages displayed like `Names's email address`.

I added a boolean property to the `Person` model to understand whether strings about this person should use `'s` or `'`, and implemented this in a few places on the contact card and other pages.

Inspired by the existing party text, in the contact card I also opted to reformat some of the messages so that they didn't require this additional property check, rewording from "X's whatever" to "whatever of X" in the places where that made sense.

It's also worth noting that this fixes a bug where the hovertext for the Wikipedia link didn't include the person's name.

I also added two model tests to verify this behaviour, but I wasn't able to run the tests locally (see #1223). I've verified the behaviour locally on a few candidates pages.

**Updated text when we don't know the person's email address**
image

**Hovertext**
image

image

### DIFF
--- a/wcivf/apps/people/models.py
+++ b/wcivf/apps/people/models.py
@@ -353,6 +353,17 @@ class Person(models.Model):
         intro = intro.strip()
         return " ".join(intro.split())
 
+    @property
+    def name_endswith_s(self):
+        """
+        If the person's name ends with s, return True, else return False,
+        for the purpose of determining possessive noun formatting
+        """
+        if self.name.endswith("s"):
+            return True
+        else:
+            return False
+
 
 class AssociatedCompany(models.Model):
     person = models.ForeignKey(Person, on_delete=models.CASCADE)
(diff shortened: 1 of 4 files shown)


VirginiaDooley:
@Samwalton9 Thanks for this! We've *just* changed our CI settings to run on forked branches. Could I ask you to push an empty commit to force CI to re-run? 

`git commit --allow-empty -m "Empty-Commit"`

sjorford:
Just to throw a spanner in the general direction of this issue, opinion does vary whether Jones' or Jones's is the correct form ([Wikipedia](https://en.wikipedia.org/wiki/Apostrophe#Singular_nouns_ending_with_an_%22s%22_or_%22z%22_sound)) but it varies depending on pronunciation which we can't know

Samwalton9:
Not currently planning to follow up on this and see comment above about whether this is even desirable :)

PR #1236, opened 2022-05-17
Election type filters

VirginiaDooley:
This work adds election type filters in the backend. 

Ref https://github.com/DemocracyClub/WhoCanIVoteFor/issues/773

-  implements the design system filter buttons
-  filters by election type

https://dev.wcivf.club/elections/

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -2,6 +2,7 @@ Django==3.2.13
 
 psycopg2-binary==2.9.3
 requests==2.27.1
+django-filter==21.1
 django-model-utils==4.2.0
 django-markdown-deux==1.0.5
 django-localflavor==3.1
--- a/wcivf/apps/elections/filters.py
+++ b/wcivf/apps/elections/filters.py
@@ -1,0 +1,85 @@
+from urllib.parse import urlencode
+from django.db.models import BLANK_CHOICE_DASH
+from django.utils.encoding import force_str
+from django.utils.safestring import mark_safe
+from django_filters.widgets import LinkWidget
+
+import django_filters
+
+from elections.models import Election
+
+
+def election_types_choices():
+    return [
+        ("europarl", "European Parliament"),
+        ("gla", "London Assembly"),
+        ("local", "Local"),
+        ("mayor", "Mayoral"),
+        ("pcc", "Police and Crime Commissioner"),
+        ("nia", "Northern Irish Assembly"),
+        ("ref", "Referendum"),
+        ("sp", "Scottish Parliament"),
+        ("senedd", "Senedd Cymru"),
+        ("parl", "UK Parliament"),
+    ]
+
+
+class DSLinkWidget(LinkWidget):
+    """
+    The LinkWidget doesn't allow iterating over choices in the template layer
+    to change the HTML wrapping the widget.
+
+    This breaks the way that Django *should* work, so we have to subclass
+    and alter the HTML in Python :/
+
+    https://github.com/carltongibson/django-filter/issues/880
+    """
+
+    def render(self, name, value, attrs=None, choices=(), renderer=None):
+        if not hasattr(self, "data"):
+            self.data = {}
+        if value is None:
+            value = ""
+        self.build_attrs(self.attrs, extra_attrs=attrs)
+        output = []
+        options = self.render_options(choices, [value], name)
+        if options:
+            output.append(options)
+        # output.append('</ul>')
+        return mark_safe("\n".join(output))
+
+    def render_option(self, name, selected_choices, option_value, option_label):
+        option_value = force_str(option_value)
+        if option_label == BLANK_CHOICE_DASH[0][1]:
+            option_label = "All"
+        data = self.data.copy()
+        data[name] = option_value
+        selected = data == self.data or option_value in selected_choices
+        try:
+            url = data.urlencode()
+        except AttributeError:
+            url = urlencode(data)
+        return self.option_string() % {
+            "attrs": selected and ' aria-current="true"' or "",
+            "query_string": url,
+            "label": force_str(option_label),
+        }
+
+
+class ElectionTypeFilter(django_filters.FilterSet):
+    def election_type_filter(self, queryset, name, value):
+        if value == "senedd":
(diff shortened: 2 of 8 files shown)


symroe (commented):
This is looking good to me! 

I think there's some clean up that needs doing (inline comments about NAW, unused imports), but the main feature is working well :+1:

VirginiaDooley:
I've checked this against `main` and it still works well. I don't think the branch off `[feature/filter-elections]` is still necessary.

wcivf/apps/elections/filters.py
```diff
@@ -0,0 +1,94 @@
...
+        ("nia", "Northern Irish Assembly"),
+        ("ref", "Referendum"),
+        ("sp", "Scottish Parliament"),
+        ("senedd", "Senedd Cymru"),
+        ("parl", "UK Parliament"),
+    ]
+
+
+def election_types_choices():
+    qs = (
```
michaeljcollinsuk, line 28: Since the election types are hardcoded, there is no need for this query or even the `_get_election_types_choices_for_qs` method, since the `qs` is not used. So you may as well return the list here directly instead

wcivf/apps/elections/templates/elections/elections_view.html
```diff
@@ -27,4 +35,14 @@ <h2>{% trans "Past Elections" %}</h2>
         {% include "feedback/feedback_form.html" %}
     </div>
 
+    <script type="text/javascript" src="//ajax.googleapis.com/ajax/libs/jquery/2.0.0/jquery.min.js">
+        var filter_box = document.getElementById('election-filters');
+        var filter_button = filter_box.querySelector(e.target);
+        $('#election-filters > ul > li:nth-child(1) > a').click(function (e) {
```
michaeljcollinsuk, line 41: I'm not too sure what this does, or if it is leftover from before and not needed?

wcivf/apps/elections/tests/test_election_views.py
```diff
@@ -97,6 +97,25 @@ def test_division_name_displayed(self):
...
+        """
+        local_election = ElectionWithPostFactory(
+            slug="local.southfields.2022-06-23",
+            election_date="2022-06-23",
+            name="Southfields local election",
+        )
+        for election in [local_election]:
+            with self.subTest(election=election):
+                url = reverse("elections_view")
+                response = self.client.get(url, follow=True)
```
michaeljcollinsuk, line 114: I dont think that the test does what the docstring above describes, as the url used doesn't do any filtering - it is the main election index `/elections/`. So even though the `local_election` is returned as expected, you are not testing If you added in an assert to check that other election is not displayed, I would expect the test to fail.

If you want to use a URL that mimics having clicked a filter button, you will need to add in the relevant query string e.g:
```
url = reverse("elections_view")
url = f"{url}?{election.election_type}"
response = self.client.get(url, follow=True)
```
Which hopefully produces the url `/elections/?election_type=local`. 

Then to make the test more robust, you could create election objects for each election type, add them to your list on line 111, and assert that for each election type, object that election object is returned.

wcivf/apps/elections/views/election_views.py
```diff
@@ -22,14 +21,10 @@ class ElectionsView(TemplateView):
...
         Election = apps.get_model("elections.Election")
-        all_elections = Election.objects.all().order_by(
-            "-election_date", "election_type", "name"
-        )
-
-        context["past_elections"] = all_elections.past()
-        context[
-            "current_or_future_elections"
-        ] = all_elections.current_or_future()
+        qs = Election.objects.all().order_by("-election_date", "name")
```
michaeljcollinsuk, line 24: You might want to keep `election_type` in the `order_by` as before, so that on the unfiltered page, the elections are grouped by election type?

wcivf/settings/base.py
```diff
@@ -45,6 +45,7 @@
     "django.contrib.staticfiles",
     "django.contrib.sitemaps",
     "django.contrib.sites",
+    "django_filters",
```
michaeljcollinsuk, line 48: Is django filters a new dependency? If so remember to add it to `requirements/base.txt`

wcivf/apps/elections/templates/elections/elections_view.html
VirginiaDooley: I was experimenting with preventing page reloads on filter click, but didn't mean to commit. I'll remove. Thanks.

michaeljcollinsuk (approved):
This is looking good, I am not clean why there are feedback from JS changes included, but perhaps its not important

VirginiaDooley:
> This is looking good, I am not clean why there are feedback from JS changes included, but perhaps its not important

There are JS errors that I temporarily removed (shouldn't have committed) and the JS edit here just puts that file back. I'll need to address those errors in a separate PR but they aren't breaking anything.

requirements/base.txt
```diff
@@ -2,6 +2,7 @@ Django==3.2.13
 
 psycopg2-binary==2.9.3
 requests==2.27.1
+django-filter
```
symroe, line 5: Could you pin this to whatever version you have in `pip freeze`?

wcivf/assets/js/scripts.js
```diff
@@ -1,28 +1,28 @@
...
-            method: "POST",
-            body: JSON.stringify(data),
-            found_useful: found_useful,
-            token: feedback_form_token,
-            csrfmiddlewaretoken: csrfmiddlewaretoken,
-            source_url: window.location.pathname
-        }).then(res => {
-            console.log(body);
-        });
+feedback.ready(function() {
```
symroe, line 1: Fallback is actually not a typo here:

https://github.com/dolox/fallback/

We used to use this library, but it looks like it's not installed any more. This change cases an error in the console for me, but so does `fallback`. I think we actually want to unwrap the call here, and just add the javascript directly. Or maybe we need to use `document.ready`?

symroe: I see above you've added the JS inline in the HTML. I don't think that'll be needed if you fix this code

symroe (approved):
Some comments about the JS and pinning the dependency. Feel free to merge once you've sorted that out :+1:

PR #1242, opened 2022-05-23
Default winner_count to 1

michaeljcollinsuk:
- Adds a default of 1 to the PostElection.winner_count field, remove
null.
- Update ballot importer to store winner_count as 1 if a None value
was returned from YNR. This needs to happen at the import stage as
YNR can still store null values for winner_count and return them.
- This relates to this [sentry error](https://sentry.io/organizations/democracy-club-gp/issues/3246374156/?project=1426221&query=is%3Aunresolved) when `winner_count` is not set.
- An alternative approach would be to set the default in Every Election or YNR. That would then ensure that a 'null' value would never be imported in WCIVF. However I was unclear if there were reasons not to do this in EE/YNR.


VirginiaDooley (approved):

symroe:
[500 error for election on 26/5](https://trello.com/c/AliUjcHm/2959-500-error-for-election-on-26-5)

PR #1251, opened 2022-06-07
Translate STV

VirginiaDooley:
This work adds the missing STV Welsh translation and tests a portion of the translated text.


pmk01 (approved):
Diolch!

PR #1254, opened 2022-06-13
How to Vote Translation

VirginiaDooley:
This work manually adds missing translations for how to vote guidance for each voting system. 
I've also: 
- refactored the sentence explaining the number of seats available to allow for simpler plural translations. 
- updated the install instructions with `compilemessages` guidance.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -162,4 +162,8 @@ Once you have pushed your latest changes to your branch:
 `git fetch origin` to get access to the development branch
 `git checkout development` 
 `git merge [YOUR BRANCH NAME]`
-`git push origin develop`
+`git push origin development`
+
+# Update Welsh Translations
+
+After adding new translation tags to a template, run `django-admin compilemessages` then ``django-admin makemessages -l cy --ignore='env*`` to generate matching translation strings to be translated.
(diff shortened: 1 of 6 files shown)


VirginiaDooley:
Changing the 'How you Vote" header in the translation file does not change to Welsh as expected. 
Screen Shot 2022-06-13 at 10 44 22 AM

VirginiaDooley:
How shall I add Welsh voting system names? 

AMS System Aelodau Ychwanegol
FPTP: Y Cyntaf i’r Felin
sv: Pleidlais Atodol
STV: Pleidlais Sengl Drosglwyddadwy
PR-CL: Rhestr Gaeedig
PR-CL: Rhestr Gaeedig

symroe:
> Ideas on how I should add Welsh voting system names?

This is a bit tricky! We could add a column for the Welsh name, but that wouldn't help if we wanted to add other languages in future.

Seeing as we're hard coding the systems in the model already (and we don't really need a database for this model), I think the best option here is to add a new method `get_name` to the model and use the same conditional as we use in `get_absolute_url` to return the translated name, with the text hard coded:

```
@property
    def get_name(self):
        if self.slug == "FPTP":
            return _("First-past-the-post")
```

This will then cause the voting system names to show up in the translations files, where you can add the Welsh. 

You then need to update the code to use this method of course

symroe (approved):

PR #1305, opened 2022-10-11
Hotfix/base template update

VirginiaDooley:
This work 

- Updates the base template structure following additional functionality added in`dc_django_utils=2.0.1`
- Additional package updates were required as a result of these changes
- Replaces `django-robots` https://django-robots.readthedocs.io/en/latest/history/#unreleased
- Addresses deprecated imports following Django upgrade


symroe (approved):

symroe:
This looks good to me!

One minor point I thought of after our call: Can you add a comment / change the name of the test that we got stuck on to indicate that it's checking for text in OG tags?

I can see this tripping future us up, so let's be kind to them!

PR #1375, opened 2022-11-26
Uncontested fixes

### LINKED ISSUE #1332: Candidate profile text for future elections
If a candidate has an election in the future, the table text on their profile reads **Not Elected (vote count not available)**

Can we change this? Perhaps to **Election to be held on [date]**, or similar

Screenshot 2022-11-02 102001

symroe:
* Refactor code to be in model and not in template
* Fix test that marked someone as not elected, but expected the text to say elected
* Show message about future elections in table (fixes #1332)
* Show people when ballot if uncontested

## Future un-cancelled elections 
image

### DIFF
--- a/wcivf/apps/elections/templates/elections/includes/_single_ballot.html
+++ b/wcivf/apps/elections/templates/elections/includes/_single_ballot.html
@@ -120,31 +120,31 @@
                         {% endif %}
                     {% endif %}
                 </p>
+            {% endif %}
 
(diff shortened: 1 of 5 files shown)


symroe:
@Peter @VirginiaDooley Let's go through this together to make sure we're ok with it

VirginiaDooley (approved):
Much clearer. ⭐

pmk01 (approved):

PR #1484, opened 2023-01-26
Remove no-ID messaging for EN/CY

Bekabyx:
_I know the if/else now functionally does the same thing in either case, but I decided to leave the structure in place for when we revisit ID messaging - happy to remove it on request though!_


Bekabyx:
@symroe There is also separate messaging for NI in this project - would you like me to whip that out too (as suggested on WDIV)?

symroe:
Yeah, good spot. Let's make the messaging the same and then we can add the "needs ID" flag to elections and bring it back in again as needed. :+1:

VirginiaDooley (approved):
Approving based on the the assumption that the translation still works based on these changes.

symroe (approved):

PR #1516, opened 2023-02-16
Fix slack notifications

VirginiaDooley:
I missed a step in my previous, related PR of adding the bot to the default slack channel which was causing the channel not to be found. This PR also corrects the indentation which was causing arguments to be missed.  I tested a passing config in the `sam_build` step to confirm the API is working as expected. 

Screenshot 2023-02-16 at 3 58 30 PM


symroe (approved):
Grr, YAML :sob:

Bekabyx (approved):

PR #1520, opened 2023-02-20
Add token variable to deploy config

VirginiaDooley:
After another attempt of adding context variables to the build process at an earlier stage in the CI process, they were still not available at deploy and therefore [the slack notification step has failed in CI](https://app.circleci.com/pipelines/github/DemocracyClub/WhoCanIVoteFor/2821/workflows/5f37ce03-805d-4f2b-b88c-8e1759c401c9/jobs/7180).  After reviewing the env variables in the CI sam deploy development step, I thought I would try to add the slack access tokens to the sam template alongside other redacted variables as a trial step to address the tokens not being found in the staging and prod deploy.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -185,6 +185,8 @@ jobs:
                    AppSentryDsn='$SENTRY_DSN' \
                    AppSecretKey='$SECRET_KEY' \
                    AppSlackFeedbackWebhookUrl='$SLACK_FEEDBACK_WEBHOOK_URL' \
+                   AppSlackAccessToken = '$SLACK_ACCESS_TOKEN' \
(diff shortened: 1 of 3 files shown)


VirginiaDooley:
[make deploy fails more visible](https://trello.com/c/7Qmm5gtY/3268-make-deploy-fails-more-visible)

VirginiaDooley:
From Sym:

> I took a look at the shared RDS and saw that there were loads of lost replication slots in there (about 50)
> 
> I did this using select * from pg_replication_slots; as per https://github.com/DemocracyClub/WhoCanIVoteFor/blob/master/docs/rds.md
> 
> I think this was caused by problems with the instances from before (when backups EE were broken)

Closing this PR as it was an attempt at a deploy fix but in the wrong direction.

PR #1534, opened 2023-03-24
Remove staging from the ci config

Bekabyx:
This is a temporary measure to get us through the weekend due to some tricky problems popping up on deploy.


symroe (approved):
CircleCI is the true reviewer in this case, but I think it looks good!

PR #1558, opened 2023-04-11
Use Devs.DC API

symroe:
This PR is the last part of https://github.com/DemocracyClub/aggregator-api/discussions/375

### DIFF
--- a/wcivf/apps/elections/devs_dc_client.py
+++ b/wcivf/apps/elections/devs_dc_client.py
@@ -1,0 +1,43 @@
+from json import JSONDecodeError
+from urllib.parse import urljoin
+
+import requests
+
+from django.conf import settings
+
+
+class DevsDCAPIException(Exception):
+    def __init__(self, response: requests.Response):
+        try:
+            self.message = {"error": response.json().get("detail", "")}
+        except JSONDecodeError:
+            self.message = ""
+        self.status = response.status_code
+        self.response = response
+
+
+class DevsDCClient:
+    def __init__(self, api_base=None, api_key=None):
+        if not api_base:
+            api_base = settings.DEVS_DC_BASE
+        self.API_BASE = api_base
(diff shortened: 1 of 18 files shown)


symroe:
All but one test passing now :tada: 

The remaining test isn't hard to fix (it involves updating the iCal fixture / view), but the thing that's hard about the iCal view is:

We offer people to subscribe to an iCal feed for upcoming elections. At the moment this is just done on postcode and this causes some problems with false positives / negatives.

Address pickers avoid this, but the problem is that the address picker will depend on the polling station data we have imported, so might not be predictable. That is, if someone subscribes to a postcode this year, they might need to pick from an address picker in future years. 

Ideally we'd just have an ical feed for each UPRN and never for a postcode, but that would require the user to pick a UPRN before they could subscribe. This isn't easy for us to do at the moment.

I'll have a think about how best to deal with this case. It might be that we just ask for a UPRN and find a way to pass this through the API to this project.

Bekabyx:
Stylistic point, I feel like the wording here should probably be more like "select your address" or "choose your address" because the user isn't "entering" anything. 

Screenshot 2023-05-03 at 15 15 12

wcivf/apps/elections/devs_dc_client.py
```diff
@@ -0,0 +1,43 @@
...
+        req = requests.get(url, params={"auth_token": self.API_KEY})
+        if req.status_code >= 400:
+            raise DevsDCAPIException(response=req)
+        return req.json()
+
+    def get_postcode(self, postcode):
+        # TODO: Logging
+        return self.make_request(postcode=postcode)
+
+    def get_uprn(self, postcode, uprn):
```
Bekabyx, line 42: Is this function (and `get_postcode`) in use anywhere? Unless you have plans for these we can probably take them out.

PR #1560, opened 2023-04-17
Update Party model

Bekabyx:
The first drop of this adds additional fields to the party model to make use of the extra data that YNR gives us. This is the preliminary phase of the party pages overhaul. [Trello](https://trello.com/c/dVLjcV1A/3225-wcivf-political-party-pages).

Largely, I've just taken the existing model code form YNR and bolted it on to the WCIVF model, being careful not to change any existing field names in WCIVF. Some of these fields have had to be made nullable or be given defaults which aren't present in YNR simply because the migration wasn't happy. These felt pretty safe changes to make because YNR is our source of truth and it has all the appropriate checks and balances in place, plus these fields aren't in use anywhere at the moment.

These are:
- `ec_id` is nullable. This will only be used to generate links to the EC party registration so not having a value isn't the end of the world.
- `date_registered` is nullable. I chose to make this nullable rather than assign a default because it feels less risky than setting an arbitrary default date. 
- `status` defaults to `Registered`. I made this choice rather than setting nullable because the in the event of a dereg party being wrongly attributed as registered, the behaviour would match our existing implementation anyway.

These will be (over)written with the YNR values when the `import_parties` command is run so there is minimal risk in my opinion.

_Emblems and descriptions to be updated in future PR as they require additional modelling_

To test locally:
- run `python manage.py import_parties`
- gaze in awe as your parties are updated to have additional information available on their database entries. I tested this with Conservative and Unionist Party `select * from parties_party where party_id = 'party:52'` but feel free to try any other party.

```[tasklist]
### PR Checklist

- [x] Note what card in Trello this work relates to
- [x] Did I explain all possible solutions and why I chose the one I did?
```


symroe (approved):

VirginiaDooley (approved):
👍 
`{'_state': <django.db.models.base.ModelState at 0x106d3c7c0>,
 'party_id': 'party:52',
 'party_name': 'Conservative and Unionist Party',
 'emblem_url': 'https://s3.eu-west-2.amazonaws.com/static-candidates.democracyclub.org.uk/media/emblems/PP52/47_Emblem_47.gif',
 'wikipedia_url': '',
 'description': '',
 'ec_id': 'PP52',
 'register': 'GB',
 'status': 'Registered',
 'date_registered': datetime.date(1999, 1, 14),
 'date_deregistered': None}`

PR #1562, opened 2023-04-17
Feat/party emblems and descriptions

Bekabyx:
Phase 2 of the party pages overhaul. [Trello](https://trello.com/c/dVLjcV1A/3225-wcivf-political-party-pages).

I've ported over the `Emblem` and `PartyDescription` models from YNR and added these to the importer. Settled on adding them as separate methods called in the existing `import_parties` method so that it's easily refactorable later if we'd like to move it elsewhere. Again, the models have a couple of changes form their YNR counterparts, just small naming tweaks though so the new models don't tread on the toes of the existing ones.

Because the descriptions don't have a PK in the YNR payload I've put a unique constraint on the description and party together, and in light of this I set up some Integrity Error logging in case there's anything in the YNR data that doesn't meet this new criterion.

To test locally:
- run the migrations
- run `python manage.py import_parties`
- in your db you should have new tables for emblems and descriptions - run something like `select * from parties_partydescription/parties_partyemblem where party_id='party:52'` (using the Conservative party again because I know they have multiple emblems and multiple descriptions). Interestingly they don't have a default emblem but that's by-the-by. 

```[tasklist]
### PR Checklist

- [x] Note what card in Trello this work relates to
- [x] Have I rebased with the latest version of master?
```


wcivf/apps/parties/management/commands/import_parties.py
```diff
@@ -16,8 +17,31 @@ def handle(self, **options):
...
+                    party=party_obj,
+                    description=description["description"],
+                    date_description_approved=description[
+                        "date_description_approved"
+                    ],
+                    defaults={
+                        "date_description_approved": description[
+                            "date_description_approved"
+                        ]
+                    },
```
symroe, line 33: This is a mix of a few things we talked about. I think we should use `update_or_create` here, passing in `party=` and `description=` as kwargs but _not_ `date_description_approved`. That can stay in the `defaults` dict as it is at the moment.

This, translated means: 

1. Find a model with the same party and description as given (by the kwargs)
2. If it exists, update the model to set the  `date_description_approved`
3. If not, make the model with all the kwargs and defaults.

wcivf/apps/parties/management/commands/import_parties.py
```diff
@@ -16,8 +17,31 @@ def handle(self, **options):
...
+                )
+
+            except IntegrityError as ex:
```
symroe, line 36: With the above suggestion, this should never happen

symroe (approved):

VirginiaDooley (approved):

Issue #1232, opened 2022-05-11
Don't display 'deceased' on WCIVF


pmk01:
If a candidate dies we cancel the election & don't show anything, so we don't need to display text about this on WCIVF.

PR #1593, opened 2023-05-10
Remove deceased property and text

VirginiaDooley:
Closes https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1232

This decrease in coveralls coverage is a bug; the coverage remains the same as master.

### DIFF
--- a/locale/cy/LC_MESSAGES/django.po
+++ b/locale/cy/LC_MESSAGES/django.po
@@ -331,12 +331,6 @@ msgstr "Symbol y blaid"
 msgid "Show %(num_candidates)s candidates"
 msgstr "Dangos %(num_candidates)s ymgeiswyr"
 
-#: wcivf/apps/elections/templates/elections/includes/_person_card.html:16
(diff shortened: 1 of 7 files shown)


pmk01 (approved):

Bekabyx:
There's another instance of "(Deceased)" in `elections/templates/elections/includes/_person_card.html L14-18`  - should this one be removed too? This one doesn't use the `display_deceased` property but I thought I'd flag it in case it is also unwanted.

VirginiaDooley:
> There's another instance of "(Deceased)" in `elections/templates/elections/includes/_person_card.html L14-18` - should this one be removed too? This one doesn't use the `display_deceased` property but I thought I'd flag it in case it is also unwanted.

Nice spot, I think you're right. I've remove that and translations as well and added to 3af2194e8505064ee7aacaf25527f6da48a051fd

Bekabyx (approved):

VirginiaDooley:
[Remove deceased text](https://trello.com/c/EGzrI8G3/3363-remove-deceased-text)

Issue #1599, opened 2023-05-17
Remove Crowdsourcing CTA on organisation view post-sopn day


pmk01:
![Screenshot 2023-05-17 135737](https://github.com/DemocracyClub/WhoCanIVoteFor/assets/61510609/9439618a-c8f7-43b2-a77b-9ea1cce1136c)

People can't add any more candidates to these ballots, so this link should be removed after SoPN day.

https://whocanivotefor.co.uk/elections/local.causeway-coast-and-glens.2023-05-18/causeway-coast-and-glens-local-election/

PR #1600, opened 2023-05-17
Show crowdsourcing link only before sopn day

VirginiaDooley:
closes https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1599 and https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1627

Past expected SOPN date - don't show crowdsourcing link
Screenshot 2023-07-06 at 12 28 17 PM

Not yet past expected SOPN date - show crowdsourcing link
Screenshot 2023-07-06 at 12 28 08 PM

### DIFF
--- a/wcivf/apps/elections/templates/elections/election_view.html
+++ b/wcivf/apps/elections/templates/elections/election_view.html
@@ -27,7 +27,6 @@
                         {% blocktrans trimmed with was_will_be_in_past=object.in_past|yesno:"was,will be" election_date=object.election_date|naturalday:"\o\n l j F Y" %} This election {{ was_will_be_in_past }} held <strong>{{ election_date }}</strong>.{% endblocktrans %}
                     {% endif %}
                 </p>
-
                 {% if object.election_type != "ref" %}
                     {% if election.person_set.count %}
                         <p>
@@ -46,7 +45,7 @@
                         {% endif %}
 
                     {% else %}
-                        {% if not object.in_past and not election.slug == 'parl.2017-06-08' %}
+                        {% if not object.postelection_set.first.past_expected_sopn_day %}
                             <p><a href="{{ object.ynr_link }}">{% trans "Add some candidates at our candidate crowd-sourcing site" %}</a></p>
                         {% endif %}
                     {% endif %}
--- a/wcivf/apps/elections/models.py
+++ b/wcivf/apps/elections/models.py
@@ -404,6 +404,7 @@ class PostElection(TimeStampedModel):
             return get_election_timetable(
                 self.ballot_paper_id, self.post.territory
             ).sopn_publish_date
+
         except AttributeError:
             return None
 
@@ -604,7 +605,7 @@ class PostElection(TimeStampedModel):
     @property
     def past_expected_sopn_day(self):
         """
-        Return boolean for if the date we expected the sopn date
+        Return boolean for the date we expected the sopn date
         """
         return self.expected_sopn_date <= timezone.now().date()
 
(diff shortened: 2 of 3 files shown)


VirginiaDooley:
[Show crowdsourcing link only before sopn day](https://trello.com/c/xXUdTDBL/3376-show-crowdsourcing-link-only-before-sopn-day)

Bekabyx:
I've checked this out locally and noticed that there is still a reference to the old function here https://github.com/DemocracyClub/WhoCanIVoteFor/blob/045ef1cf9c445d9651d90c206cbb814b916cb84b/wcivf/apps/elections/templates/elections/includes/_single_ballot.html#L140 which should probably be updated to use the new `past_expected_sopn_date` function.
Also the test is still in place for the previous version of the function (`past_expected_sopn_day`) so this should probably be re-written or removed.

wcivf/apps/elections/templates/elections/election_view.html
```diff
@@ -41,12 +40,12 @@ <h3>{{ object.nice_election_name }}</h3>
                             {% endblocktrans %}
                         </p>
 
-                        {% if not object.in_past and not object.locked %}
+                        {% if not object.is_past and not object.locked %}
```
Bekabyx, line 43: I can't find the `is_past` property in the codebase, and if it's missing then this will always evaluate as True. I can only find `in_past`.

Bekabyx:
RE: the failure of `test_crowdsourcing_not_visible` - my local debugging shows that `{{ object.past_expected_sopn_date }}` is either not present or doesn't have a value. I think the reason for this is that you're accessing it from `election_view.html` as `object.past_expected_sopn_date` but at this point `object` seems to be an `Election` not a `PostElection`. The `past_expected_sopn_date` property is specific to `PostElection`.

VirginiaDooley:
> RE: the failure of `test_crowdsourcing_not_visible` - my local debugging shows that `{{ object.past_expected_sopn_date }}` is either not present or doesn't have a value. I think the reason for this is that you're accessing it from `election_view.html` as `object.past_expected_sopn_date` but at this point `object` seems to be an `Election` not a `PostElection`. The `past_expected_sopn_date` property is specific to `PostElection`.

Good call. I've changed this to iterate through postelection objects on the election before hitting the conditional.

wcivf/apps/elections/templates/elections/election_view.html
```diff
@@ -46,9 +45,11 @@ <h3>{{ object.nice_election_name }}</h3>
                         {% endif %}
 
                     {% else %}
-                        {% if not object.in_past and not election.slug == 'parl.2017-06-08' %}
-                            <p><a href="{{ object.ynr_link }}">{% trans "Add some candidates at our candidate crowd-sourcing site" %}</a></p>
-                        {% endif %}
+                        {% for postelection in object.postelection_set.all %}
```
symroe, line 48: This is going to show the line below for _each_ ballot in the election, isn't it? if `object` is an `Election`, it could have ~30 related `PostElection` objects.

Better to take the first one and find out if the SOPN date has past. It's not true that all ballots in an election will have the same timetable, but it's close enough for this PR I think.

wcivf/apps/elections/templates/elections/election_view.html
```diff
@@ -46,9 +45,11 @@ <h3>{{ object.nice_election_name }}</h3>
...
+                            {% if not postelection.past_expected_sopn_day and not election.slug == 'parl.2017-06-08' %}
+                                <p><a href="{{ object.ynr_link }}">{% trans "Add some candidates at our candidate crowd-sourcing site" %}</a></p>
```
symroe, line 50: We can drop the parl 2017 logic I think, and just stick with the first conditional.

symroe (approved):

PR #1618, opened 2023-06-07
Hotfix/make token more unique

VirginiaDooley:
co-authored @symroe 

Ref https://democracy-club-gp.sentry.io/issues/4152565476/?project=1426221&query=feedback&referrer=issue-stream&statsPeriod=14d&stream_index=0

Previous to this change, tokens were being set to `None` as a default which caused a `Multiple objects returned` error. Here, we've set the token in the model and tested for uniqueness. 

In the course of this work, I ran into a deprecation issue with `setuptools` which is a dependency. 

---
  - https://app.asana.com/0/0/1205086888432584


symroe (commented):
Could you add a test for this please? Just need to call `get_object` and make sure it actually works!

requirements/base.txt
```diff
@@ -16,6 +16,7 @@ redis==4.6.0
 Pillow==9.5.0
 icalendar==5.0.7
 dealer==2.1.0
+setuptools==58.2.0
```
symroe, line 19: It looks like this is an error with `pyYAML`. We're using `5.4.1` and the latest is `6.0.1`. I took a look [at the changelog](https://github.com/yaml/pyyaml/blob/master/CHANGES) and the version after the one we're on [seems to fix this](https://github.com/yaml/pyyaml/pull/564). 

So I think the solution here is to upgrade `pyYAML` rather than pin setuptools to an older version.

PR #1678, opened 2023-08-16
Filter out legacy parties

Bekabyx:
[Asana](https://app.asana.com/0/1204880927741389/1205258090051719)

This work addresses a bug which Peter found where we're serving some legacy parties on the party lookup page. There's a little more detail on the Asana ticket, but essentially any party starting with "ynmp" is legacy.

There are two notable exceptions to this, and those are Speaker Seeking Re-election and Independent. This means the logic isn't quite as straightforward so I've had to import `Q` to get the filter just right 👌 

To test locally:
- Open up the parties page locally
- Make sure Independent and Speaker seeking re-election are in the list
- "National Liberal Party, The Third Way" should **NOT** be in the list _(gotcha: there is a party called National Liberal Party - True Liberalism. This one is supposed to be in the list.)_

```[tasklist]
### PR Checklist
- [x] Note what card in Asana this work relates to
- [x] Did I explain all possible solutions and why I chose the one I did?
- [x] Added any comments to make new functions clearer
```

### DIFF
--- a/wcivf/apps/parties/views.py
+++ b/wcivf/apps/parties/views.py
@@ -1,3 +1,4 @@
+from django.db.models import Q
 from django.views.generic import DetailView, TemplateView
 
 from .filters import PartyRegisterFilter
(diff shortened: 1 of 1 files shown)


VirginiaDooley (approved):
Tested locally 🚀

Issue #134, opened 2017-05-20
More or better links to election pages


JoeMitchell:
Jag expected there to be a link here:

![image](https://cloud.githubusercontent.com/assets/8096305/26277372/cd3fc80e-3d7d-11e7-88de-abd2d0068164.png)

Because he didn't want to share his postcode.

Presumably if there was a link, it'd go to the constituency page via /election, but there's some funny stuff going on there...

![image](https://cloud.githubusercontent.com/assets/8096305/26277380/e4a780ea-3d7d-11e7-827e-1b123db102b2.png)

pmk01:
This is still valid. 
Also:
![Screenshot 2020-12-22 110013](https://user-images.githubusercontent.com/61510609/102881563-f42d3380-4444-11eb-97c1-b4db0921ba15.png)

VirginiaDooley:
> This is still valid. Also: ![Screenshot 2020-12-22 110013](https://user-images.githubusercontent.com/61510609/102881563-f42d3380-4444-11eb-97c1-b4db0921ba15.png)

I think this is ok? https://whocanivotefor.co.uk/elections/TE1%201ST/

pmk01:
Thanks! It is better than ok. 

It's not an answer to the original issue, though, which is devising an easy way to share the ballot without sharing your postcode. Reading back over the tweets, I like the idea of a 'share this ballot' button. However, I think that's a bigger piece of work. Perhaps the answer is to write that up and close this issue?

pmk01:
Yes, you got it. As things stand, there's no obvious way to find a link to the ballot. The original issue suggested simply linking the ballot title text, but on reflection I'm not sure that's the best way of telling people they can share the ballot. A button would be clearer, I think. This is probably something we need to talk over.

pmk01:
Oh right, sorry. I want people to be able to share the ballot. I hadn't thought about the multi-ballot option. That sounds complicated...

GeoWill:
It is. Sharing the ballot is fine, and what Virginia is doing in #1682 should address that. Admittedly it will require c&p a link for now, but adding a button can happen later/needs design thought as you say. The multi ballot option is best served by a postcode imo.

PR #1682, opened 2023-08-22
Test ballot anchors when more than one exists

VirginiaDooley:
This PR addresses part 2 of  https://github.com/DemocracyClub/WhoCanIVoteFor/issues/134 by testing that both ballot anchor links work as expected when the user is presented with more than one ballot. 

- [x] test anchor links are working on post code page as expected when there is more than one ballot available, with hard code links
- [x] test corresponding election name exists

---
  - https://app.asana.com/0/0/1205303544790839

### DIFF
--- a/wcivf/apps/elections/tests/test_postcode_views.py
+++ b/wcivf/apps/elections/tests/test_postcode_views.py
@@ -226,15 +226,22 @@ class TestPostcodeViewPolls:
         )
 
     def test_multiple_elections_london(self, mock_response, client):
+        """This test is for the case where there are multiple elections in
+        London on the same day and we want to make sure that the correct
+        polling station opening times are displayed for each election
+        as well as the correct anchor links to the election pages
+        """
         local_london = PostElectionFactory(
             ballot_paper_id="local.city-of-london.aldgate.2021-05-06",
-            election__slug="local.city-of-london.2021-05-06",
-            election__election_date="2021-05-06",
+            election__slug="local.city-of-london.2024-05-06",
+            election__election_date="2024-05-06",
+            election__name="City of London Corporation local election",
         )
         parl_london = PostElectionFactory(
             ballot_paper_id="parl.cities-of-london-and-westminster.by.2021-05-06",
-            election__slug="parl.2021-05-06",
-            election__election_date="2021-05-06",
+            election__slug="parl.2024-05-06",
+            election__election_date="2024-05-06",
+            election__name="Cities of London and Westminster by-election",
         )
 
         mock_response.json.return_value["dates"].extend(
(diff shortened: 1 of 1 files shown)


GeoWill:
Bit of a drive by, so apologies if I've got the wrong end of the stick. You've modified `inline_elections_nav_list.html` and `_single_ballot.html`. These are both included in `postcode_view.html`. This is controlled by `PostcodeView` which is tested by `test_postcode_views.py` - which is where I think your tests want to be rather than `test_election_views.py`. 

I think the issue happens when there are multiple ballots for a single postcode. This means there is more than one type of election happening on a given day - eg a `local` and a `pcc`. Your test is setting up a scenario where there two elections of the same type - but in different wards (`Posts` in wcifv) - this won't appear for the user as multiple ballot papers, as they will only vote in one ward. 

There are some examples of how to set up the mocks for having multiple different election types in `test_postcode_views.py`: https://github.com/DemocracyClub/WhoCanIVoteFor/blob/0918d80c363aedde14479b7c6f65134aaa38aee0/wcivf/apps/elections/tests/test_postcode_views.py#L269-L280

I guess you could just add some tests that the expected urls are present in the response to those test methods.

VirginiaDooley:
> Bit of a drive by, so apologies if I've got the wrong end of the stick, but hopefully useful in how to fix the testing approach.
> 
> You've modified `inline_elections_nav_list.html` and `_single_ballot.html`. These are both included in `postcode_view.html`. This is controlled by `PostcodeView` which is tested by `test_postcode_views.py` - which is where I think your tests want to be rather than `test_election_views.py`.
> 
> I think the issue happens when there are multiple ballots for a single postcode. This means there is more than one type of election happening on a given day - eg a `local` and a `pcc`. As you say, your test is setting up a scenario where there two elections of the same type - but in different wards (`Posts` in wcifv) - this won't appear for the user as multiple ballot papers, as they will only vote in one ward.
> 
> There are some examples of how to set up the mocks for having multiple different election types in `test_postcode_views.py`:
> 
> https://github.com/DemocracyClub/WhoCanIVoteFor/blob/0918d80c363aedde14479b7c6f65134aaa38aee0/wcivf/apps/elections/tests/test_postcode_views.py#L269-L280
> 
> I guess you could just add some tests that the expected urls are present in the response to those test methods.

Thank you! I don't think this is actually an issue after a closer look, but this comment is helpful. I am going to add some tests regardless. Never hurts.

wcivf/apps/elections/tests/test_postcode_views.py
```diff
@@ -304,6 +304,54 @@ def test_multiple_elections_not_london(self, mock_response, client):
...
+                },
+            ]
+        )
+
+        response = client.get(
+            reverse("postcode_view", kwargs={"postcode": "TE11ST"}), follow=True
+        )
+        asserts.assertContains(
+            response,
+            f'<a href="{post_election1.get_absolute_url()}">copeland</a>',
```
GeoWill, line 348: I think it would be better to hard code these. At the moment if [get_absolute_url()](https://github.com/DemocracyClub/WhoCanIVoteFor/blob/0918d80c363aedde14479b7c6f65134aaa38aee0/wcivf/apps/elections/models.py#L177) changes to return `None` (or whatever) then the test will pass but the link will be broken.

GeoWill:
At the moment the links referenced in the issue are from here https://github.com/DemocracyClub/WhoCanIVoteFor/blob/228fae0464e49c4940854c05762491cbc1a06921/wcivf/apps/elections/templates/elections/includes/inline_elections_nav_list.html#L53. So this PR isn't currently testing those. That doesn't mean the tests written aren't useful, but it's a bit confusing relating them to that issue. 

I also think the test should hard code the string you expect to see, e.g. `<a href="#election_local.faketown.2022-05-05">`,  rather than use the method that the view/template uses (`.get_absolute_url()`). This is because it makes the intention of the test clearer, and guards against the method being broken.

GeoWill (approved):
LGTM - it's testing current behaviour as is - that's a good thing!

Issue #1625, opened 2023-06-19
wikipedia isn't updating


pmk01:
![Screenshot 2023-06-19 111007](https://github.com/DemocracyClub/WhoCanIVoteFor/assets/61510609/2b637013-cf37-4211-8226-383f2b0ce263)

VirginiaDooley:
Update: After not running for several months due to a failed rollback, I've confirmed this job has run on 5/9/23. However, the job is designed to only run imports for current candidates, which would not include the individual above. I think if we remove the `--current` flag from the job then all candidates - current and past - will receive up to date wikipedia bios. Before I raise this PR, is there any other limitation we want to place on wikipedia imports? Or any unforeseen consequence of removing the flag? @symroe

PR #1683, opened 2023-08-22
Import wikipedia bios  with cron expressions

VirginiaDooley:
Closes https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1625

Although the WCIVFController Lambda Function has been running as expected, two jobs appear to be missing from the Cloudwatch logs: `import_wikipedia_bios` and `import_parties`. Both of these jobs have a rate expression `Schedule: rate(1 day)`. Here, I've changed it to run with a cron expression `Schedule: cron(0 14 * * *)`. Although the AWS docs do not indicate that rate expressions are deprecated, all the jobs with cron expressions appear to be running on schedule.

### DIFF
--- a/sam-template.yaml
+++ b/sam-template.yaml
@@ -151,7 +151,7 @@ Resources:
         ImportWikipediaBios:
           Type: Schedule
           Properties:
-            Schedule: rate(1 day)
+            Schedule: cron(0 7 * * *)
             Name: import-wikipedia-bios
             Description: Import wikipedia bio extracts
             Input: '{"command": "import_wikipedia_bios", "args": ["--current"]}'


Bekabyx (approved):
(Not a blocker!) Just wanted to check that the (0 14 * * *) cron is the one you actually want to use? In the PR description you've used (15 10 * * *) as your example so thought it best to ask 😁

VirginiaDooley:
> (Not a blocker!) Just wanted to check that the (0 14 * * *) cron is the one you actually want to use? In the PR description you've used (15 10 * * *) as your example so thought it best to ask 😁

Picked a daily time at random and updated above as well, but thanks for checking 👍

Issue #935, opened 2021-11-01
Display turnout & spoilt ballot info


pmk01:
We collect turnout & spoilt ballots but don't display it on WCIVF. It's been mentioned a couple of times in feedback that people would like to see this, so, maybe we should?

VirginiaDooley:
Where are we collecting this? I am aware that it's an option to collect for results bot but I don't think we are storing it at the moment. In which case, this is: collect & display

VirginiaDooley:
> It's blank for resultsbot, but it's an option in the manual result entry and is [included in the CSV](https://candidates.democracyclub.org.uk/uk_results/).

So it is! Where would you like this info displayed? Previous elections table?

pmk01:
No, on the ballot page, something like this (I'll leave it to you to order/design it):

![Screenshot 2023-08-24 111531](https://github.com/DemocracyClub/WhoCanIVoteFor/assets/61510609/8d9cb6b8-888d-469b-a0b9-7542a3f872e4)

"ballots issued" = turnout (raw number). All the other numbers should be under that name in the data we collect.

Issue #1663, opened 2023-07-21
Single column list on all ballots?


pmk01:
We currently show a double column candidate list on WCIVF for users of computer monitors... but this isn't what the ballot paper will actually look like. Should we switch to a single column?

jf1:
My suggestion was just for the 321-sorted view, once results have been entered. Before then, with abc-sorting, the default 2-column view looks natural enough to leave as-is.

PR #1685, opened 2023-08-24
Feature/add more election results data

VirginiaDooley:
Closes https://github.com/DemocracyClub/WhoCanIVoteFor/issues/935 and https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1663
Dependent on https://github.com/DemocracyClub/yournextrepresentative/pull/2174

This work adds election results for past elections, when results are available.

- [x] Add turnout, ballot papers issues, electorate and spoilt ballots to `Election` model
- [x] Display ^^ results data in a new table on the ballot template as shown below, only when they exist
- [x] Add Welsh translations
- [x] Import this data as part of the `import_ballots` process
- [x] Add election results to YNR ballot serializer
- [x] Switches candidate lists to a single column

To test:
- run migrations
- run the ballot importer for recent elections with `python manage.py import_ballots --current`
- visit a [recently past election](http://localhost:8000/elections/local.st-albans.marshalswick-east-jersey-farm.by.2023-08-17/marshalswick-east-jersey-farm/) with results to view a new table similar to the one shown below

Screenshot 2023-09-19 at 4 41 03 PM

Screenshot 2023-09-19 at 4 41 19 PM

---
  - https://app.asana.com/0/0/1205303544790826


VirginiaDooley:
@pmk01 Would you prefer the table not appear at all if results are not available? Or should it read, as is the case currently, eg "Electorate not available"?

VirginiaDooley:
This works well with current ballots, but for a full import, I hit an issue with formatting.  For example: https://candidates.democracyclub.org.uk/api/next/results/local.reading.norcot.by.2023-08-03/ which points to an issue with the way I've formatted it in the serializer or the way it's arriving via `resultsbot`

pmk01:
> @pmk01 Would you prefer the table not appear at all if results are not available? Or should it read, as is the case currently, eg "Electorate not available"?

We should not show the table until we have some results.

It will be common for us to lack some numbers. It's often the case that councils only publish turnout percentage and nothing else, for example. We could just leave the field blank if we don't have the number? I think this is preferable to "[thing] not available".

VirginiaDooley:
> > @pmk01 Would you prefer the table not appear at all if results are not available? Or should it read, as is the case currently, eg "Electorate not available"?
> 
> We should not show the table until we have some results.
> 
> It will be common for us to lack some numbers. It's often the case that councils only publish turnout percentage and nothing else, for example. We could just leave the field blank if we don't have the number? I think this is preferable to "[thing] not available".

I think we have to have something in the box if a data point isn't available - if nothing else to maintain the formatting of the table, is "None" ok?

pmk01:
Sorry, no, 'none' wouldn't work becuase turnout: "none" doesn't make any sense. If we have nothing at all, does the column disappear? I am fine with that if the case.

symroe:
Sorry to throw a spanner in the works late on, but would it be better for the table to be flopped 90°?

So we have

```
a: 1
b: 2
c: 3
```
rather than 

```
a    b    c
1    2    3

```

Thinking about mobile interfaces mostly,

VirginiaDooley:
> Sorry to throw a spanner in the works late on, but would it be better for the table to be flopped 90°?
> 
> So we have
> 
> ```
> a: 1
> b: 2
> c: 3
> ```
> 
> rather than
> 
> ```
> a    b    c
> 1    2    3
> ```
> 
> Thinking about mobile interfaces mostly,

Agree. Better? Or both and make it responsive based on device? 

Mobile
Screenshot 2023-09-19 at 12 35 57 PM

Desktop 
Screenshot 2023-09-19 at 12 35 32 PM

wcivf/assets/scss/style.scss
```diff
@@ -156,3 +156,10 @@ input[type=radio] {
 .ds-status-message p {
   color: $white;
 }
+
+#results-table {
```
symroe, line 160: I think this can go now?

symroe (approved):
One comment and I think the commits need cleaning up, but this is good to go.

One thing I've not tested is the `has_results` function in reality. It is possible that we have candidate results but no values in the listed fields?

wcivf/assets/scss/style.scss
VirginiaDooley: Still using this.

wcivf/apps/elections/models.py
```diff
@@ -239,6 +243,19 @@ def pluralized_division_name(self):
...
+    @property
+    def has_results(self):
+        """
+        Returns a boolean for if the election has results
+        """
+        return bool(
+            self.spoilt_ballots
+            or self.ballot_papers_issued
+            or self.turnout
+            or self.electorate
```
VirginiaDooley, line 255: I think this is what you suggested? @symroe

symroe: I think the last clause can also be a n `or`?

VirginiaDooley: I'm not sure about that; I didn't think we wanted to display the table unless at least one of the above fields are not `None`.

VirginiaDooley: I've renamed this for clarity `has_election_results`.

PR #1686, opened 2023-08-24
Feature/cancellation reason

Bekabyx:
[Asana](https://app.asana.com/0/1204880927741389/1205310969856272)
Ref https://github.com/DemocracyClub/EveryElection/issues/1972 

Added cancellation reason to WCIVF and update existing cancellation reason logic. Devs.DC is really only used in WCIVF to get ballots from a postcode search, so instead I've added cancellation reason to the EE importer and modified the logic from there.

There are some frontend changes here as well on the ballot pages. The text is all pretty similar to what it was before, but the underlying logic has changed. E.g.

Screenshot 2023-09-21 at 20 32 55

I also updated the property `short_cancelled_message_html` to use the new logic. 

I've left the metadata as a fallback for these logic checks whilst we transition but happy to remove it if that's preferred. I figured that for elections cancelled in exceptional circumstances though, keeping it might prove useful.

Unfortunately this is a pain to test locally due to tight project coupling. Instructions:
- bring up WCIVF, YNR and EE on separate ports 
- run the WCIVF migrations
- make sure that the `*_BASE` variables are set in local config to match the ports you've brought these up on locally
- manually set an election's cancellation reason in EE (either via the frontend or the database)
- for the same election, manually set the ballot as cancelled in YNR
- run the ballot importer on WCIVF (use the `--current` flag to save time)
- find the election whose cancellation reason you set in the WCIVF frontend
- you should see a cancellation message on the ballot page (variable, depending on what cancellation reason you set)

```[tasklist]
### PR Checklist
- [x] Note what card in Asana this work relates to
```


symroe:
This looks fine, but we need to actually use the newly imported data and replace the template logic that's in at the moment. Can you add this to the PR plesae?

VirginiaDooley (approved):
Tested and confirmed this works as expected 👍

symroe (approved):
Nice one, I think this is good to go! :tada:

PR #1690, opened 2023-09-05
Only show about section for current candidates

VirginiaDooley:
Closes https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1625

We limit wikipedia imports to current candidates to be nice to Wikipedia's API. To reduce inaccuracies, this change only shows the wikipedia excerpt for current candidates. The link to the wikipedia page still appears in the contact card for all candidates, where it exists. 

Screenshot 2023-09-05 at 11 22 55 AM

To test, `runserver` and navigate to any past candidate such as `person/4546/boris-johnson`

---
  - https://app.asana.com/0/0/1205303545552858


pmk01 (commented):
Aha, mystery solved. Agree with your solution, thanks!

PR #1695, opened 2023-09-18
Update no election text logic

### LINKED ISSUE #1694: 'No election' text unexpectedly displaying on multi-ballot view
In the screenshot below, the text following "There may be parish, town or community council elections in some areas" should only appear on postcodes which have no upcoming elections. For some reason it's displaying when a postcode has more than one ballot.

Screenshot 2023-09-14 144830

Bekabyx:
Closes #1694 

The logic for this was totally broken, referencing variables that seemingly weren't in the page payload. As this message should only be displayed when there are no elections to show, I've simply moved the html up to the same conditional block that is responsible for the "No elections in your area" text.

This means there's a slight change in text order on the "No elections" style page, but imo this looks fine.

To test:
- Run `manage.py import_ballots --current` if you're not totally up to date 
- Use B773AE in the postcode search
- You should see the Elections in Your Area card without the "Local and devolved elections..." message

```[tasklist]
### PR Checklist
```

### DIFF
--- a/wcivf/apps/elections/templates/elections/includes/inline_elections_nav_list.html
+++ b/wcivf/apps/elections/templates/elections/includes/inline_elections_nav_list.html
@@ -4,6 +4,8 @@
     <div class="ds-card-body">
         {% if postelections.count == 0 %}
             <h2>{% trans "We don't know of any upcoming elections in your area." %}</h2>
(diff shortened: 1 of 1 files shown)


VirginiaDooley (approved):
That postcode didn't work for me even after an import, but I was able to look at an upcoming election and did not see the text as expected. 👍

PR #1701, opened 2023-09-26
Set url content

Bekabyx:
Peter reported an issue with WCIVF links [posted to Mastodon](https://mastodon.me.uk/@DemocracyClub/111125385497019926) not routing to the correct page. In this case it's the `election_view` not being configured properly, but I decided to update both that and `ballot_view` so we don't have to come back again to fix that too.

Full disclosure, I don't know if this is going to work. That said, I can't make it any more broken than it already is and local testing options are limited 🤷  

I verified this by running the project, going to `http://localhost:8000/elections/parl.2023-10-19/uk-parliament-elections/`, viewing the page source, and verifying that the `<head>` contains `<meta property="og:url" content="http://localhost:8000/elections/parl.2023-10-19/uk-parliament-elections/">` (with the correct URL). Hopefully that's enough! 

Any page (except postcode pages) should now have the complete url because I updated it in the base template. Postcode pages are overridden with their former default (`CANONICAL_URL`) to maintain the same behaviour as before.

```[tasklist]
### PR Checklist
```

### DIFF
--- a/wcivf/apps/elections/templates/elections/postcode_view.html
+++ b/wcivf/apps/elections/templates/elections/postcode_view.html
@@ -7,6 +7,7 @@
 {% block page_description %}{% include "elections/includes/_postcode_meta_description.html" %}{% endblock page_description %}
 {% block og_title_content  %}{% include "elections/includes/_postcode_meta_title.html" %}{% endblock og_title_content  %}
 {% block og_description_content %}{% include "elections/includes/_postcode_meta_description.html" %}{% endblock og_description_content %}
+{% block og_url_content %}{{ CANONICAL_URL }}{% endblock og_url_content %}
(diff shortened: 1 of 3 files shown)


VirginiaDooley:
Can I suggest we push this to staging to test fully? Then you can use a tool such as https://www.opengraph.xyz/ to check links and previews, etc.

VirginiaDooley (approved):

VirginiaDooley:
> Can I suggest we push this to staging to test fully? Then you can use a tool such as https://www.opengraph.xyz/ to check links and previews, etc.

We discussed and happy to approve this.

Bekabyx:
Going to merge this now as I believe I've sufficiently addresses Sym's comments and he's out poorly for the foreseeable.

PR #1702, opened 2023-09-26
Bump DC signup form

symroe:
Bump to version `3.0.0`


VirginiaDooley:
Looking at v3.0 of dc_signup_form and I'm still not sure what change needs to be made here?

symroe:
> Looking at v3.0 of dc_signup_form and I'm still not sure what change needs to be made here?

I didn't realise the server part was installed here. It shouldn't have been in this project, so just pushed another commit to remove it

VirginiaDooley (approved):

Issue #1719, opened 2023-10-25
General results info is duplicated between all `PostElection` pages that share a parent election.


GeoWill:
When there are two by ballots (`PostElection` objects) on the same day for the same organisation (eg parl.tamworth.by.2023-10-19 & parl.mid-bedfordshire.by.2023-10-19) they are both pulling `electorate`, `ballot_papers_issued`, `turnout` and `spoilt_ballots` from the `Election` object (i.e. `parl.2023-10-19`). This means it is duplicated between those by elections, even though they have different information for those fields in YNR. 

I suspect the solution is to move those fields onto the `PostElection` object, but I guess there might be a reason this wasn't done originally?

GeoWill:
https://whocanivotefor.co.uk/elections/local.haringey.2023-10-04/haringey-local-election/

VirginiaDooley:
> When there are two by ballots (`PostElection` objects) on the same day for the same organisation (eg parl.tamworth.by.2023-10-19 & parl.mid-bedfordshire.by.2023-10-19) they are both pulling `electorate`, `ballot_papers_issued`, `turnout` and `spoilt_ballots` from the `Election` object (i.e. `parl.2023-10-19`). This means it is duplicated between those by elections, even though they have different information for those fields in YNR.
> 
> I suspect the solution is to move those fields onto the `PostElection` object, but I guess there might be a reason this wasn't done originally?

This sounds right. I don't think this was intentional. I'll work on the fix, thanks.

PR #1720, opened 2023-10-25
Move results to postelection

VirginiaDooley:
Closes https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1719

To test: 

- Run the ballot importer
- Run the server and check that the results for both the post elections related to `local.haringey.2023-10-04/haringey-local-election/` are unique.

---
  - https://app.asana.com/0/0/1205782782568392


VirginiaDooley:
While I've fixed the importer and moved the results to the `PostElection` model, I can't figure out why the election listed in the issue isn't importing.

VirginiaDooley:
> While I've fixed the importer and moved the results to the `PostElection` model, I can't figure out why the election listed in the issue isn't importing.

Update: I've run the re-run migrations with data imported while on `master`. I've then rerun the ballot importer. The missing elections are now present, but the results still have not updated for the Haringey post elections. I have confirmed results appear for other elections with more than one post election.

wcivf/apps/elections/models.py
```diff
@@ -417,12 +400,29 @@ class PostElection(TimeStampedModel):
...
+    def has_results(self):
+        """
+        Returns a boolean for if the election has results
+        """
+        return bool(
+            self.spoilt_ballots
+            or self.ballot_papers_issued
+            or self.turnout
+            or self.electorate
+            or self.postelection_set.filter(personpost__elected=True).exists()
```
GeoWill, line 423: Given that this is on the `PostElection` model, I don't think `self.postelection_set` is referring to anything. An `Election` object had a `postelection_set` because there's a many-to-one (aka foreign key) relationship from `PostElection` to `Election`. 

TLDR: I think this line should be `self.personpost_set.filter(elected=True)`.

More broadly this property is a bit confusing to me because there seems to be 5 different things that are sufficient to show results - is it the case that we want to show results when any one of these is true? have you checked that this makes sense on the front end (don't necessarily need to write tests for all outcomes, more just sanity checking it.

symroe: This is partly my fault...

The property is called `has_results`, so I was worried that future us might use it to, e.g change a template to indicate that the election was over (or something). In other words, `has_results` suggests some sort of interface contract.

Seeing as the first 4 things are optional values that can all be `None`, there could be a situation where we _do_ have results, but just no values in these fields. 

This is why I suggested adding the final check: if someone was elected then we surely have results, and if someone wasn't then we surely don't. The previous checks are just an early exit system to save a query.

VirginiaDooley: Changed with https://github.com/DemocracyClub/WhoCanIVoteFor/pull/1720/commits/c3539742fef29cebb258c263e5f6bfe66f5114a0

VirginiaDooley: We don't show the table item if the cell is blank.

symroe (approved):

PR #1730, opened 2023-11-15
Fixup for verb tenses and number of candidacies.

VirginiaDooley:
Ref https://app.asana.com/0/1204880927741389/1205938214595963/f

This PR includes:

- a grammar edit when an candidate with a past candidate "stood X times vs once vs twice"
- change to logic when a candidate has more than once candidacy; going forward the list will only include future candidacies and past ones will be visible in the table. 
- a fix to match verb tense when a candidate has one election marked as current.
- Translation update for the above edits

Screenshot 2023-11-21 at 11 33 39 AM

### DIFF
--- a/locale/cy/LC_MESSAGES/django.po
+++ b/locale/cy/LC_MESSAGES/django.po
@@ -13,7 +13,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2023-09-19 15:01+0100\n"
+"POT-Creation-Date: 2023-11-21 11:28+0000\n"
 "PO-Revision-Date: 2022-03-24 09:04+0000\n"
 "Last-Translator: Sym Roe <sym@talusdesign.co.uk>, 2022\n"
 "Language-Team: Welsh (https://www.transifex.com/democracy-club/teams/61326/"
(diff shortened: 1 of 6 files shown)


symroe:
I think the logic is good (although it's hard to think about!), but I have some comments about the way we're building the paragraph in the HTML:

At the moment the logic is at the word level (more or less). This works, but makes translations and parsing the logic harder than I think it might need to be.

On translations: some languages might put "twice" before the verb but "once" after (I have no idea!). This is impossible to do correctly at the moment because the words are inline.

On parsing: it's hard to work out the grammar in all cases because of reading the logic at the same time.

I think we should embrace duplication (in this case) and just write out the line for each case.

e.g, turn this:
```jinja
<p>{{ person_name }}{% blocktrans %} stood for election{% endblocktrans %}
    {% if object.past_not_current_candidacies.count == 1%}
        {% blocktrans %} once.{% endblocktrans %}
    {% elif object.past_not_current_candidacies.count == 2 %}
        {% blocktrans %} twice.{% endblocktrans %}
    {% else %}
        {% blocktrans trimmed with past_candidacy_count=object.past_not_current_candidacies.count %} {{ past_candidacy_count }} times.{% endblocktrans %}
    {% endif %}
</p>
```

Into:


```jinja
{% if object.past_not_current_candidacies.count == 2 %}
  {% blocktrans %}
    <p>{{ person_name }} stood for election twice.</p>
  {% endblocktrans %}
{% else %}
  {% blocktrans trimmed with past_candidacy_count=object.past_not_current_candidacies.count count object.past_not_current_candidacies.count %}
    <p>{{ person_name }} stood for election once.</p>
  {% plural %}
      <p>{{ person_name }} stood for election {{ past_candidacy_count }} times.</p>
  {% endblocktrans %}
{% endif %}
```

This also lets us use the plural forms feature of `ugettext`, giving translators even more flexibility.

symroe (approved):

Issue #1731, opened 2023-11-15
No candidates elected uncontested pre-2022 are marked as such (and hence don't appear in the 'elected' CSV)


VirginiaDooley:
See eg https://whocanivotefor.co.uk/person/52701/mike-johnson

> Has this always been the case or is it the result of recent changes?
> I know pre-2019 election results are patchy, but am surprised by 2021 being missing.
> More to the point I am unsure how to fix it without unlocking and locking the profiles.
> Perhaps we only introduced uncontested in 2022 and we didn’t create a migration to deal with historic data?
> It looks like all the elections themselves have been marked as cancelled.
> it's just the candidates missing the elected flag

PR #1732, opened 2023-11-17
Mark elected when candidacy is uncontested

VirginiaDooley:
Ref https://github.com/DemocracyClub/yournextrepresentative/issues/2205

### DIFF
--- a/wcivf/apps/people/management/commands/import_people.py
+++ b/wcivf/apps/people/management/commands/import_people.py
@@ -192,6 +192,9 @@ class Command(BaseCommand):
                 ballot = PostElection.objects.get(
                     ballot_paper_id=ballot_paper_id
                 )
+                if not ballot.contested:
+                    candidacy["elected"] = True
+
             except PostElection.DoesNotExist:
                 # This might be because we've not run import_ballots
                 # recently enough. Let's import just the ballots for this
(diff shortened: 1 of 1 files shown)


symroe:
I'm not sure this project is the right place to set this...we want the candidate to be marked as elected in YNR as well, and this isn't going to change that. In fact, [the issue](https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1731) is with the CSV, and that's in YNR. 

The solution here is, I think, to make sure candidates are marked as elected when an election is marked as uncontested (and locked) (I think this is happening already, hence the bug being pre-2022). Then it's to make a data migration in YNR to fix the older elections.

VirginiaDooley:
> I'm not sure this project is the right place to set this...we want the candidate to be marked as elected in YNR as well, and this isn't going to change that. In fact, [the issue](https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1731) is with the CSV, and that's in YNR.
> 
> The solution here is, I think, to make sure candidates are marked as elected when an election is marked as uncontested (and locked) (I think this is happening already, hence the bug being pre-2022). Then it's to make a data migration in YNR to fix the older elections.

Makes total sense, will close this and move this issue to YNR 👍

VirginiaDooley:
> I'm not sure this project is the right place to set this...we want the candidate to be marked as elected in YNR as well, and this isn't going to change that. In fact, [the issue](https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1731) is with the CSV, and that's in YNR.
> 
> The solution here is, I think, to make sure candidates are marked as elected when an election is marked as uncontested (and locked) (I think this is happening already, hence the bug being pre-2022). Then it's to make a data migration in YNR to fix the older elections.

Makes total sense. I will close this and move the issue to YNR.

PR #1744, opened 2023-12-05
Hotfix/remove mailing list form

VirginiaDooley:
Ref https://app.asana.com/0/1205277833047542/1206064730900380/f

Screenshot 2023-12-13 at 10 18 07 AM copy


symroe:
Hmm, something's up with the button in the footer isn't it? I reckon this should either be a link, or we update the design system to remove the background?

symroe (approved):
Other than the button, this is all great. Thanks for setting up the redirect too!

VirginiaDooley:
> Hmm, something's up with the button in the footer isn't it? I reckon this should either be a link, or we update the design system to remove the background?

I just looked again at the design system and it looks like I forgot to add `class="ds-dark"` - but that assumes this should be a call to action. Should this be a button instead?

VirginiaDooley:
> > Hmm, something's up with the button in the footer isn't it? I reckon this should either be a link, or we update the design system to remove the background?
> 
> I just looked again at the design system and it looks like I forgot to add `class="ds-dark"` - but that assumes this should be a call to action. Should this be a button instead?

@symroe Better? 

Screenshot 2023-12-13 at 10 18 07 AM copy

symroe:
@VirginiaDooley Yes that's better. Have you pushed whatever you changed to get to that?

PR #1764, opened 2024-01-29
Catch and ignore BlankRowException

VirginiaDooley:
Ref https://democracy-club-gp.sentry.io/issues/4678125245/events/a90b37ba68b6401ea8ad46f00ed973ce/

Includes ruff reformatting.

### DIFF
--- a/wcivf/apps/ppc_2024/management/commands/import_2024_ppcs.py
+++ b/wcivf/apps/ppc_2024/management/commands/import_2024_ppcs.py
@@ -48,7 +48,8 @@ class CSVRow:
     def from_csv_row(cls, row: dict):
         party_id = clean_party_id(row.pop("Party ID", None))
         if not party_id:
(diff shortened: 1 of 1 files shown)


wcivf/apps/ppc_2024/management/commands/import_2024_ppcs.py
```diff
@@ -48,7 +48,9 @@ class CSVRow:
     def from_csv_row(cls, row: dict):
         party_id = clean_party_id(row.pop("Party ID", None))
         if not party_id:
-            raise BlankRowException()
+            with contextlib.suppress(BlankRowException):
```
symroe, line 51: I think we still want to raise the exception here, but catch it where we call `from_csv_row` in the loop below. At that point we can use `suppress`

symroe (approved):

PR #1769, opened 2024-02-12
Catch non-JSON errors from devs.DC API

symroe:
Fixes https://democracy-club-gp.sentry.io/issues/4245462722/events/50221bef31cd482582820ee8d96e2c82/?project=1426221

### DIFF
--- a/wcivf/apps/elections/devs_dc_client.py
+++ b/wcivf/apps/elections/devs_dc_client.py
@@ -9,7 +9,7 @@ class DevsDCAPIException(Exception):
     def __init__(self, response: requests.Response):
         try:
             self.message = {"error": response.json().get("message", "")}
-        except JSONDecodeError:
+        except (JSONDecodeError, AttributeError):
             self.message = ""
         self.status = response.status_code
         self.response = response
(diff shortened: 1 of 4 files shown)


wcivf/apps/elections/devs_dc_client.py
```diff
@@ -9,7 +9,7 @@ class DevsDCAPIException(Exception):
     def __init__(self, response: requests.Response):
         try:
             self.message = {"error": response.json().get("message", "")}
-        except JSONDecodeError:
+        except (JSONDecodeError, AttributeError):
             self.message = ""
         self.status = response.status_code
         self.response = response
```
GeoWill, line 15: Do these lines mean that a 404/`Not found` will be returned to the user?

GeoWill (approved):
I think this looks good. I guess it would be nice to be able to say "that's not a valid uprn", but I guess how to do that isn't simple, and is out of scope for fixing this.

wcivf/apps/elections/devs_dc_client.py
symroe: I think _any_ error is simply caught and a generic error is given on the home page

Issue #1537, opened 2023-03-24
Review card styling on postcode results page


VirginiaDooley:

pmk01:
Both @symroe and @VirginiaDooley have said that they want to remove the background card ("card within a card"). I agree, so we should do this.

VirginiaDooley:
Addressed with https://github.com/DemocracyClub/WhoCanIVoteFor/pull/1790

PR #1790, opened 2024-03-11
Not everything is a card

symroe:
Ref https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1537
Ref for broken format on PCC 'About this position' component (aka "explainers") https://github.com/DemocracyClub/WhoCanIVoteFor/pull/1807

- Remove the `ds-card` from a load o pages. This is especially noticeable on the ballot and postcode results pages on mobile. This change includes some general reformatting and spacing edits. 
- Put voter FAQs into details list and add anchors to those in the elections list
- Hide party emblem on candidate lists for small mobile devices 

Changes are available to view on https://dev.wcivf.club/

---
  - https://app.asana.com/0/0/1206863292550553

### DIFF
--- a/locale/cy/LC_MESSAGES/django.po
+++ b/locale/cy/LC_MESSAGES/django.po
@@ -13,7 +13,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-01-11 14:52+0000\n"
+"POT-Creation-Date: 2024-04-05 18:48+0100\n"
 "PO-Revision-Date: 2022-03-24 09:04+0000\n"
 "Last-Translator: Sym Roe <sym@talusdesign.co.uk>, 2022\n"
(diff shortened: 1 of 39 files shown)


VirginiaDooley:
> ⚠️ Not everything is a card!
> At the time of writing, Cards are used liberally across the Democracy Club sites. Don’t “turn everything up to eleven”; as a rule of thumb, only use a standalone Card once or twice in a page.
> 
> Only use multiple successive cards within a [Grid](https://democracyclub.github.io/design-system/components/grid). See an example on the [Grid](https://democracyclub.github.io/design-system/components/grid) page.

pmk01:
This is great, thanks! Had a browse around WCIVF, one thing, party lists need to be re-listed.
Screenshot 2024-03-25 144652

Could you show me what the front page list of by-elections looks like? There isn't anything on the dev site.

pmk01:
Whoops, sorry, I meant re-carded. As in, I think the parties should each have their own card, like candidates do on a ballot page.

pmk01 (approved):

PR #1792, opened 2024-03-18
Translations incl footer

VirginiaDooley:
- bumps versions of the design system and dc_utils to make the header logo and footer uniform with other DC sites. 
- updates translations site wide where possible https://dev.wcivf.club/
- records where external support is needed on translations [here](https://docs.google.com/document/d/1sjfm4qwMwvkYPnZgea0M3-mMkM-NGBQEFXLVaULYryo/edit?usp=sharing)

Screenshot 2024-04-18 at 10 36 17 AM

---
  - https://app.asana.com/0/0/1206929626417191

### DIFF
--- a/wcivf/templates/base.html
+++ b/wcivf/templates/base.html
@@ -71,7 +71,12 @@
             </aside>
         </form>
     {% endif %}
+
 {% endblock base_language_menu %}
+{% get_current_language as LANGUAGE_CODE %}
+{% if LANGUAGE_CODE == 'cy' %}
+    {% block language_code %}<em lang="cy">Pwy alla i bleidleisio ar gyfer?</em>{% endblock language_code %}
+{% endif %}
 
 {% block messages %}
     {% if messages %}
--- a/wcivf/templates/base.html
+++ b/wcivf/templates/base.html
@@ -111,6 +111,7 @@
             <li><a href="https://democracyclub.org.uk/blog/">{% trans "Blog" %}</a></li>
             <li><a href="https://twitter.com/democlub">{% trans "Twitter" %}</a></li>
             <li><a href="https://github.com/DemocracyClub">{% trans "GitHub" %}</a></li>
+            <li><a href="{% url 'ppc_2024:details' %}" >{% trans "Browse the list of people and parties standing in the next general election" %}</a></li>
         </ul>
     </div>
 {% endblock footer_menu %}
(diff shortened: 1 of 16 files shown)


wcivf/templates/base.html
```diff
@@ -71,7 +71,12 @@
             </aside>
         </form>
     {% endif %}
+
 {% endblock base_language_menu %}
+{% get_current_language as LANGUAGE_CODE %}
+{% if LANGUAGE_CODE == 'cy' %}
```
VirginiaDooley, line 77: I can't seem to implement this the way it's suggested in the design system. The subheader site title shows in both English and Welsh versions but I only want it to appear in the latter.

pmk01 (approved):
Diolch!

wcivf/templates/base.html
VirginiaDooley: @GeoWill I tried to follow what you have done for WDIV, but I'm not clear on how this works.

GeoWill: How is it not working? 
Have you tried printing `LANGUAGE_CODE` (or inspecting in debug toolbar) to check what it's coming out as - i.e. is it always returning `en`?

VirginiaDooley: The conditional isn't working, it's always on display in Welsh as shown here image

GeoWill: OK. So what are the results of the things I suggested?

VirginiaDooley: The value of `LANGUAGE_CODE` prints as `cy`

GeoWill: Huh. I checked it out and had a look. I think this is what you want: 
```
{% block language_code %}
    {% get_current_language as LANGUAGE_CODE %}
    {% if LANGUAGE_CODE == 'cy' %}
        <em lang="cy">Pwy alla i bleidleisio ar gyfer?</em>
    {% endif %}
{% endblock language_code %}
```
 I don't really understand why, but I guess something to do with template scopes.

GeoWill: There's no subtitle when it's English though. Is that what you want?

VirginiaDooley: I figured it out. The conditional needed to go inside the block tags. Thanks.

VirginiaDooley: Oh yes, you did too :)

Issue #1787, opened 2024-03-08
Independent candidate cards on list elections are too big


pmk01:
There's a biiiig scaling issue going on here.

https://whocanivotefor.co.uk/elections/gla.a.2024-05-02/london-assembly-elections-additional/

![Screenshot 2024-03-08 110247](https://github.com/DemocracyClub/WhoCanIVoteFor/assets/61510609/263d8e3e-1c2d-44ba-93e9-d66b3c12e5c6)

PR #1794, opened 2024-03-20
Resize Independent candidates

VirginiaDooley:
Closes https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1787

This change also decreases font size for grids in detail components as is the case with candidate card lists inside of parties groups.

Screenshot 2024-03-20 at 1 51 43 PM

---
  - https://app.asana.com/0/0/1206830986212028

### DIFF
--- a/wcivf/apps/elections/templates/elections/includes/_people_list_with_lists.html
+++ b/wcivf/apps/elections/templates/elections/includes/_people_list_with_lists.html
@@ -8,23 +8,22 @@
     {% if person_post.list.0.party.is_independent %}
         {#  This is a special case where we don't want to group independants, but show a single card per person  #}
         {% for pp in person_post.list %}
-            <div class="ds-candidate">
-                <div class="ds-candidate-body ds-stack-smallest">
-                    <h4 class="ds-candidate-name">
-                        <a href="{{ pp.person.get_absolute_url }}" class="ds-card-link">
+            <div class="ds-with-sidebar">
+                <div>
+                    <div class="ds-sidebar" style="flex-basis: 5rem">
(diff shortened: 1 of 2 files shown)


pmk01:
Actually, something related: on mobile the 'show n candidates' text is larger than the party name, and when I open them...

Screenshot_20240320-133418

VirginiaDooley:
> Actually, something related: on mobile the 'show n candidates' text is larger than the party name, and when I open them...
> 
> Screenshot_20240320-133418

I had to override the design system, but in this case I think it's warranted. 
Screenshot 2024-03-20 at 1 51 31 PM

VirginiaDooley:
> Member

How's this? 
Screenshot 2024-03-20 at 1 51 43 PM

VirginiaDooley:
> One thought: how would it look if the candidate's name was underneath the Independent title, same font size as the 'show candidates'? Thinking about long names/mobile

How's this? 
Screenshot 2024-03-20 at 1 51 43 PM

pmk01:
Thanks! What do you think? Does it look ok on monitor?

Card squashing might be a bit improved by this? https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1537

VirginiaDooley:
> Thanks! What do you think? Does it look ok on monitor?
> 
> Card squashing might be a bit improved by this? #1537

I think it's a good improvement. Fix for https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1537 is in the works with https://github.com/DemocracyClub/WhoCanIVoteFor/pull/1790

pmk01 (approved):

Issue #1708, opened 2023-10-04
Winners in non-FPTP election don't float to the top


pmk01:
https://whocanivotefor.co.uk/elections/local.highland.tain-and-easter-ross.by.2023-09-28/tain-and-easter-ross/
https://whocanivotefor.co.uk/elections/pcc.north-yorkshire.by.2021-11-25/police-fire-and-crime-commissioner-for-north-yorkshire/

![Screenshot 2023-10-04 100914](https://github.com/DemocracyClub/WhoCanIVoteFor/assets/61510609/ec9c489e-1baa-4412-8b91-732108567d7b)

PR #1796, opened 2024-03-21
Elected candidates should be at the top of the list

VirginiaDooley:
Closes https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1708

Screenshot 2024-03-21 at 2 04 06 PM

---
  - https://app.asana.com/0/0/1206889313171652

### DIFF
--- a/wcivf/apps/elections/views/mixins.py
+++ b/wcivf/apps/elections/views/mixins.py
@@ -82,16 +82,12 @@ class PostcodeToPostsMixin(object):
 
 class PostelectionsToPeopleMixin(object):
     def people_for_ballot(self, postelection, compact=False):
-        key = PEOPLE_FOR_BALLOT_KEY_FMT.format(
-            postelection.ballot_paper_id, compact
-        )
+        key = PEOPLE_FOR_BALLOT_KEY_FMT.format(postelection.ballot_paper_id, compact)
         people_for_post = cache.get(key)
         if people_for_post:
             return people_for_post
         people_for_post = postelection.personpost_set.all()
-        people_for_post = people_for_post.annotate(
-            last_name=LastWord("person__name")
-        )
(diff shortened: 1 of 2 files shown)


VirginiaDooley:
> This should already be done in [this block](https://github.com/DemocracyClub/WhoCanIVoteFor/blob/73623a44ac32b2ace5a4bdbac8b740c4df235a05/wcivf/apps/elections/views/mixins.py#L98-L108). Can you take a look at why it might not be working?

The following block that sorts by `votes_cast` is the culprit. I've made a change to sort by elected then votes cast in the same step and it seems to work as expected.

wcivf/apps/elections/views/mixins.py
```diff
@@ -100,12 +96,8 @@ def people_for_ballot(self, postelection, compact=False):
         else:
             order_by = ["name_for_ordering", "person__name"]
 
-        people_for_post = people_for_post.order_by(
-            F("elected").desc(nulls_last=True), *order_by
-        )
-        people_for_post = people_for_post.order_by(
-            F("votes_cast").desc(nulls_last=True), *order_by
-        )
+        people_for_post = people_for_post.order_by("elected", "-votes_cast", *order_by)
```
symroe, line 99: This makes sense. I think we need to keep the `F("elected").desc(nulls_last=True)` though, so this line should look like:

```
people_for_post = people_for_post.order_by(F("elected").desc(nulls_last=True), F("votes_cast").desc(nulls_last=True), *order_by)
```

(needs checking)

symroe:
Can you squash the commits down to one? I think this is ok otherwise!

symroe:
Oh and we don't need `by_elected` any more I don't think?

symroe (approved):

Issue #1768, opened 2024-02-04
Explainers aren't updating


pmk01:
It's true!

pmk01:
Actually, I think the problem here is that explainers are only updating when a candidate is added to the ballot.

VirginiaDooley:
@pmk01 Do you have an example of an explainer that wasn't updating?

pmk01:
Any PCC, or mayoral ballot,

pmk01:
https://whocanivotefor.co.uk/elections/pcc.devon-and-cornwall.2024-05-02/police-and-crime-commissioner-for-devon-cornwall/

VirginiaDooley:
Fixed with https://github.com/DemocracyClub/WhoCanIVoteFor/pull/1825

PR #1807, opened 2024-03-28
Hotfix/explainers not updating

VirginiaDooley:
Closes https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1768

In the `import_ballots --recently-updated` command, we currently check YNR for recently updated ballots, however explainers aren't needed/imported from EE to YNR so explainer changes will not filter through to WCIVF - unless a change is made to the ballot in YNR, such as adding a candidate to a ballot. In other words, an election update in EE will not be imported to WCIVF unless there is a recent ballot change related to that election in YNR. This PR changes that by checking EE for recent updates based on a `modified` timestamp and stores that timestamp on the WCIVF `Election` model as `ee_last_updated`. 

Ultimately, I didn't see a way to filter base on group_type as this would exclude, for example, the PCC mentioned in the issue and below. Although we may use `ee_last_modified` in the future, after discussing with Will, passing the modified date into the params didn't seem to offer any greater benefit.

I tested these changes with the following steps:

1. With a fresh db, I ran a full ballot import with `python manage.py import_ballots`
2. I ran the server and checked the explainer on `elections/pcc.devon-and-cornwall.2024-05-02/police-and-crime-commissioner-for-devon-cornwall/` (image below)
3. I made a very small formatting change on a PCC explainer in EE 
4. I ran an import again but this time with `--recently-updated` (to get EE changes from within the past 1 hour, although this could be reduced further to match the current cron schedule)
5. I ran the server and visited  `elections/pcc.devon-and-cornwall.2024-05-02/police-and-crime-commissioner-for-devon-cornwall/` to confirm the edit was imported.

A. After full import, before EE explainer edit, with an `<h3>` header

Screenshot 2024-04-03 at 2 06 52 PM

Screenshot 2024-04-03 at 2 06 43 PM

B. After EE explainer edit, after running the command again with `--recently-updated` flag, now with an `<h4>` header. 

Screenshot 2024-04-03 at 2 30 08 PM

Screenshot 2024-04-03 at 2 30 01 PM

---
  - https://app.asana.com/0/0/1206913268288733


wcivf/apps/elections/import_helpers.py
```diff
@@ -299,7 +305,9 @@ def do_import(self, params=None):
         if self.should_run_post_ballot_import_tasks:
             self.attach_cancelled_ballot_info()
 
-        if self.recently_updated:
+        qs = self.ee_helper.iter_recently_modified_election_ids(group_type="election")
+
+        if self.recently_updated or qs:
```
VirginiaDooley, line 310: I'm unsure this is the way to go as I think this will always return as True. I want to build a qs that checks for recently modified elections in EE.

wcivf/apps/elections/helpers.py
```diff
@@ -81,19 +81,20 @@ def get_data(self, election_id):
         self.ee_cache[election_id] = None
         return None
 
-    def iter_recently_modified_election_ids(self):
+    def iter_recently_modified_election_ids(self, group_type=None):
+        modified_date = timezone.datetime.now().date() - timezone.timedelta(days=20)
```
GeoWill, line 85: Why is this change (to now minus 20 days vs minus 1 hour) necessary? How often will this be run?

VirginiaDooley: I changed this for testing purposes, I wanted to be sure to something was returned to confirme it's working as hoped. I plan to change it back.

wcivf/apps/elections/import_helpers.py
```diff
@@ -576,10 +576,16 @@ def check_for_ee_updates(self):
         print("checking for recently updated EE")
         for election_id in self.ee_helper.iter_recently_modified_election_ids():
             print(election_id)
-            if self.ee_helper.ee_cache[election_id]["group_type"]:
+            group_type = self.ee_helper.ee_cache[election_id]["group_type"]
+            if group_type in ["election", "organisation"] or self.is_pcc_or_gla(
+                group_type, election_id
+            ):
                 election = Election.objects.get(slug=election_id)
```
GeoWill, line 583: I think this is going to fail when the election id is something like `local.2024-05-02` or  - i.e. the group type is an `election`. Because I don't think WCIVF has those types of ids.

GeoWill: I'm also not sure how explanations assigned to things like `mayor.london.2024-05-02` which are 'PostElections` are going to work, but that might be a separate issue.

VirginiaDooley: It seems simpler not to rely on group_type and go in the direction (which I think is where you were headed late yest) of splitting the election_ids and basing the choice of Election vs PostElection on the the org_type such as `org_type = election_id.split('.')[0]`. What do you think?

VirginiaDooley: Or more simply, 
if the `election_id` starts with local (and has 3 parts after being split), get the related PostElection
else, get the related Election

GeoWill: I think you probably do need to consider group type, because that tells you (in conjunction with what kind of election it is) whether it is an (WCIVF) `Election`, `PostElection` or neither. 

i.e. a local with `group_type` of  `election` is neither. `group_type` of organisation corresponds to an `Election`, `group_type` of `None` is a `PostElection`.
i.e. 
https://whocanivotefor.co.uk/elections/local.2024-05-02/ -> 404
https://whocanivotefor.co.uk/elections/local.exeter.2024-05-02/ -> list of ballots
https://whocanivotefor.co.uk/elections/local.exeter.alphington.2024-05-02/ -> single ballot

In the case of PCC I'm not sure how it works. 
https://whocanivotefor.co.uk/elections/pcc.west-midlands.2024-05-02/ -> single ballot
https://whocanivotefor.co.uk/elections/pcc.2024-05-02/ -> 404

I don't know if there is an WCIVF `Election` object that corresponds to PCC elections happening on 2024-05-02. Is it the case that there's a `PostElection` with no `Election` object, or do the `Election` objects and `PostElection` objects have the same ids? 

Tricky.

For the purposes of this PR I would break it down into the following steps. 

1) When there is an explainer in EE that is on something that exists in the WCIVF `Election` table (i.e. ids match) then update the description field. Check that this is working and picking up updates - i.e. when the explainer is modified in EE WCIVF should grab it, even though nothing will have changed in YNR.

2) Do something about EE explainers that are attached to EE elections with `group_type == 'election'`. I'm not sure what the right thing to do is here, it might be specific to the kind of election. So in the case of locals we probably want to attach the explainer on `local.2024-05-02` to every WCIVF `Election` with an id that starts with `local` and ends with `2024-05-02`. In the case of `mayor.2024-05-02`  I don't think that will be the case, because there are different explainers for different mayoral elections. I think enumerating the different cases in a way @pmk01 can have input is probably wise here.

VirginiaDooley: > > In the case of PCC I'm not sure how it works.
> https://whocanivotefor.co.uk/elections/pcc.west-midlands.2024-05-02/ -> single ballot
> https://whocanivotefor.co.uk/elections/pcc.2024-05-02/ -> 404
> 
> I don't know if there is an WCIVF Election object that corresponds to PCC elections happening on 2024-05-02. Is it the case that there's a PostElection with no Election object, or do the Election objects and PostElection objects have the same ids?
> 

Yes they do. It is tricky, esp since we don't know which fields have been updated in EE. Given that `import_metadata_from_ee` can take an election or a ballot argument, maybe we do both.

VirginiaDooley: After trying some matching, I realised that
`pcc` doesn't always have a `group_type` and it has both a ballot and an election in WCIVF. 

With [51c680c](https://github.com/DemocracyClub/WhoCanIVoteFor/pull/1807/commits/51c680c6230ec2a2cdf3cff916aa03a556dd791d), I've started by skipping `election_ids` that are a local with `group_type` of election, since it is neither an Election nor a PostElection in WCIVF. Then I've done simple matching of ids - I found this was more effective than splitting up `election_ids` and matching on `org_type` and/or date once we threw away what wasn't needed. But I've possibly oversimplified. I did test it and it has worked as expected on a `pcc` but need to do more testing today.

GeoWill: Right so I think that's step one, if you can also make use of the field that records the last time EE was checked. 

The limitation of that approach is if the explainer for all local elections is added to the `local.2024-05-02` then this will miss it. But I think that should be step 2 :)

wcivf/apps/elections/helpers.py
```diff
@@ -81,19 +81,23 @@ def get_data(self, election_id):
...
         return None
 
-    def iter_recently_modified_election_ids(self):
+    def iter_recently_modified_election_ids(self, group_type=None):
+        modified_date = timezone.datetime.now().date() - timezone.timedelta(days=3)
+
         params = {
-            "modified": timezone.datetime.now().date()
-            - timezone.timedelta(hours=1),
+            "modified": modified_date,
```
GeoWill, line 88: Is this change from t-1hr to 1-3days intentional?

wcivf/apps/elections/import_helpers.py
```diff
@@ -576,10 +577,41 @@ def check_for_ee_updates(self):
...
-                ballot = PostElection.objects.get(ballot_paper_id=election_id)
-                print(f"importing metadata from EE for {ballot}")
                 self.import_metadata_from_ee(ballot)
+            elif ballot and not election:
+                self.import_metadata_from_ee(ballot)
+            elif election and not ballot:
+                self.election_importer.import_metadata_from_ee(election)
+            if not election and not ballot:
+                print(f"neither election nor ballot {election_id} exists")
+                continue
```
GeoWill, line 617: Missing blank line at end of file.

wcivf/apps/elections/helpers.py
VirginiaDooley: Given the cron, I think this should be moved back to 1 hour, but I was adapting this as needed for testing purposes. I should not have committed the change.

VirginiaDooley:
Addressed in https://github.com/DemocracyClub/WhoCanIVoteFor/pull/1825

Issue #1246, opened 2022-05-25
Add 'report a problem with this page' button


pmk01:
This is super useful on WDIV, as people use it for bug reports they wouldn't otherwise put in the feedback form. 

I think we would catch a lot more candidates or other errors on WCIVF if we had a similar button here.

VirginiaDooley:
For reference: https://github.com/DemocracyClub/UK-Polling-Stations/tree/49cd5c23a2187d87488ecf66e30694c8a9c47daf/polling_stations/apps/bug_reports

VirginiaDooley:
@pmk01 Do we want this on specific pages or just in the footer?

pmk01:
Specific pages - in the way we already do on WDIV

PR #1808, opened 2024-03-29
Feature/add bug report form

VirginiaDooley:
Close https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1246

1. Update the choices and question in the likelihood of voting question
2. Move js and styles to feedback form template to address errors
3. Add a report a problem form that toggles visibility of text areas depending on choices
4. The bug report form is linked in the footer and appears within the feedback form on the postcode results page
5. Updates slack feedback with new voting choice
6. Removes unused js that is now handled in design system

Screenshot 2024-04-29 at 1 26 14 PM

---
  - https://app.asana.com/0/0/1206929626417215


VirginiaDooley:
@symroe  I’ve added this work to dev.wcivf.club. I’m temporarily commented out failing tests so that this could deploy; the tests that are failing relate to the redirect after submitting the form (it does not produce a 302). 

If you search a postcode, find the feedback form, and click the bug report option, you will see that there is still the submit option for the feedback form. Clicking this redirects to the feedback form template. I am not clear on 1) the best way to fix this behaviour 2) as well as add a “Send report” button to the feedback form.

VirginiaDooley:
@symroe Another issue I'm unsure about is why the 'Yes' button is active from page load. I assume it has something to do with the initial feedback post, but not sure the best way to reset it.

wcivf/apps/elections/templates/elections/election_view.html
```diff
@@ -85,7 +85,8 @@ <h3>{% blocktrans trimmed with title=object.pluralized_division_name|title %} {{
     {% include "elections/includes/_postcode_search_form.html" %}
 
     {% include "feedback/feedback_form.html" %}
-
-
+    <div id="report_form_container" style="display: none;">
+        {% include "bug_reports/report_form_inline.html" %}
+    </div>
```
symroe, line 90: See what Peter thinks, but I don't think we want this on anything but the postcode and ballot views?

wcivf/apps/elections/templates/elections/postcode_view.html
```diff
@@ -62,19 +62,22 @@
 
             {% if postelections %}
                 {% include "feedback/feedback_form.html" %}
+                <div id="report_form_container" style="display: none;">
```
symroe, line 65: I don't think we should inline the style at the point we include the template. If it's meant to be hidden by default, then that should be left up to the template we're including. Can you change this to just be the include line, and move the wrapping div to the included template?

wcivf/apps/feedback/templates/feedback/feedback_form.html
```diff
@@ -19,24 +22,104 @@ <h5>
...
             {{ feedback_form.comments }}
         </div>
         <button type="submit" class="ds-button-pink">{% trans "Send feedback" %}</button>
     </form>
 </div>
+
+{% block in_page_javascript %}
+    <script>
+        document.addEventListener("DOMContentLoaded", function(event) {
+            var feedback_form = document.forms[1]
```
symroe, line 53: `document.forms[1]` feels fragile here: what if we add another form? This will break silently.

Can you use `getElementById` to get this form?

wcivf/apps/feedback/templates/feedback/feedback_form.html
```diff
@@ -19,24 +22,104 @@ <h5>
...
+            var feedback_form_token = feedback_form.elements[1].value
+            var csrfmiddlewaretoken = feedback_form.elements[0].value
+            var found_useful_label = feedback_form.elements[3].labels[0]
+            // TO DO: the label should not be active on page load
+
+
+            var feedback_choices_container = document.getElementById('feedback_choices_container')
+            var voting_choices_container = document.getElementById('voting_choices_container')
```
symroe, line 61: Once we have the form, we can use `feedback_form.querySelector` to ensure we're getting the children of the form. This should keep it more encapsulated

wcivf/apps/feedback/templates/feedback/feedback_form.html
```diff
@@ -19,24 +22,104 @@ <h5>
...
+            var feedback_labels = feedback_choices_container.getElementsByTagName('label')
+            var voting_labels = voting_choices_container.getElementsByTagName('label')
+
+            /// Remove the comments box until the user has selected 'YES' or 'NO'
+            var found_useful = feedback_form.elements[3].value
```
symroe, line 66: Again, element indexing feels fragile. Can you use an ID or class selector?

wcivf/apps/feedback/templates/feedback/feedback_form.html
```diff
@@ -19,24 +22,104 @@ <h5>
...
+
+            var comments = document.getElementById('comments')
+            comments.hidden = true
+
+            var last_feedback_choice = document.getElementById('feedback_choices_container').lastElementChild
+            var problem_radio = '<input type="radio" name="report_problem" value="PROBLEM" required data-toggle="button" id="report_problem" />'
```
symroe, line 72: I'm not really sure what this code is doing. If it's showing / hiding form elements, then can you write the HTML out in the template and then just toggle a hidden flag on the elements? Making the form in JS feels like it might break something at some point.

Also, because the Django tests don't run JS, I think this will make testing harder. If the form works without JS enabled, we're able to write tests for it

pmk01:
Apologies if these have already been mentioned:

I wonder if we want 'no change' instead of/in addition to 'I always vote'? (sorry, should have thought about this earlier).

On WDIV 'report a problem' unselected yes or no, but it doesn't seem to do it here.

wcivf/apps/feedback/templates/feedback/feedback_form.html
VirginiaDooley: I overthought how this might done, fixed up now, thank you 👍

VirginiaDooley: The only remaining issues:

- test failure: submitting the form from `/feedback/` (which the user may never do since we don't link to it from anywhere) does not create a `Feedback` object
- active status of buttons when they are not selected. I've tried playing around with removing styles and making button links inactive, but have not figured this out yet.

VirginiaDooley:
@symroe way cleaner, thank you. Shall I clean up?

symroe:
Yep, over to you!

VirginiaDooley:
> Yep, over to you!

Ok. Looks like the js action for 'report a problem' is swapped with the 'yes' action so I'll fix that and squash.

PR #1811, opened 2024-04-01
Fix AMS text

pmk01:
I *think* the problem here is that we forgot to update the 'if' statement.

### DIFF
--- a/wcivf/apps/elections/templates/elections/includes/_how-to-vote.html
+++ b/wcivf/apps/elections/templates/elections/includes/_how-to-vote.html
@@ -65,7 +65,7 @@ PR-CL: Closed List
                 </p>
             {% endif %}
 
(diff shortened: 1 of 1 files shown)


symroe (commented):
@pmk01 this will work, in that it will display for AMS elections, however it will stop displaying on PR-CL elections. As this was only `europarl` and we don't show this 'how to vote' block on historic elections, I don't think it matters. 

Does that sound right to you?

pmk01:
Ah! of course. Good remembering. This feels like it may come back to bite us. I thought I was being clever; I will do it properly. (how do I cancel a PR? 'close'?)

VirginiaDooley:
Fixed in https://github.com/DemocracyClub/WhoCanIVoteFor/pull/1812

PR #1818, opened 2024-04-09
Open details elements when URL hash links to them

symroe:
The new TOC on WCIVF is great, but it's a bit of a odd experience clicking on a link to find a closed `details` element. 

This change opens them when they're clicked on :tada:


VirginiaDooley (approved):
Good idea!

PR #1821, opened 2024-04-11
All details elements should have h2 tags

VirginiaDooley:
All details elements should have h2 tags and be nested within `ds-details` except for the case of blockquotes

This also changes the size of the booklet to be more consistent with surrounding elements. 
Screenshot 2024-04-11 at 3 13 28 PM


pmk01 (approved):
This is good, though could be more subliminal.

PR #1825, opened 2024-04-15
Update recently modified Elections as well as child objects

VirginiaDooley:
This is a scaled down version of https://github.com/DemocracyClub/WhoCanIVoteFor/pull/1807 which implements the smallest change possible to ensure that recently modified Elections are updated along with their children, which could be either/both Election or PostElection objects in WCIVF.

### DIFF
--- a/wcivf/apps/elections/import_helpers.py
+++ b/wcivf/apps/elections/import_helpers.py
@@ -62,15 +62,11 @@ class YNRElectionImporter:
         ballot_paper_id = ballot_dict["ballot_paper_id"]
 
         # Look up the dict of possible weights for this election type
-        weights = charisma_map.get(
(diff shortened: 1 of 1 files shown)


wcivf/apps/elections/helpers.py
```diff
@@ -81,18 +81,23 @@ def get_data(self, election_id):
         self.ee_cache[election_id] = None
         return None
 
-    def iter_recently_modified_election_ids(self):
+    def iter_recently_modified_election_ids(self, group_type=None):
```
GeoWill, line 84: you've added and arg `group_type` to the method definitition here. But it isn't used anywhere. This makes me think it's not needed for this change?

symroe (approved):

PR #1830, opened 2024-04-23
Add missing internationalization tag

VirginiaDooley:
Closes https://democracy-club-gp.sentry.io/issues/5216141923/?project=1426221&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=3

TO DO fix 500 page


symroe:
Could you add these URLs to https://github.com/DemocracyClub/WhoCanIVoteFor/blob/master/wcivf/apps/core/tests/test_frontend.py#L43, so this sort of things gets picked up in tests in future?

PR #1836, opened 2024-04-25
Postal vote hand in - voter needs to fill out a form

VirginiaDooley:
Ref https://app.asana.com/0/1204880927741389/1206863706261602/f

> 
> For non-devolved elections: 
> "If you have a postal vote, you can hand it in at this polling station on election day up to 10pm."
Screenshot 2024-04-25 at 2 34 48 PM

> For parl., PCC, London and English locals: 
> "If you have a postal vote, you can hand it in at this polling station on election day up to 10pm. When handing in postal votes, you will need to complete a form, giving your name, address, how many postal votes you are handing in, and why you are handing in those postal votes."

Screenshot 2024-04-25 at 2 33 56 PM

> For Northern Ireland: 
> "Postal votes cannot be accepted at polling stations in Northern Ireland."
> Each of the above should end with a "Learn more about voting by post.", linking to https://www.electoralcommission.org.uk/voting-and-elections/ways-vote/how-vote-post

Screenshot 2024-04-25 at 2 35 37 PM

### DIFF
--- a/wcivf/apps/elections/models.py
+++ b/wcivf/apps/elections/models.py
@@ -468,6 +468,20 @@ class PostElection(TimeStampedModel):
         ).postal_vote_application_deadline
 
     @property
+    def postal_vote_requires_form(self):
+        matcher = PostalVotingRequirementsMatcher(
+            election_id=self.election.slug, nation=self.post.territory
+        )
+
+        voting_requirements_legislation = (
+            matcher.get_postal_voting_requirements()
+        )
+
+        if voting_requirements_legislation == "EA-2022":
+            return True
+        return False
+
+    @property
     def is_mayoral(self):
         """
         Return a boolean for if this is a mayoral election, determined by
--- a/fixtures/vcr_cassettes/test_mayor_elections.yaml
+++ b/fixtures/vcr_cassettes/test_mayor_elections.yaml
@@ -15,7 +15,7 @@ interactions:
       "dates": [{
           "date": "2018-05-03",
           "polling_station": {
-              "polling_station_known": false
+              "polling_station_known": true
           },
           "ballots": [{
                   "ballot_paper_id": "local.tower-hamlets.bow-east.2018-05-03"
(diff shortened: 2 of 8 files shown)


pmk01:
Thanks. Can we change the order of the 'read more' link, so it comes immediately after 'you don't need your poll card'?

pmk01 (approved):

VirginiaDooley:
Note to self: use territory code to get country.

wcivf/apps/elections/models.py
```diff
@@ -467,6 +472,20 @@ def postal_vote_application_deadline(self):
             self.ballot_paper_id, self.post.territory
         ).postal_vote_application_deadline
 
+    @property
+    def postal_vote_requires_form(self):
```
symroe, line 476: This logic can be replaced by the matcher code. See example here: https://github.com/DemocracyClub/uk-election-ids/blob/master/tests/test_meta_data_matcher.py#L30

wcivf/apps/elections/models.py
```diff
@@ -467,6 +467,18 @@ def postal_vote_application_deadline(self):
...
+        matcher = PostalVotingRequirementsMatcher(
+            election_id=self.election.slug
+        )
+        voting_requirements_legislation = (
+            matcher.get_postal_voting_requirements()
+        )
+        if voting_requirements_legislation == "RPA2000":
```
symroe, line 478: It's `EA-2022` that requires the form, not `RPA2000`

wcivf/apps/elections/models.py
```diff
@@ -467,6 +467,18 @@ def postal_vote_application_deadline(self):
             self.ballot_paper_id, self.post.territory
         ).postal_vote_application_deadline
 
+    @property
+    def postal_vote_requires_form(self):
+        matcher = PostalVotingRequirementsMatcher(
+            election_id=self.election.slug
```
symroe, line 473: I think you'll need to pass nation in here

symroe (approved):

Issue #1689, opened 2023-09-05
Fix deployment fail alerts (key issue)


VirginiaDooley:
This was working earlier this year but now has a error message: 

Posting Status
Checking For JQ + CURL: Debian
Sending to Slack Channel: ***********
{
  "ok": false,
  "error": "account_inactive"
}

PR #1852, opened 2024-05-02
Hotfix/slack notifications

VirginiaDooley:
Ref https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1689

I tested this config by temporarily adding a failing test and the notifications setup to the test and build job.
Screenshot 2024-05-14 at 2 11 44 PM

---
  - https://app.asana.com/0/0/1207254815979868

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@
 #
 version: 2.1
 orbs:
-  slack: circleci/slack@4.12.5
+  slack: circleci/slack@4.13.2
 
 executors:
   py310:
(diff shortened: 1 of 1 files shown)


symroe:
If you're using webhooks then do we still need the Slack orb?

And if the API has changed, then does that mean the orb no longer works? It was last updated in March, so I suspect it's still ok. 

Can you either explore the org more (It's likely to be more robust when working) or remove it and just use the webbook?

VirginiaDooley:
> If you're using webhooks then do we still need the Slack orb?
> 
> And if the API has changed, then does that mean the orb no longer works? It was last updated in March, so I suspect it's still ok.
> 
> Can you either explore the org more (It's likely to be more robust when working) or remove it and just use the webbook?

Started from scratch with the orb setup and it worked!

symroe (approved):

PR #1867, opened 2024-05-09
Hotfix/always show council

VirginiaDooley:
Ref https://app.asana.com/0/1204880927741389/1207218844713564/f

[Mayor of West Midlands candidates in WV12 4QW (whocanivotefor.co.uk)](https://whocanivotefor.co.uk/elections/WV12%204QW/)

Always show council contact details regardless of whether a polling station is known. 

Screenshot 2024-05-09 at 4 04 12 PM


symroe:
There are two more cases to consider here:

1. If the 'council' in the API response is EONI, then we should say "contact EONI" rather than "your local council"
2. If there is different registration to council contact details, then we should show both.

See the logic here:

https://github.com/DemocracyClub/ec-postcode-lookup-pages/blob/main/postcode_lookup/templates/includes/organisation_contact.html

I expect we can just use the wording we use there.

VirginiaDooley:
> There are two more cases to consider here:
> 
> 1. If the 'council' in the API response is EONI, then we should say "contact EONI" rather than "your local council"
> 2. If there is different registration to council contact details, then we should show both.
> 
> See the logic here:
> 
> https://github.com/DemocracyClub/ec-postcode-lookup-pages/blob/main/postcode_lookup/templates/includes/organisation_contact.html
> 
> I expect we can just use the wording we use there.

In the second case, what do you reckon is the most effective comparison: address, postcode, name, etc? Simply comparing the objects, as seen in the above screen shot, might not be very effective as the council has more properties than the registration so will always differ.

symroe (approved):

Issue #1858, opened 2024-05-03
Hide voter id and where to vote sections after the election has passed


VirginiaDooley:
<img width="735" alt="Screenshot 2024-05-03 at 9 17 02 PM" src="https://github.com/DemocracyClub/WhoCanIVoteFor/assets/7017118/c2cb68ae-ae1f-41d8-a2a1-0f475489f16f">
<img width="939" alt="Screenshot 2024-05-03 at 9 17 16 PM" src="https://github.com/DemocracyClub/WhoCanIVoteFor/assets/7017118/a0051315-e2ea-41d2-8c2f-cf7cb4956aa0">

PR #1872, opened 2024-05-13
Hotfix/dont show voter info after election day

VirginiaDooley:
Ref https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1858

This change adds to the context the ability to identify whether any postelections in a list are in the future, and in that case, shows the 'get ready to vote' anchors and details below. 
Screenshot 2024-05-13 at 11 18 14 AM
Screenshot 2024-05-13 at 11 18 20 AM

---
  - https://app.asana.com/0/0/1207297516776629

### DIFF
--- a/wcivf/apps/elections/templates/elections/includes/inline_elections_nav_list.html
+++ b/wcivf/apps/elections/templates/elections/includes/inline_elections_nav_list.html
@@ -26,13 +26,13 @@
         {% endif %}
         {% for election_group in header_elections_by_date %}
 
-            {% if election_group.list.0.past_date %}
-                {% ifchanged election_group.list.0.past_date %}
+            {% if election_group.list.0.election.in_past %}
+                {% ifchanged election_group.list.0.election.past_date %}
                     <h3>{% trans "Recently past elections" %}</h3>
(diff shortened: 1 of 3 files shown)


wcivf/apps/elections/templatetags/postcode_tags.py
```diff
@@ -23,3 +23,10 @@ def convert_str_date(value):
 @register.filter(name="totime")
 def convert_str_time(value):
     return datetime.strptime(value, "%H:%M:%S").time()
+
+
+@register.filter(name="anyfuturedates")
```
symroe, line 28: Can you explain why you used a template tag here? I would have thought adding this to the context in the view would have been better, but maybe I'm missing something?

The reason I'm cautious about template tags is that they can quickly cause `n+1` bugs. For example, this tag is assuming that we've called `select_related` on `postelection.election`, and if we've not it could cause loads of queries if not uses correctly.

VirginiaDooley: Great point, will make the edit. Thanks.

symroe (approved):

PR #1903, opened 2024-05-28
Import NationalParty

VirginiaDooley:
Ref https://app.asana.com/0/1204880927741389/1206830990963594/f

This PR creates a `GeneralElectionPartyImporter`, `GeneralElectionParty` and related management command that is all based on `LocalParty` model and importer, with some differences. There is a lot of repetition here, but naming is clearer. 

TO DO: 

- [x] This work does not (yet) include adding any general election party logic to the remainder of the code or rendering the new fields. Happy to do that next if the work in progress seems like the right direction. 
- [x] Please check the parties and manifesto card to make sure that the word 'local' is removed.

Screenshot 2024-06-04 at 12 16 56 PM
Screenshot 2024-06-04 at 12 16 34 PM


pmk01 (approved):

symroe (approved):
This seems reasonable but I think you should talk Peter through it on a screen share to make sure it's looking ok (if you've not already done so)

VirginiaDooley:
> This seems reasonable but I think you should talk Peter through it on a screen share to make sure it's looking ok (if you've not already done so)

Done. Is there anything else I need to do in AWS to ensure the import is run as scheduled?

PR #1940, opened 2024-06-11
Hotfix/city of london

VirginiaDooley:
Ref https://app.asana.com/0/1206157967511125/1207499175901398/f

1/2 by elections in the City of London which had different polling station than the GE polling station was uncontested so this change just covers the difference in polling times. 

http://127.0.0.1:8000/elections/EC4M%207BU/100020000026/

Screenshot 2024-06-18 at 1 10 57 PM

### DIFF
--- a/wcivf/apps/elections/models.py
+++ b/wcivf/apps/elections/models.py
@@ -76,7 +76,7 @@ class Election(models.Model):
         return self.election_date < datetime.date.today()
 
     @property
-    def is_city_of_london(self):
+    def is_city_of_london_local_election(self):
         """
         Returns boolean for if the election is within City of London district.
         The city often has different rules to other UK elections so it's useful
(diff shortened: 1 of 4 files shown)


symroe:
Two things here:

1. We [already have](https://github.com/DemocracyClub/WhoCanIVoteFor/blob/d8f6418fa916213b0f092c0fb1393c5f4985c287/wcivf/apps/elections/views/postcode_view.py#L152) code for special casing this. See [this issue](https://github.com/DemocracyClub/WhoCanIVoteFor/issues/441) for more 
2. The issue here is that there are two polling stations, one set open 7-10 and the other 8-8. We can't just change the text to say 8-8

symroe:
It looks like this is still saying _either_ 8-8 _or_ 7-10? We need to say 7-10 _and_ 8-8 in this case. @pmk01 do you have some example text already?

VirginiaDooley:
> It looks like this is still saying _either_ 8-8 _or_ 7-10? We need to say 7-10 _and_ 8-8 in this case. @pmk01 do you have some example text already?

Sorry, I think this bit was lost in a rebase; I've added a fixup _with the intent to_ list each election, polling station and the corresponding open/close times.

VirginiaDooley:
I've also just realised that we are basing what to show for the city of london based on whether there multiple ballots for TODAY. I think today isn't relevant in this context; we want to know if there two on the same date. I've added another fixup to reflect this.

pmk01 (approved):

PR #1951, opened 2024-06-18
Hotfix/add menu to single ballot

VirginiaDooley:
This anchor list was added to the multi-ballot view in May but had not been added to the single ballot view. 

Screenshot 2024-06-18 at 12 08 47 PM


pmk01 (approved):

VirginiaDooley:
In other words, in every case, post_name does not return "ward", "constituency", etc.

PR #1956, opened 2024-06-19
Fix leaflet importing

symroe:
Before this, the importer wasn't paging properly


VirginiaDooley (approved):
Just FYI, this was quite slow to import locally  - some people took approx 15 seconds to check while others were quick. In any case, the command ran without error.

symroe:
Yeah, I found this too. I'll look out for timeout errors in production

PR #2039, opened 2024-08-13
fix ruff CI

chris48s:
When I checked this out locally I was getting fails running pytest that didn't seem to reproduce in CI.

As written, the CI will never fail due to ruff formatting errors because we were running `ruff format .` (which exits with code 0 even if files were written) before running pytest.

This means if there were formatting errors we just fix them in CI before checking.

In this PR I have modified the build to run `ruff format --check .` (which doesn't write anything but exits with non-zero if files would have been written). Pushed it to prove the build fails. Then I have pushed another commit which fixes the errors to get the build passing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,7 +78,7 @@ jobs:
           name: Ruff
           command: |
             . .venv/bin/activate
-            ruff format .
+            ruff format . --check
       - run:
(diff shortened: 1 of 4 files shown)


chris48s:
I'm going to just merge this one so I have passing tests locally

PR #2040, opened 2024-08-14
postgres 16/python 3.12

chris48s:
In general, this project requires fewer changes than EE and WDIV

- There is no node/npm usage at all
- No lambdas as far as I can tell, except for the cron jobs
- We're already using `ruff format` rather than `black` so the linting setup was already closer to where we want to be across projects


sam-template.yaml
```diff
@@ -362,7 +362,9 @@ Resources:
         # TODO create new role without packer permissions
         IamInstanceProfile:
           Name: 'CodeDeploy-EC2-Instance-Profile'
-        ImageId: 'ami-014a706cddda2a79a' # AMI number
+        # output of
+        # https://eu-west-2.console.aws.amazon.com/imagebuilder/home?region=eu-west-2#/images/arn%3Aaws%3Aimagebuilder%3Aeu-west-2%3A732292556707%3Aimage%2Feeimage-ubuntu%2F0.0.35%2F2/details
+        ImageId: 'ami-0c8d319e3a044e7da'
```
chris48s, line 367: It seems this project is built off the EE base image, rather than the generic one.
I found this surprising but `ami-014a706cddda2a79a` is definitely an EE image and https://github.com/DemocracyClub/WhoCanIVoteFor/commit/5fa46e477cb9a74ff4d03d429725d78bcda0b13d was the last commit touching this line.

INSTALL.md
```diff
@@ -81,17 +81,14 @@ GEOS_LIBRARY_PATH= '{YOUR PATH}/libgeos_c.1.17.0.dylib'
```
chris48s, line 81: There's some other bits of docs in this file that I reckon are a bit suspect.

Like do we really need all this stuff about GEOS/GDAL/PostGIS for this project? In CI we are just running plain old postgres. No GIS extensions at all.

wcivf/settings/base.py
```diff
@@ -42,8 +43,32 @@ def repo_root(*x):
...
+# Hosts/domain names that are valid for this site; required if DEBUG is False
+# See https://docs.djangoproject.com/en/1.5/ref/settings/#allowed-hosts
+ALLOWED_HOSTS = [
+    os.environ.get("FQDN", None),
+    "localhost",
+    "127.0.0.1",
+]
+
+
+def get_ec2_ip():
```
chris48s, line 55: I guess we could consider making this a shared function in a package? e.g: dc_django_utils. I think for now doing it right with a bit of copy & paste is a step forward.

INSTALL.md
chris48s: Also the setup instructions for importing data are b0rked

.circleci/config.yml
```diff
@@ -107,21 +107,23 @@ jobs:
...
-      - image: public.ecr.aws/sam/build-python3.10:latest
+      - image: public.ecr.aws/sam/build-python3.12:latest
     working_directory: ~/repo
     steps:
       - checkout
       - run:
           name: sam build
           command: |
+            dnf install -y postgresql-devel
+            pip install setuptools wheel
```
chris48s, line 118: I'm not too worried we now have to install steuptools and wheel here. That kinda tracks.

I'm a bit confused about why we get the `Error: pg_config executable not found` error without installing `postgresql-devel` because we're using psycopg2-binary which should be pre-compiled :confused:

chris48s:
Question: How can I check this on dev. How does https://dev.wcivf.club get deployed?

INSTALL.md
chris48s: https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2040/commits/d339d3417471c2fb66581e2470cf55c6315b4756

.circleci/config.yml
chris48s: This was because the version of psycopg2-binary we were using didn;t have python 3.12 weheels. Now fixed

wcivf/settings/base.py
```diff
@@ -42,8 +43,36 @@ def repo_root(*x):
...
+# Hosts/domain names that are valid for this site; required if DEBUG is False
+# See https://docs.djangoproject.com/en/1.5/ref/settings/#allowed-hosts
+ALLOWED_HOSTS = [
+    "localhost",
+    "127.0.0.1",
+]
+
+if os.environ.get("DC_ENVIRONMENT") and os.environ.get("FQDN"):
+    ALLOWED_HOSTS.append(os.environ.get("FQDN"))
+    ALLOWED_HOSTS.append(get_ec2_ip())
```
chris48s, line 73: We can't rely on "`DC_ENVIRONMENT` is set" as a proxy for "on EC2" in this application because it is also set when we run on lambda, and `169.254.169.254` is not contactable from lambda (also `ALLOWED_HOSTS` doesn't matter for scheduled jobs) so I've taken a slightly different approach here.

sam-template.yaml
chris48s: So on this one, we want to get to a stage where we are just basing off the generic DC base image, but there are a lot of things in the EE base we need to replicate in order to do it. I think we do want to do this, but I'm going to punt this for another PR.

chris48s: https://app.asana.com/0/1205090401342417/1208112366254722/f

chris48s:
OK. This one is nearly done.

The 2 final changes I am going to make will be:

- Cut a new `dc_logging`, upgrade that package and revert adding `typing-extensions==4.12.2`
- Revert https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2040/commits/cbeaf8a92fc07576eefc4105c5058bd5aedd8f54

and then this one is good to review.

requirements/base.txt
```diff
@@ -19,6 +19,9 @@ git+https://github.com/DemocracyClub/design-system.git@0.5.0
 git+https://github.com/DemocracyClub/dc_django_utils.git@6.1.0
 https://github.com/DemocracyClub/dc_logging/archive/refs/tags/1.0.1.tar.gz
 
+# TODO https://github.com/DemocracyClub/dc_logging/pull/656
+typing-extensions==4.12.2
+
```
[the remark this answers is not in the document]
chris48s: https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2040/commits/daafe060c4f2ceddd39b4ef80451c71fa4de7c3e

chris48s:
OK. I'm calling it. This one is good to review.
I have tested this on https://dev.wcivf.club/ but I've now backed out the commit that changes the deploy branch because we don't want to merge it.

sam-template.yaml
symroe: When you say "a lot of things"...do you have a list, or just a general sense of dread about it?

symroe (approved):

sam-template.yaml
chris48s: You mean more than the usual background level of dread? :smile: 

I basically got as far as deciding there were a non-zero number of things, looked at

- https://github.com/DemocracyClub/EveryElection/blob/master/cdk_stacks/components/add_user.yml
- https://github.com/DemocracyClub/EveryElection/blob/master/cdk_stacks/components/install_app.yml
- https://github.com/DemocracyClub/EveryElection/blob/master/cdk_stacks/components/instance_connect.yml
- https://github.com/DemocracyClub/EveryElection/blob/master/cdk_stacks/components/manage_py_command.yml 

and decided it is an issue for another day. Maybe it will be easier than we think when we come to deal with it.

INSTALL.md
```diff
@@ -132,15 +120,11 @@ Set the password to, for example, `wcivf`. Then create the database, owned by th
...
-Finally, add the PostGIS extension to the database:
-
-    sudo -u postgres psql -d wcivf -c "CREATE EXTENSION postgis;"
-
 Then, create a file `wcivf/settings/local.py` with the following contents, assuming you used the same username, password and database name as above:
 
     DATABASES = {
         'default': {
-            'ENGINE': 'django.contrib.gis.db.backends.postgis',
+            'ENGINE': 'django.db.backends.postgresql_psycopg2',
```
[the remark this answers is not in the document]
chris48s: Yes I think postgresql_psycopg2 is now deprecated or will be in some future django release. We seem to have a bit of a mix on this project. I've updated them all to `postgresql` in https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2040/commits/65ea70f4f5fc785d472bc414557f0b156406be30

PR #2041, opened 2024-08-15
add additional CI checks

chris48s:
~TODO: need to update the docs, but its the same section I've modified in https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2040/ so will conflict.~ docs updates done. Ready for review


.circleci/config.yml
```diff
@@ -79,11 +79,17 @@ jobs:
...
+          name: Pip Check
           command: |
             . .venv/bin/activate
-            curlylint wcivf/**/*.html
+            pip check
+      - run:
+          name: Djhtml
+          command: |
+            . .venv/bin/activate
+            git ls-files '*.html' | xargs djhtml --check
```
chris48s, line 92: Exact same comment as https://github.com/DemocracyClub/EveryElection/pull/2210#discussion_r1718178871 for the same reasons

VirginiaDooley (approved):
Do you think `pip check` should be added to the pre-commit config?

chris48s:
I'm slightly trying to avoid opening the pre-commit worm can at the moment :sweat_smile: but I've added a pre-commit hook for you in 5f4b444 that will run `pip check` on commit if any files under `requirements/*.txt` are modified in the commit.

PR #2046, opened 2024-09-23
Substitute akismet package and reconfigure spam filtering

VirginiaDooley:
In this PR, I've swapped the `python-akismet` spam filtering package for `akismet` with the (so far unproved) theory that a package that is maintained more regularly/recently would help filter out increasingly sophisticated spam submissions. 

---
  - https://app.asana.com/0/0/1208159469948700

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -26,4 +26,4 @@ sentry-sdk==1.27.1
 uk-election-timetables==3.0.0
 uk-election-ids==0.8.0
 django-dotenv==1.4.2
-python-akismet==0.4.3
+akismet==24.5.1
--- a/wcivf/apps/feedback/views.py
+++ b/wcivf/apps/feedback/views.py
@@ -1,4 +1,4 @@
-from akismet import Akismet
+import akismet
 from django.conf import settings
 from django.contrib import messages
 from django.http import HttpResponse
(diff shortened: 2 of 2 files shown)


wcivf/settings/base.py
```diff
@@ -279,9 +279,9 @@ def get_ec2_ip():
 # DC Base Theme settings
 SITE_TITLE = "Who Can I Vote For?"
 
-AKISMET_API_KEY = os.environ.get("AKISMET_API_KEY")
+PYTHON_AKISMET_API_KEY = os.environ.get("PYTHON_AKISMET_API_KEY")
```
symroe, line 282: If you want to change this environment viable, you'll need to change it in the parameter store. 

However, I don't really see the need to change the var name here. The content is the same, and for me `AKISMET_API_KEY` is better than `PYTHON_AKISMET_API_KEY`...in the context of a python project the prefix feels redundant to me.

VirginiaDooley: I changed this on the advice on the package installation instructions https://akismet.readthedocs.io/en/latest/sync_client.html but happy now that I'm setting the config variable, I can just revert and do: 

`config = akismet.Config(
            key=settings.PYTHON_AKISMET_API_KEY,
            url=settings.PYTHON_AKISMET_BLOG_URL,
        )`

wcivf/settings/base.py
```diff
@@ -281,7 +281,7 @@ def get_ec2_ip():
 
 AKISMET_API_KEY = os.environ.get("AKISMET_API_KEY")
 
-AKISMET_BLOG_URL = CANONICAL_URL
+PYTHON_AKISMET_BLOG_URL = CANONICAL_URL
```
symroe, line 284: You're still using `AKISMET_BLOG_URL`, so can you revert this line too?

VirginiaDooley: That's done now, thanks for catching.

symroe (approved):

PR #2068, opened 2024-10-25
fix incorrect var name

chris48s:
Just a little "leave things a bit better than we found them" PR.
I got really confused reading this code.
This variable holds a response, not a request


symroe (approved):

symroe:
You will find that I write this everywhere. I should change it, but my default is to do `req = requests.get()`

chris48s:
ohno

PR #2069, opened 2024-10-25
fix a couple of bits of invalid html

chris48s:
another "leave things a bit better than we found them" PR

This is far from all the invalid HTML, but these 2 were at quick and easy to fix

### DIFF
--- a/wcivf/apps/elections/templates/elections/includes/_single_ballot.html
+++ b/wcivf/apps/elections/templates/elections/includes/_single_ballot.html
@@ -56,7 +56,7 @@
                                 </span>
                             </h2>
                         </summary>
-                        <p>{% blocktrans with description=postelection.election.description|markdown %} {{ description }}{% endblocktrans%}</p>
+                        {% blocktrans with description=postelection.election.description|markdown %} {{ description }}{% endblocktrans%}
                     </details>
                 </div>
             {% endif %}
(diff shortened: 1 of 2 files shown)


wcivf/apps/elections/templates/elections/includes/_single_ballot.html
```diff
@@ -56,7 +56,7 @@ <h2>
                                 </span>
                             </h2>
                         </summary>
-                        <p>{% blocktrans with description=postelection.election.description|markdown %} {{ description }}{% endblocktrans%}</p>
+                        <div>{% blocktrans with description=postelection.election.description|markdown %} {{ description }}{% endblocktrans%}</div>
```
chris48s, line 59: nested <p> is invalid and `|markdown` will generate HTML containing `<p></p>`s

symroe: Sorry, what do you mean by "nested p is invalid"? This is a `p` in a `details`, what's wrong with that?

If the `|markdown` is going to make a `p`, can we just remove the wrapping `div`?

chris48s: There's nothing wrong with a `<p>` in a `<details>`

`<p>` inside another `<p>` is invalid.

`|markdown` will generate at least one `<p>`. I've just removed the container element (`<div>`) completely.

symroe: Got it, sorry I got confused by that :+1:

symroe (approved):

PR #2071, opened 2024-10-28
Show different registration info for City of London Council ballots

chris48s:
Refs https://app.asana.com/0/1204880927741389/1208540419887296/f

City of London has different registration rules to everywhere else in the country.
The point of this PR is to show people the correct "register to vote" information.
Before we started, "register to vote" was what I'm going to call "global" information: It is shown at the page level, not the ballot level.

The tradeoff I have decided to make here is to move it onto the ballot level for City of London Council elections.

In most cases (user is not in City Of London), we'll continue to just show one "global" gov.uk/register-to-vote card

If you're inside the City of London:
- If you have one City of London council election upcoming and nothing else, you'll see registration information on the ballot, but no global card
- If you have two City of London council elections upcoming, you'll see the same registration information duplicated on both ballots. This is sub-optimal, but sufficiently rare I think we can deal with it. You'd have to either have Common Councilman and Alderman elections both upcoming and before the registration deadline, or you'd have to have a by-election and scheduled election for the same ward both upcoming.
- If you have a City of London council election upcoming and also another ballot (e.g: Parliamentary or GLA election) you'll see City of London council registration information on the ballot and also the global gov.uk/register-to-vote card further down

I think this is probably a reasonable approach.

Out of scope for this PR is fixing the polling station opening times for City of London. Lets make that a second follow-up PR.

Screenshot (City of London):

Screenshot 2024-10-28 at 16-05-14 City of London local election candidates in EC2Y 8BT

Screenshot (Everywhere else):

Screenshot 2024-10-28 at 16-05-19 Somerset local election candidates in BA8 0DS

I wasn't able to easily make an example with multiple City of London Council elections or a mix of City + GLA/Parl

### DIFF
--- a/locale/cy/LC_MESSAGES/django.po
+++ b/locale/cy/LC_MESSAGES/django.po
@@ -13,7 +13,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-07-06 12:21+0100\n"
+"POT-Creation-Date: 2024-10-28 14:52+0000\n"
 "PO-Revision-Date: 2022-03-24 09:04+0000\n"
 "Last-Translator: Sym Roe <sym@talusdesign.co.uk>, 2022\n"
 "Language-Team: Welsh (https://www.transifex.com/democracy-club/teams/61326/"
@@ -29,19 +29,19 @@ msgstr ""
 msgid "Enter your postcode"
 msgstr "Rhowch eich côd post"
 
-#: wcivf/apps/elections/models.py:756
+#: wcivf/apps/elections/models.py:765
 msgid "First-past-the-post"
 msgstr "Y Cyntaf i’r Felin"
 
-#: wcivf/apps/elections/models.py:758
+#: wcivf/apps/elections/models.py:767
 msgid "Additional Member System"
 msgstr "Y System Aelodau Ychwanegol"
 
-#: wcivf/apps/elections/models.py:760
+#: wcivf/apps/elections/models.py:769
 msgid "Supplementary Vote"
 msgstr "Pleidlais Atodol"
 
-#: wcivf/apps/elections/models.py:762
+#: wcivf/apps/elections/models.py:771
 msgid "Single Transferable Vote"
 msgstr "Pleidlais Sengl Drosglwyddadwy"
 
@@ -295,6 +295,39 @@ msgstr "Bydd yn awr yn digwydd ar"
 msgid "Read more"
 msgstr "Darllenwch ragor"
 
+#: wcivf/apps/elections/templates/elections/includes/_city_of_london_registration_details.html:8
+#: wcivf/apps/elections/templates/elections/includes/_registration_details.html:10
+#: wcivf/apps/elections/templates/elections/includes/_single_ballot.html:139
+#: wcivf/apps/elections/templates/elections/includes/inline_elections_nav_list.html:90
+msgid "Register to vote"
+msgstr "Cofrestru i bleidleisio"
+
+#: wcivf/apps/elections/templates/elections/includes/_city_of_london_registration_details.html:12
+msgid ""
+"City of London council elections do not use the same electoral register as "
+"other elections."
+msgstr ""
+
+#: wcivf/apps/elections/templates/elections/includes/_city_of_london_registration_details.html:17
+msgid ""
+"Both residents and city workers are eligible to vote. There is one register "
+"published annually, and the deadline to apply is 30 November each year. The "
+"new register comes into force on 16 February each year, and cannot be "
+"modified after that date. For more information, visit the <a href=\"https://"
+"www.speakforthecity.com/\">Speak for the City</a> website or contact City of "
+"London electoral services."
+msgstr ""
+
+#: wcivf/apps/elections/templates/elections/includes/_city_of_london_registration_details.html:28
+#: wcivf/apps/elections/templates/elections/includes/_registration_details.html:30
+msgid ""
+"For questions about your poll card, polling place, or about returning your "
+"postal voting ballot, contact your council."
+msgstr ""
+"Ar gyfer cwestiynau ynghylch eich cerdyn pleidleisio, eich gorsaf "
+"bleidleisio, neu ynghylch dychwelyd eich pleidlais drwy'r post, cysylltwch "
+"â'ch cyngor."
+
 #: wcivf/apps/elections/templates/elections/includes/_elections_breadcrumbs.html:3
 #: wcivf/apps/parties/templates/parties/parties_view.html:9
 msgid "You are here"
@@ -428,7 +461,7 @@ msgid "Photo of %(person_name)s"
 msgstr "Llun o %(person_name)s"
 
 #: wcivf/apps/elections/templates/elections/includes/_polling_place.html:9
-#: wcivf/apps/elections/templates/elections/includes/_single_ballot.html:124
+#: wcivf/apps/elections/templates/elections/includes/_single_ballot.html:131
 #: wcivf/apps/elections/templates/elections/includes/inline_elections_nav_list.html:82
 msgid "Where to vote"
 msgstr "Ble i bleidleisio"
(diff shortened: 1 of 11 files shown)


wcivf/apps/elections/tests/test_postcode_views.py
```diff
@@ -254,8 +254,6 @@ def test_multiple_elections_london(self, mock_response, client):
             election__election_date="2024-05-06",
             election__name="Cities of London and Westminster by-election",
         )
-        parl_london.is_before_registration_deadline = True
-        local_london.is_before_registration_deadline = True
```
chris48s, line 258: None of these properties I removed were doing anything. This is not a property of a ballot.

wcivf/apps/elections/views/mixins.py
```diff
@@ -177,11 +177,21 @@ def get_advance_voting_station_info(self, polling_station: Optional[dict]):
         advance_voting_station["open_in_future"] = open_in_future
         return advance_voting_station
 
-    def is_before_registration_deadline(self, post_elections):
-        if not post_elections:
+    def get_global_registration_card(self, post_elections):
+        # City of London local elections have different
+        # registration rules to every other election
+        non_city_of_london_ballots = [
```
chris48s, line 183: Today on the naming things is hard channel...

wcivf/apps/elections/templates/elections/includes/_single_ballot.html
```diff
@@ -47,19 +47,26 @@ <h3>{% if postelection.is_london_assembly_additional %}{% trans "Additional memb
             {% if object.election.slug == "europarl.2019-05-23"%}
                 {% include "elections/includes/eu_results.html" with card=0 %}
             {% endif %}
-            {% if postelection.election.description %}
-                <div class="ds-details">
```
chris48s, line 51: Is this correct usage of `ds-details` if you've only got one, or should it always be a `<ul>`?
https://democracyclub.github.io/design-system/components/details/

I've made it a `<ul>` here as now we might have 2, but I reckon there are other places in WCIVF where we are doing this wrong if it should always be a `<ul>`

wcivf/apps/elections/tests/test_postcode_views.py
```diff
@@ -618,27 +612,31 @@ def test_show_polling_card(self, view_obj, post_elections):
...
-            view_obj.is_before_registration_deadline(
-                post_elections=post_elections
-            )
-            is True
+        card = view_obj.get_global_registration_card(
+            post_elections=post_elections
         )
+        assert card["show"] is True
+        assert card["registration_deadline"] == "20 April 2020"
+        assert card["election_date"] == "2020-05-06"
```
chris48s, line 639: 🙀 Stop pulling the thread, Chris. Walk away from the pull request.

wcivf/apps/elections/views/mixins.py
```diff
@@ -177,11 +177,21 @@ def get_advance_voting_station_info(self, polling_station: Optional[dict]):
...
+            ballot
+            for ballot in post_elections
+            if not ballot.election.is_city_of_london_local_election
+            and not ballot.cancelled
```
chris48s, line 187: More info in https://github.com/DemocracyClub/WhoCanIVoteFor/commit/cb3e726bab9b9198e39b4a3189cb890aec4029ed
I suspect some variation of this oversight exists in all sorts of other places in this codebase
I've fixed it here, but I haven't gone digging

wcivf/apps/elections/templates/elections/includes/_city_of_london_registration_details.html
```diff
@@ -0,0 +1,40 @@
...
+                {% blocktrans trimmed %}Register to vote{% endblocktrans %}
+            </h2>
+        </summary>
+        <p>
+            {% blocktrans trimmed %}
+                City of London council elections do not use the same electoral register as other elections.
+            {% endblocktrans %}
+        </p>
+        <p>
+            {% blocktrans trimmed %}
```
chris48s, line 17: I wrapped all this text up for translation, but I don't really think there is a big rush to get it translated given this is text we would only ever show to users in England

locale/cy/LC_MESSAGES/django.po
```diff
@@ -13,7 +13,7 @@ msgid ""
 msgstr ""
```
chris48s, line 13: I ran makemessages and a bunch of other stuff unrelated to this PR happened
I guess changes have been made to the templates and merged without updating the translations

Slight tangent: I think we could do with a CI check to fail the build if you have changed text but not updated the PO files. I'm not going to shave this yak right this second in this PR though.

wcivf/apps/elections/templates/elections/includes/_city_of_london_registration_details.html
```diff
@@ -0,0 +1,40 @@
...
+                Both residents and city workers are eligible to vote. There is one register
+                published annually, and the deadline to apply is 30 November each year.
+                The new register comes into force on 16 February each year,
+                and cannot be modified after that date. For more information, visit the
+                <a href="https://www.speakforthecity.com/">Speak for the City</a>
+                website or contact City of London electoral services.
```
chris48s, line 23: This was Peter's suggested text, but we can tweak it if you want now we have the logic in place to show/hide text in the right situations.

I think the rules are sufficiently simple here that we might as well just hard-code the dates like this.

wcivf/apps/elections/models.py
```diff
@@ -457,30 +457,39 @@ def expected_sopn_date(self):
...
     def registration_deadline(self):
-        date = get_election_timetable(
-            self.ballot_paper_id, self.post.territory
-        ).registration_deadline
+        try:
+            date = get_election_timetable(
+                self.ballot_paper_id, self.post.territory
+            ).registration_deadline
+        except AttributeError:
+            return None
```
chris48s, line 470: There's more info on this in the commit message for https://github.com/DemocracyClub/WhoCanIVoteFor/commit/ec92c400dd1c54169cdd294727a09fd6b0bdd330

I don't really like what I've done here. Basically the motivation is that once I started calling `ballot.registration_deadline` in `get_global_registration_card()` tests started failing all over the place because we don't handle the case where `get_election_timetable()` returns `None` and we don't really set up "realistic" test data. I actually suspect this condition doesn't really exist with "real" data. I don't think we're handling this right at all, but I decided to just stop pulling the thread at this layer of abstraction and hit everything with the same hammer we hit `expected_sopn_date` with.

wcivf/apps/elections/templates/elections/includes/_city_of_london_registration_details.html
```diff
@@ -0,0 +1,40 @@
...
+                {% blocktrans trimmed %}
+                    For questions about your poll card, polling place, or about returning your postal voting ballot, contact your council.
+                {% endblocktrans %}
+            </p>
+            {% if registration %}
+                {% include "elections/includes/_council_contact_details.html" with contact_details=registration %}
+            {% endif %}
+            {% if council %}
+                {% include "elections/includes/_council_contact_details.html" with contact_details=council %}
+            {% endif %}
```
chris48s, line 37: This is kinda wrong, but for now lets put a pin in it, accept it is no more wrong than every other place we are presenting council contact information, and get into this one in detail in https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2072

wcivf/apps/elections/models.py
symroe: Yeah, I'd like to see some decent documented exceptions from the timetables library, but more generally we should be using better test data in this project so we don't need to do this. Agree that's for another day. Can you write an issue?

wcivf/apps/elections/templates/elections/includes/_single_ballot.html
symroe: No need to use a `ul` when there is only one item...I think a list of one item isn't really a list? Ideally we'd be clever about wrapping it in a list if there were more than one, but I don't think this matters a lot. 

I think it might be easier to read if you put the wrapping `li` in the parent template, and make the include just contain the `details` element. This way we know that the include is 'self contained' and doesn't need to be used inside a `ul`

symroe (commented):
I think this is a good incremental solution. I'd really like to find a better pattern for controlling 'global' content than this somewhat ad-hoc approach, but that's a bigger issue for another day :+1:

wcivf/apps/elections/templates/elections/includes/_single_ballot.html
chris48s: makes sense. Done in https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2071/commits/15084cf18dd9686af2685bfc3124e7f2fa0a4355

wcivf/apps/elections/models.py
chris48s: The package itself actually throws some fairly specific exceptions

https://github.com/DemocracyClub/uk-election-timetables/blob/b631dcea3e3ff7590ddca41cf680c3a72949fa05/uk_election_timetables/election_ids.py#L20-L53

The issue is that in `get_election_timetable()` we just catch them all (and some other errors) and return a null value. I just decided I needed to stop pulling the thread here.

I'll write up a ticket about it :+1:

symroe (approved):

PR #2086, opened 2024-11-12
Grep for errors in migrating

symroe:
As per https://github.com/DemocracyClub/Website/pull/889

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,8 @@ lambda-migrate:  ## Invoke lambda to migrate the database
 	aws lambda invoke \
 	--function-name WCIVFControllerFunction \
 	--payload '{ "command": "migrate" }' \
-	/dev/stdout
+	--cli-binary-format raw-in-base64-out \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -28,7 +28,8 @@ lambda-migrate:  ## Invoke lambda to migrate the database
 	aws lambda invoke \
 	--function-name WCIVFControllerFunction \
 	--payload '{ "command": "migrate" }' \
-	/dev/stdout
+	--cli-binary-format raw-in-base64-out \
+	/dev/stdout | grep -v 'FunctionError'
```
chris48s, line 32: In general, I am in favour of using long flags in scripts as per https://changelog.com/posts/use-long-flags-when-scripting This goes double for tools that have unusual short aliases. I'd normally assume `-v` is `--verbose`. I'd find it easier to read this in future if we write this with `--invert-match` instead of `-v`.

symroe: Pushed an amended commit :+1:

chris48s (approved):

PR #2087, opened 2024-11-12
silence sentry DisallowedHost errors

chris48s:
Refs https://app.asana.com/0/1205090401342417/1208407136740662/f

Similar to https://github.com/DemocracyClub/EveryElection/pull/2311 and https://github.com/DemocracyClub/UK-Polling-Stations/pull/8152


symroe:
Couple of things:

1. The file names in `settings` are really about _environments_ rather than collections of settings (`base.py` is an exception, of course). I wonder if there's a better place for this, or a sub-folder we can make that groups settings (I can see this would be useful for, e.g pipeline settings)
2. Using `before_send` is fine, but is it possible to [ignore the logger](https://docs.sentry.io/platforms/python/integrations/logging/#ignoring-a-logger) with `ignore_logger("django.security.DisallowedHost")`?

chris48s:
Yeah I tried doing it with `ignore_logger` which also works and is way simpler.
I think given that method requires adding less code, it makes sense to just do it inline anyway, so I've got rid of the separate file.
I agree that the config files themselves should correspond to environments. In WDIV, we have split the config into modules but put them in a subdirectory. So the files in https://github.com/DemocracyClub/UK-Polling-Stations/tree/master/polling_stations/settings correspond to environments but the files in https://github.com/DemocracyClub/UK-Polling-Stations/tree/master/polling_stations/settings/constants are groups of related settings. That's probably the way to do it.
But also, if we don't need to define a filtering function this can probably just live in the base config. Locality of behaviour is also a thing to optimise for.

symroe (approved):

symroe:
Nice one :+1:

PR #2107, opened 2024-12-03
Use prod URL for 'layers of state'

symroe:


wcivf/settings/base.py
```diff
@@ -332,7 +332,7 @@ def get_ec2_ip():
     "ENABLE_LAYERS_OF_STATE_FEATURE", False
 )
 LAYERS_OF_STATE_URL = (
-    "https://developers.environments.womblelabs.co.uk/api/v1/layers_of_state/"
+    "https://developers.democracyclub.org.uk/api/v1/layers_of_state/"
```
chris48s, line 335: Shouldn't this use `DEVS_DC_BASE` ?

symroe: Good point, pushed a change to remove this setting and just use the base URL in the app code

chris48s (approved):

PR #2117, opened 2024-12-09
remove get_voter_id_requirements property

chris48s:
Refs https://app.asana.com/0/1204880927741389/1208687275764621/f

These two places in WCIVF where we are showing ID requirements are not doing the same thing
In the frontend, we are looking at the `PostElection.requires_voter_id` property
https://github.com/DemocracyClub/WhoCanIVoteFor/blob/afcd3c6a366d1a2a380088858643caa26e151759/wcivf/apps/elections/models.py#L420
This is a database field
In the API endpoint, we are looking at the `PostElection.get_voter_id_requirements` property
https://github.com/DemocracyClub/WhoCanIVoteFor/blob/afcd3c6a366d1a2a380088858643caa26e151759/wcivf/apps/elections/models.py#L715-L723
This function basically answers the same question, but it works it out from first principles using the election type and territory code.

In this PR I have removed `get_voter_id_requirements`. Everywhere now looks at `requires_voter_id`.

### DIFF
--- a/wcivf/apps/api/views.py
+++ b/wcivf/apps/api/views.py
@@ -96,7 +96,7 @@ class BaseCandidatesAndElectionsViewSet(
                 "voting_system": VotingSystemSerializer(
                     postelection.voting_system
                 ).data,
-                "requires_voter_id": postelection.get_voter_id_requirements,
+                "requires_voter_id": postelection.requires_voter_id,
                 "postal_voting_requirements": postelection.get_postal_voting_requirements,
                 "seats_contested": postelection.winner_count,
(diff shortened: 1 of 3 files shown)


symroe:
This looks good. It would be nice at some point to copy this pattern for `postal_vote_requires_form` and drop `uk_election_ids` from this project

symroe (approved):

chris48s:
Yes. I agree there are a bunch of functions in WCIVF which are using either uk_election_ids or uk_election_timetables that should really be getting their data from EE instead of working it out from first principles.

PR #2118, opened 2024-12-09
show correct opening times for City of London local elections

chris48s:
Refs https://app.asana.com/0/1204880927741389/1208687275764619/f

Initially, I was just going to implement the same thing we've done for the widget and EC site. However in this codebase, it looks like we've kind of tried to solve this at some point but not really, or possibly fudged something for the Parl/City of London on same day you had to deal with this year and then reverted some of it but not all of it. Dunno.

Anyway, I've attempted to fir this into the existing constructs and this should sort it.

### DIFF
--- a/locale/cy/LC_MESSAGES/django.po
+++ b/locale/cy/LC_MESSAGES/django.po
@@ -13,7 +13,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-11-06 16:32+0000\n"
+"POT-Creation-Date: 2024-12-09 15:05+0000\n"
 "PO-Revision-Date: 2022-03-24 09:04+0000\n"
 "Last-Translator: Sym Roe <sym@talusdesign.co.uk>, 2022\n"
 "Language-Team: Welsh (https://www.transifex.com/democracy-club/teams/61326/"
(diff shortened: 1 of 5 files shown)


chris48s:
I'm pretty sure this `AmbiguousElectionIdError` error sentry is flagging here:

a) is unrelated to this PR
b) will be fixed when I deploy https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2116 and run the update tomorrow as I think it is fundamentally caused by missing territory codes

locale/cy/LC_MESSAGES/django.po
```diff
@@ -520,18 +520,19 @@ msgid "<address>%(polling_station_address)s</address>"
 msgstr ""
 
 #: wcivf/apps/elections/templates/elections/includes/_polling_place.html:57
-msgid "It will be open from <strong>7am to 10pm</strong>."
-msgstr "Bydd ar agor o <strong>7am tan 10pm</strong>."
+#, python-format
+msgid "It will be open from <strong>%(open_time)s to %(close_time)s</strong>"
+msgstr "Bydd ar agor o <strong>%(open_time)s tan %(close_time)s</strong>"
```
chris48s, line 525: I know we've agreed we're going to ditch translations on WDIV and WCIVF
https://app.asana.com/0/1205090401342417/1208714863410843/f
but it was relatively easy to just do this right
so I did.

symroe (approved):

PR #2158, opened 2025-01-23
refactor set_territory

awdem:
territory_code is now a required field on divisions in EE so we don't need to account for it being empty with '-'. We still have elections with no divisions, though, so we still want to fall back to the org's territory_code in that case. And terrority_code on orgs is not a required field yet, so we still need to account for it being blank.

Tested locally with:

`mayor.cambridgeshire-and-peterborough.2025-05-01`
`local.gloucestershire.grange-kingsway.2025-05-01`


wcivf/apps/elections/import_helpers.py
```diff
@@ -466,13 +466,11 @@ def set_territory(self, ballot):
...
-                "territory_code",
-                ee_data["organisation"].get("territory_code", "-"),
-            )
-        else:
-            territory = "-"
+        # If an election has no divisions, try the org territory code
+        if ee_data["division"] is None:
+            territory = ee_data["organisation"].get("territory_code", "-")
+        # Otherwise use the division territory code
+        territory = ee_data["division"].get("territory_code")
```
symroe, line 473: I think this needs to be in an 'else'? `ee_data["division"]` always exists, so thhe `.get()` will fail when it's `None`

awdem: You're right! Have fixed in 929ea133143424b72b86381bdcae4ae26b8fa73b

symroe (approved):

PR #2169, opened 2025-02-01
Move to uv

symroe:
This should be fairly uncomplicated. 

I've not (yet) tested a deploy, so I want to do that before we merge this.


pyproject.toml
```diff
@@ -28,3 +63,29 @@ addopts =[
...
+    "pytest-mock>=3.14.0",
+    "pytest-subtests>=0.13.1",
+    "tomlkit>=0.13.2",
+    "freezegun==1.2.2",
+    "django-debug-toolbar==4.3.0",
+    "djhtml==3.0.6",
+    "ruff==0.1.15",
+    "pytest-ruff==0.2.1",
+    "pytidylib>=0.3.2",
+]
```
chris48s, line 91: This is an application, so can we also set

```
[tool.uv]
package = false
```

somewhere in here. This tells UV to stop us from accidentally pushing it to PyPI.

pyproject.toml
```diff
@@ -28,3 +63,29 @@ addopts =[
     "--ruff",
     "--ruff-format",
 ]
+
+[dependency-groups]
+deploy = [
```
chris48s, line 68: This is a bit of a nitpick, but our pyproject.toml has 2 categories of thing in it:

- Package management stuff (`project`, `dependency-groups`, `tool.uv.<things>`)
- Config for unrelated tools that happen to live in pyproject.toml (`tool.ruff`, `tool.pytest`)

Could we group all the package management stuff together in the file seperately from the unrelated config just for readability. At the moment it is a bit of a random mashup.

requirements.txt_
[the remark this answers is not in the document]
chris48s: Also base.txt is still knocking about and needs deleting

chris48s (commented):
Right. I am going to leave it with you to work out the deploy to dev, but I have had a read over the diff and left you a first round of comments

chris48s:
As noted in https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2179#issuecomment-2653249695
There is now an extra package we need to add in here

pyproject.toml
```diff
@@ -1,3 +1,38 @@
...
+    "markdown==3.5.1",
+    "django-localflavor==4.0",
+    "django-extensions==3.2.3",
+    "djangorestframework==3.14.0",
+    "icalendar==5.0.7",
+    "dc-django-utils",
+    "dc-logging-utils",
+    "djangorestframework-jsonp==1.0.2",
+    "feedparser==6.0.10",
+    "uk-election-timetables==4.1.0",
```
chris48s, line 22: Also need to bump this in line with https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2183

symroe:
@chris48s I've pushed a single commit addressing all of these comments. I suggest I squash them all down into one before merging.

.circleci/config.yml
```diff
@@ -66,62 +62,64 @@ jobs:
...
-            ruff format . --check
-            ruff check .
+            uv run ruff format . --check
+            uv run ruff check .
       - run:
-          name: Pip Check
-          command: |
-            . .venv/bin/activate
-            pip check
+          name: Check lockfile us up to date
```
chris48s, line 77:
```suggestion
          name: Check lockfile is up to date
```

chris48s:
Had a quick look over the latest changes. One minor typo. Aside from that, looks good but we still have the build/deploy failure to work out

symroe:
Ok, I've managed a dev deploy with this, so I think we're good. I picked out the commit that deployed there, but I can add back in if you'd like to see it in action

deploy/files/systemd/gunicorn.service
```diff
@@ -5,7 +5,7 @@ After=network.target
 User=wcivf
 Group=wcivf
 WorkingDirectory=/var/www/wcivf/code/
-ExecStart=/bin/bash -c 'PATH=/var/www/wcivf/venv/bin/:$PATH exec /var/www/wcivf/venv/bin/gunicorn \
+ExecStart=/bin/bash -c 'PATH=/var/www/wcivfcode/.venv/bin/:$PATH exec /var/www/wcivf/code/.venv/bin/gunicorn \
```
chris48s, line 8: Hmm. This looks like it should be

```suggestion
ExecStart=/bin/bash -c 'PATH=/var/www/wcivf/code/.venv/bin/:$PATH exec /var/www/wcivf/code/.venv/bin/gunicorn \
```

symroe: Ok, deployed and tested again. I'll leave that deploy up in case you want to take a look

chris48s (approved):
will leave you to delete the "deploy to dev commit", squash, merge, etc

PR #2173, opened 2025-02-06
render whitespace in cancelled election details

chris48s:
Refs https://app.asana.com/0/1209332771640377/1209290265508276/f
Refs https://app.asana.com/0/1209332771640377/1209332771640386/f

I've already edited the descriptions in EE to already include line breaks in the right places.

Just a small change here as the handling of cancelled elections is already pretty good.

### DIFF
--- a/wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
+++ b/wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
@@ -47,7 +47,7 @@
 {% else %}
     {% if object.metadata.cancelled_election %}
         <h4>{{ object.metadata.cancelled_election.title }}</h4>
-        <p>{{ object.metadata.cancelled_election.detail }}</p>
+        {{ object.metadata.cancelled_election.detail | linebreaks }}
     {% else %}
         <h4>{% trans "Cancelled Election" %}</h4>
         <p>{% trans "This election was cancelled." %}</p>


wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
```diff
@@ -47,7 +47,7 @@ <h4>{% trans "Cancelled and Rescheduled Election" %}</h4>
 {% else %}
     {% if object.metadata.cancelled_election %}
         <h4>{{ object.metadata.cancelled_election.title }}</h4>
-        <p>{{ object.metadata.cancelled_election.detail }}</p>
+        {{ object.metadata.cancelled_election.detail | linebreaks }}
```
chris48s, line 50: Using the `| linebreaks` filter renders the whole thing in a `<p>` for us and interprets `\n\n` as starting a new `<p>` so that's why I've dropped the `<p>` wrapper here
https://docs.djangoproject.com/en/5.1/ref/templates/builtins/#linebreaks

wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
```diff
@@ -47,7 +47,7 @@ <h4>{% trans "Cancelled and Rescheduled Election" %}</h4>
...
     {% else %}
         <h4>{% trans "Cancelled Election" %}</h4>
         <p>{% trans "This election was cancelled." %}</p>
```
chris48s, line 53: There is still this fallback case here where we're saying "cancelled".
This isn't a case I'd expect to hit often as it only happens if we didn't have a cancelled reason or metadata object, but I guess we can end up here and if we do, we don't know if this election is postponed or cancelled.

Similar to WDIV, the best thing I can think of to say here would be
"This election will not go ahead as planned"
but if you've got any other ideas for generic words, I'm open to ideas on this.

symroe: >  "This election will not go ahead as planned"

This is a good suggestion, but I'll let @pmk01 decide on what's best. I don't think saying "cancelled" is _so_ bad that we must avoid it at all costs!

symroe (approved):

wcivf/apps/elections/templates/elections/includes/_cancelled_election.html
pmk01: I am happy with either text! For context the case I am thinking about is this one: https://www.dailyrecord.co.uk/in-your-area/renfrewshire/shambles-key-election-cancelled-under-25547891

chris48s: OK. Lets leave this one as it stands for now

PR #2179, opened 2025-02-11
Read from primary in admin

chris48s:
Refs https://app.asana.com/0/1204880927741389/1209211842699505/f
The scope of this one is a bit narrower than
https://github.com/DemocracyClub/UK-Polling-Stations/pull/8221
We're basically just trying to make it so changing things in the admin works.

There will probably still be race conditions if we try to SSH in and run scripts.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -10,6 +10,7 @@ django-localflavor==4.0
 django-extensions==3.2.3
 djangorestframework==3.14.0
 django-redis==5.3.0
+django-middleware-global-request==0.3.5
 hiredis==2.3.2
 redis==5.0.4
 icalendar==5.0.7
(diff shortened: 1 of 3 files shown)


requirements/base.txt
```diff
@@ -10,6 +10,7 @@ django-localflavor==4.0
 django-extensions==3.2.3
 djangorestframework==3.14.0
 django-redis==5.3.0
+django-middleware-global-request==0.3.5
```
chris48s, line 13: Lets assume we will merge https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2169 ahead of this and I'll go back and change this

chris48s:
Oh. The other thing I was going to look at here was the feedback DB/Router.
As far as I can see, we still do have a separate feedback DB configured in prod so I haven't attempted to get rid of that here. It is another issue for another day.

symroe (approved):

PR #2181, opened 2025-02-12
show expected SOPN date even if there are no candidates

chris48s:
Refs https://app.asana.com/0/1204880927741389/1209167759131436/f


chris48s:
It might be helpful to look at the diff ignoring whitespace for this one
https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2181/files?w=1

The key detail here is that previously we had code to the effect of

```django
{% if not postelection.people and postelection.contested %}
    some content
{% else %}
    {% if not postelection.people and postelection.contested %}
        different content
        but this block is unreachable
    {% endif %}
{% endif %}
```

so I pulled the content from the unreachable block out to the top `{% if not postelection.people and postelection.contested %}` and then the rest is just indenting and closing tags

symroe (approved):

PR #2188, opened 2025-02-18
expose VAC in section on ID

chris48s:
Refs https://app.asana.com/0/1204880927741389/1209387679034176/f

I've punted doing something with the postal vote deadline to https://app.asana.com/0/1204880927741389/1209386545391512/f

### DIFF
--- a/wcivf/apps/elections/models.py
+++ b/wcivf/apps/elections/models.py
@@ -494,6 +494,26 @@ class PostElection(TimeStampedModel):
             return None
 
     @property
+    def past_vac_application_deadline(self):
+        try:
+            vac_application_deadline = get_election_timetable(
+                self.ballot_paper_id, self.post.territory
+            ).vac_application_deadline
+        except AttributeError:
+            return None
+
+        return vac_application_deadline < datetime.date.today()
+
+    @property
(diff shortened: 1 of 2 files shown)


wcivf/apps/elections/models.py
```diff
@@ -493,6 +493,26 @@ def postal_vote_application_deadline(self):
         except AttributeError:
             return None
 
+    @property
```
chris48s, line 496: This code is bobbins, but I wrote it anyway.

We should totally fix this
https://app.asana.com/0/1207876275349083/1208707350429290

but we decided to ignore it for a bit so I'm just following the same pattern for this field, even though I don't like it.

wcivf/apps/elections/templates/elections/includes/_voter_id.html
```diff
@@ -25,6 +25,21 @@ <h2 id="requirements">
                     </li>
                 </ul>
             {% endblocktrans %}
+
+            {% if not postelections.0.past_vac_application_deadline %}
```
chris48s, line 29: Here I am using the property that all elections that use ID (usefully) have the same VAC deadline

symroe (approved):

PR #2201, opened 2025-03-06
2025 sheets imports

awdem:
This PR:
- Adds bluesky and instagram to the LocalParty model
- Updates the headers for the local party importer
- Updates the local parties google sheet endpoints
- Restarts the regular schedule for running import_local_parties

---
  - https://app.asana.com/0/0/1209453389332377

### DIFF
--- a/wcivf/apps/people/templates/people/includes/_person_contact_card.html
+++ b/wcivf/apps/people/templates/people/includes/_person_contact_card.html
@@ -30,9 +30,9 @@
 
             {% if object.twitter_username %}
                 <div>
-                    <dt>{% trans "Twitter" %}</dt>
+                    <dt>{% trans "X" %}</dt>
                     <dd>
-                        <a href="https://twitter.com/{{ object.twitter_username }}" title="{% blocktrans trimmed with person_name=object.name%}{{ person_name }}'s Twitter profile{% endblocktrans%}">
+                        <a href="https://twitter.com/{{ object.twitter_username }}" title="{% blocktrans trimmed with person_name=object.name%}{{ person_name }}'s X profile{% endblocktrans%}">
                             {{ object.twitter_username}}
                         </a>
                     </dd>
--- a/wcivf/apps/people/templates/people/includes/_person_local_party_card.html
+++ b/wcivf/apps/people/templates/people/includes/_person_local_party_card.html
@@ -11,7 +11,7 @@
     <dl class="ds-descriptions">
         {% if person.local_party.twitter %}
             <div>
-                <dt>Twitter</dt>
+                <dt>X</dt>
                 <a href="https://twitter.com/{{ person.local_party.twitter }}">
                     {{ person.local_party.twitter }}
                 </a>


wcivf/apps/parties/importers.py
```diff
@@ -125,11 +125,13 @@ def add_local_party(self, row, party, ballots, file_url):
...
                 "name": name,
                 "twitter": twitter,
+                "instagram": row.get("Instagram", "").strip(),
+                "bluesky": row.get("Bluesky", "").strip(),
                 "facebook_page": row["Facebook"],
                 "homepage": row["Website"],
                 "email": row["Email"],
                 "is_local": country == "Local",
-                "youtube_profile_url": row.get("Youtube profile", "").strip(),
+                "youtube_profile_url": row.get("Party broadcast", "").strip(),
```
chris48s, line 134: Any particular reason why we're treating these fields slightly differently: i.e: `row["Facebook"]` but `row.get("Instagram", "").strip()` ?

If not, I'd be inclined to say:
- Fetch everything with square bracket notation i.e: the importer should fail if we've broken the spreadsheet
- Strip whitespace from all fields

wcivf/apps/parties/models.py
```diff
@@ -310,6 +310,8 @@ class LocalParty(TimeStampedModel):
     )
     name = models.CharField(blank=True, max_length=100)
     twitter = models.CharField(blank=True, max_length=100)
+    bluesky = models.URLField(blank=True, max_length=100)
+    instagram = models.URLField(blank=True, max_length=100)
```
chris48s, line 314: Having added these exciting new fields to the `LocalParty` model, do we need to do anything to show them to the user in the frontend?

wcivf/apps/parties/importers.py
awdem: Hmm there are only two templates that seem to show info about local parties and one is currently has the local party card commented out:

https://github.com/DemocracyClub/WhoCanIVoteFor/blob/0ffa757e92fddd59731a3d5d88062764dca11da2/wcivf/apps/people/templates/people/person_detail.html

https://github.com/DemocracyClub/WhoCanIVoteFor/blob/0ffa757e92fddd59731a3d5d88062764dca11da2/wcivf/apps/elections/templates/elections/party_list_view.html

Might be have to review with Peter and/or Sym about what we actually want to show. On the above list view we show facebook, homepage, and email for both national and local party, We don't show Instagram for the national party even though there is a field on the model for it.

wcivf/apps/parties/models.py
awdem: Hmm there are only two templates that seem to show info about local parties and one is currently has the local party card commented out:

https://github.com/DemocracyClub/WhoCanIVoteFor/blob/0ffa757e92fddd59731a3d5d88062764dca11da2/wcivf/apps/people/templates/people/person_detail.html

https://github.com/DemocracyClub/WhoCanIVoteFor/blob/0ffa757e92fddd59731a3d5d88062764dca11da2/wcivf/apps/elections/templates/elections/party_list_view.html

Might be have to review with Peter and/or Sym about what we actually want to show. On the above list view we show facebook, homepage, and email for both national and local party, We don't show Instagram for the national party even though there is a field on the model for it.

wcivf/apps/parties/importers.py
awdem: The reason is I copied and pasted. I will make it consistent.

awdem: Done in 85fd77e46756e8077138a759f7a07b693a004124

wcivf/apps/parties/models.py
chris48s: OK. Uncommenting that might be part of re-enabling it, like uncommenting the cron job. Lets follow it up in standup on Monday.

chris48s: Aside from that, I think the code here so far is good

awdem: I've added both fields to the templates in 62405ed5d18421b3c00127b8857418869f1d2816. Sym said that national parties are only used for GEs so I think it's safe to uncomment the `local_party_card` on the `person_detail_view`.

chris48s:
One more job:
We don't need to the DB field or variables, but in the frontend templates can we change "twitter" to "X".

awdem:
> One more job: We don't need to the DB field or variables, but in the frontend templates can we change "twitter" to "X".

Done in b41fb798de8f5ae3b5307fd3f3e36c7d3dde42a1

chris48s (approved):

PR #2228, opened 2025-04-02
Refactor/always import current metadata

awdem:
This PR addresses an issue we were having where updating metadata objects on EE wasn't propagating to WCIVF.

 The problem was that we wouldn't update metadata objects if a ballot on WCIVF already had one. However, this PR goes further so that now we always import metadata objects for current ballots. 

---
  - https://app.asana.com/0/0/1209387104787681

### DIFF
--- a/wcivf/apps/elections/import_helpers.py
+++ b/wcivf/apps/elections/import_helpers.py
@@ -188,7 +188,6 @@ class YNRBallotImporter:
         current_only=False,
         exclude_candidacies=False,
         force_metadata=False,
-        force_current_metadata=False,
         recently_updated=False,
         base_url=None,
         api_key=None,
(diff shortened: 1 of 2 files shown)


chris48s:
This should do the job. Tbh, I wonder if we can completely remove all the --force* params as I feel like they are all there to solve a problem we don't really have anymore. I reckon we probably have variations of this bug on all the `YNRBallotImporter.set_*` methods when `force_update` is not true.

Maybe it is another job for another day, but can we at least stick a ticket in inbox for it.

awdem:
> This should do the job. Tbh, I wonder if we can completely remove all the --force* params as I feel like they are all there to solve a problem we don't really have anymore. I reckon we probably have variations of this bug on all the `YNRBallotImporter.set_*` methods when `force_update` is not true.
> 
> Maybe it is another job for another day, but can we at least stick a ticket in inbox for it.

Yeah I did think about removing more, but decided that I didn't want to get into the weeds with it. I'll create a follow-up ticket to review the --force* params, though.

PR #2246, opened 2025-04-17
Fix/indie candidate intros

awdem:
Old:
image
New:
image

Old:
image
New:
image

---
  - https://app.asana.com/0/0/1209107007910837


awdem:
> Ideally we want to say 'an' independent candidate, rather than 'the'. Otherwise these are perfect, thanks!
> 
> (it might be the case that we say 'the' for party candidates, but that's presumably because there will only be one? Perhaps there's some logic here which uses seats contested value)

It's not that smart, sadly: it's just hard-coded. I've changed 'the' to 'an' here: 8346b911cdffbdc1a3bcafc165dc02a0c3fc7bbf

chris48s (approved):

PR #2248, opened 2025-04-18
list current but past elections

### LINKED ISSUE #2029: Post-election header text for candidates in more than one election on the same day
If a candidate stands in more than one election on the same day, we list all the elections at the top of their profile _before_ polling day. However, after polling day we only display one election. The candidate below stood in both parl and council elections on 4 July, but only the latter is displayed

Screenshot_7-8-2024_9322_whocanivotefor co uk

awdem:
Now if a candidate has has no future candidacies, and multiple current but past candidacies, we will list them in their intro card.

Before we'd only show one of the elections, (e.g. the below candidate stood in both parl and local elections but we only show the local one in the intro card).
image

Now we'll show both in a list:
image

closes  #2029 

---
  - https://app.asana.com/0/0/1210003516031209

### DIFF
--- a/wcivf/apps/people/templates/people/includes/_person_multiple_current_candidacies.html
+++ b/wcivf/apps/people/templates/people/includes/_person_multiple_current_candidacies.html
@@ -1,0 +1,20 @@
+{% load i18n static %}
+
+<h5>{{ person_name }}
+    {% with party=candidacies.0.party %}
+        {% blocktrans trimmed with party_name=party.party_name party_link=party.get_absolute_url a_or_an=party.is_independent|yesno:_("an,a") %}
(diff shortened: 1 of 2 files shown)


wcivf/apps/people/templates/people/includes/_person_multiple_current_candidacies.html
```diff
@@ -0,0 +1,20 @@
+{% load i18n static %}
+
+<h5>{{ person.name }}
```
chris48s, line 3: Here you are referencing `person.name` but where you're including the template you are calling it with
```
person_name=object.name
```
Looks like this should probably be `{{ person_name }}`

chris48s (approved):

PR #2261, opened 2025-06-19
Replaced elections on home page

symroe:
A few commits to clean up the upcoming elections block on the home page:

1. Include any ballot that replaces a ballot. This catches polls that have been countermanded from scheduled elections
2. Include referendums
3. Move the code to a manager
4. Feature flag the code for easier disabling around scheduled elections. 

---
  - https://app.asana.com/0/0/1210519443877124


awdem (approved):
This looks good to me!

PR #2262, opened 2025-07-16
bump uk-election-ids=0.10.0

chris48s:
See the changelog notes for uk-election-ids==0.10.0 at
https://github.com/DemocracyClub/uk-election-ids/releases/tag/0.10.0

This application shouldn't require any changes as long as every post has a sensible value for `territory` on it (where sensible means one of `ENG`, `WLS`, `SCT`, `NIR`. Empty string and `GBN` are both non-sensible values).

Unfortunately, we do have some posts with non-sensible values on them.

This means we have to make an assumption here

https://github.com/DemocracyClub/WhoCanIVoteFor/blob/5f96a03575f75816de64db83589b4e43ac9a18ae/wcivf/apps/elections/models.py#L548-L560

I think it probably is right to push the onus to make that assumption on to client applications. If we don't have a good territory code here, what's the pragmatic assumption here?

### DIFF
--- a/wcivf/apps/elections/tests/factories.py
+++ b/wcivf/apps/elections/tests/factories.py
@@ -40,6 +40,7 @@ class PostFactory(factory.django.DjangoModelFactory):
     ynr_id = "WMC:E14000647"
     label = "copeland"
     organization_type = "local-authority"
+    territory = "ENG"
 
 
 class PostElectionFactory(factory.django.DjangoModelFactory):


symroe:
Can you quantify "some posts with non-sensible values"? I'd really like to be in a position where we don't have bad data in the system at all, so is it worth going back and fixing that rather than trying to wallpaper over the cracks?

chris48s:
I think we should just get https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2264 (and less crucially https://github.com/DemocracyClub/yournextrepresentative/pull/2547 ) in ahead of this one.

PR #2263, opened 2025-07-16
update is_constituency and is_regional for Senedd changes

chris48s:
- Update `is_constituency` and `is_regional` to correctly classify new Senedd IDs
- Fix bug in `is_regional`: `gla.r` does not exist
- More extensive tests for these functions

### DIFF
--- a/wcivf/apps/elections/models.py
+++ b/wcivf/apps/elections/models.py
@@ -591,13 +591,11 @@ class PostElection(TimeStampedModel):
 
     @property
     def is_constituency(self):
-        if self.ballot_paper_id.startswith("senedd.r."):
-            return False
-        if self.ballot_paper_id.startswith(("gla.c.", "senedd.c.", "sp.c.")):
+        if self.ballot_paper_id.startswith("senedd."):
+            if self.ballot_paper_id.startswith("senedd.r."):
+                return False
             return True
-        if re.match(
-            r"^senedd\.[a-z\-]+.\d{4}-\d{2}-\d{2}$", self.ballot_paper_id
-        ):
+        if self.ballot_paper_id.startswith(("gla.c.", "senedd.c.", "sp.c.")):
             return True
         return False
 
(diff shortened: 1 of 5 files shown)


wcivf/apps/elections/models.py
```diff
@@ -591,11 +591,19 @@ def is_pcc(self):
...
     @property
     def is_constituency(self):
-        return self.ballot_paper_id.startswith(("gla.c", "senedd.c", "sp.c"))
+        if self.ballot_paper_id.startswith("senedd.r."):
+            return False
+        if self.ballot_paper_id.startswith(("gla.c.", "senedd.c.", "sp.c.")):
+            return True
+        if re.match(
+            r"^senedd\.[a-z\-]+.\d{4}-\d{2}-\d{2}$", self.ballot_paper_id
+        ):
```
symroe, line 600: I'm not so keen on using a regex here (or anywhere!). Can we just make this simpler with:

```py
if self.ballot_paper_id.startswith("senedd"):
        if self.ballot_paper_id.startswith("senedd.r."):
            return False
        return True
```

chris48s: Yep, so the thing I was trying to avoid by writing a regex is `is_constituency()` returning true for `senedd.2025-07-17` (which would be wrong) although your suggestion is fine as long as `ballot_paper_id` is definitely a _ballot_ id (as opposed to any _election_ ID).

Having pushed af9b6c2, that has revealed that there are tests that need rewriting because they're setting `ballot_paper_id` to group IDs.

chris48s: As I'm getting into fiddling with the tests, I am also realising that the new Senedd elections have a unique property, which is we want to describe the divisions as a "constituency" - that's the correct terminology. But also they use party lists so I am slightly ambiguous whether we want to use `_constituency.html` or `_list_ballot.html` or a completely new template.

Also this is hard to test because we have zero examples of this type of ballot yet.

Bah. I was hoping this would be a nice quick and easy issue :smile:

symroe: I think given the way the models work in this project, something would have gone very wrong to get `senedd.date` as a ballot. You could always `assert self.ballot_paper_id.count(".") > 1` if you wanted to be sure, but I think that's overkill here.

PR #2277, opened 2025-08-28
Feat/no election feedback form

awdem:
This PR adds a simpler feedback form to the no election pages:
feedbackgif

It also adds an admin view with an export all to csv button:
feedbackgif2

This could definitely be DRYer and a bit more thought out, but it's likely only going to be temporary and it gets the job done. My plan is to squash all these commits into one before merging to make it easier to revert.


awdem:
Added a label to discourage personal details:
image

awdem:
> I think it would be better to ask if what would make 'this page' more useful, to be more specific.
> 
> I am a bit concerned that the feedback box, in blue, draws the eye more than the text. Feedback should be something to read after the main text. Interested what others think!

Yes it does draw the eye, but if my understanding is correct that this form is only going to be temporary, then maybe that's fine? One possibility could be to move it further down the page, but I expect less people will fill it out if we do that. Or maybe removing the blue background? Then I think it won't stick out enough..

awdem:
Final form: 
image

symroe (approved):

PR #2292, opened 2025-09-17
Updates for senedd and sp

awdem:


awdem:
> Maybe we should make it impossible for someone to add a candidate to a list election without ranking them?

Yes? I wasn't sure if it's ever possible that we know the candidate or candidates without knowing their party list position yet.

awdem:
`pr-cl.html` will need an `_cy` version. I think @pmk01 is on the case for this.

chris48s:
> Would be good to get other opinions.

I am not a fan of `?.` personally. I think either completely omit it, or maybe a `—` ?
@symroe - any thoughts on that one?

wcivf/apps/elections/templates/elections/party_list_view.html
```diff
@@ -100,15 +100,22 @@ <h4>
...
         </div>
     </div>
-
+    {% if not ballot.election.in_past %}
+        <div>
+            <p>
+                {% blocktrans trimmed with party_name=party.party_name %} This is the {{ party_name }} list. In this election you will vote for a party, not an individual candidate.
+                    Each candidate is ranked by the party, and they will be elected in order of their ranking. For example,
+                    if the {{ party_name }} win two seats in this region, the first two candidates in the list will be elected. {% endblocktrans %}
+            </p>
```
chris48s, line 109: I feel like there are some party names which aren't going to fit well in here. For example:

> if the Plaid Cymru - The Party of Wales win two seats

> if the Reform UK win two seats

I think we can weaseal out of this with..

> This is the list for `{{ party_name }}`

> if this party wins two seats

Also "region" isn't the ideal word here.. if we can easily get region/constituency as applicable lets do that. Otherwise, "division" is a less specific word that is correct for both cases.

symroe:
> > Would be good to get other opinions.
> 
> I am not a fan of `?.` personally. I think either completely omit it, or maybe a `—` ? @symroe - any thoughts on that one?

I think nothing is better than either, but agree that `?` isn't very useful.

Also, looking at this, if we have some with a list position and some without, are the ones with ordered before the null values, or after?

wcivf/apps/elections/templates/elections/party_list_view.html
```diff
@@ -14,8 +14,12 @@
 
                 <h2>{{ party_name }}</h2>
                 <h4>
-                    {{ ballot.election.nice_election_name }}
-                    {% if not ballot.is_pcc and not ballot.is_mayoral %}: {{ ballot.friendly_name }}{% endif %}
+                    {% with ballot_url=ballot.get_absolute_url election_url=ballot.election.get_absolute_url %}
+                        <a href="{{ election_url }}"> {{ ballot.election.nice_election_name }} </a>
```
chris48s, line 18: really minor, but can we trim the whitespace here so the link "hugs" the text and we don't get underlined whitespace

wcivf/apps/elections/templates/elections/party_list_view.html
pmk01: Good solutions, thanks.

pmk01:
One thought on the party list page:

Can we put the constituency name on a new line, like we do on the [ballot page](https://whocanivotefor.co.uk/elections/senedd.brycheiniog-tawe-nedd.2026-05-07/brycheiniog-tawe-nedd/)? I think this is clearer than a colon (the example above looks like the constituency name is simply 'Rhondda'

awdem:
> Can we put the constituency name on a new line

Have done so:
image
image

I don't think it's quite as clear as the ballot pages since they're both links, but it does look better.

wcivf/apps/elections/templates/elections/party_list_view.html
awdem: I've updated the explainer with the above suggestions in https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2292/commits/a48127401ba5e34886db905e0c652764b1b45ce3

pmk01:
Thanks - this looks much better. I agree that it's a bit awkward them both being links, but I also don't know what the answer is, short of adding [view all regions], or similar. This is good for now.

PR #2298, opened 2025-10-02
fix import_wikipedia_bios command

chris48s:
Refs https://app.asana.com/1/1204880536137786/project/1204880927741389/task/1211078799759181?focus=true

There were a couple of problems with this command:

1. We save every person object and every ballot object regardless of whether anything actually changed or not. This meant we were doing a lot of unnecessary database I/O but it also meant we were pointlessly touching the modified flag for all these objects, even if nothing changed.
2. The `get_wikipedia_extract()` function was set up to squash all errors and just return `None` if we get any HTTP error. Also, there were no sleeps in this command and sending a user agent like `python-requests/2.31.0` puts you in the worst rate limit bucket. This meant on any given run we were getting a lot of 429s from the wikipedia API but just returning `None` and ploughing on. The result of running this command was that we were setting data on some subset of people and ballots and removing useful data from others.

In this PR, I have done several things to fix that:

1. Only save person/ballot objects if something actually changed
2. Set a user agent with contact details. This also gets us into a better rate limit bucket
3. Implement 429 handling - we probably don't really need this as setting a user agent gives us a rate limit we will not come close to maxing out, but it is there for completeness

### DIFF
--- a/wcivf/apps/people/helpers.py
+++ b/wcivf/apps/people/helpers.py
@@ -44,7 +44,7 @@ def get_wikipedia_extract(wikipedia_url):
         sleep_time = 1
         if retry_after is not None:
             with contextlib.suppress(ValueError):
-                # we expect this to be expressed as aa number of second to wait for
+                # we expect this to be expressed as a number of seconds to wait for
                 # not a datetime to retry after
                 # https://www.mediawiki.org/wiki/Manual:Maxlag_parameter
                 sleep_time = int(retry_after)


wcivf/apps/people/helpers.py
```diff
@@ -9,15 +12,51 @@ def get_wikipedia_extract(wikipedia_url):
...
+                # not a datetime to retry after
+                # https://www.mediawiki.org/wiki/Manual:Maxlag_parameter
+                sleep_time = int(retry_after)
+        time.sleep(sleep_time)
+        resp = requests.get(url, headers=headers)
+
+    if resp.status_code == 404:
+        return ""
+
+    resp.raise_for_status()
```
chris48s, line 57: My logic here is: If we get a 404, that's going to be a 404 next time too. The content has gone away.
If we get something like a `503` (Gateway Timeout), treat that as a transient error. Raise it here and then ignore it in the calling code. In that case we won't want to delete the content we currently hold.

wcivf/apps/people/helpers.py
```diff
@@ -9,15 +12,51 @@ def get_wikipedia_extract(wikipedia_url):
...
+    Just making sequential requests we shouldn't
+    need to sleep at all to stay under this limit but we implement sleep
+    behaviour if we get a `429 Too Many Requests` back for completeness
+
+    https://foundation.wikimedia.org/wiki/Policy:Wikimedia_Foundation_User-Agent_Policy
+    https://en.wikipedia.org/api/rest_v1/
+    """
+    headers = {
+        "User-Agent": "DemocracyClubBot/1.0 (https://democracyclub.org.uk/contact/; developers@democracyclub.org.uk)"
+    }
```
chris48s, line 33: We could make this a setting/env var, if you prefer

wcivf/apps/people/helpers.py
```diff
@@ -9,15 +12,51 @@ def get_wikipedia_extract(wikipedia_url):
...
+                sleep_time = int(retry_after)
+        time.sleep(sleep_time)
+        resp = requests.get(url, headers=headers)
+
+    if resp.status_code == 404:
+        return ""
+
+    resp.raise_for_status()
+    data = resp.json()
+    return data.get("extract", "")
```
chris48s, line 59: I'm in 2 minds about whether to treat this as transient or permanent. My hunch is that this actually doesn't happen if we get a `200 OK` back, but meh. This is a big improvement on what we had before anyway.

wcivf/apps/people/management/commands/import_wikipedia_bios.py
```diff
@@ -15,30 +16,47 @@ def add_arguments(self, parser):
...
-                )
-                ballot.save()
+                current_bio = ballot.wikipedia_bio
+
+                try:
+                    new_bio = get_wikipedia_extract(ballot.wikipedia_url)
+                    if current_bio != new_bio:
+                        ballot.wikipedia_bio = new_bio
+                        ballot.save()
+                except RequestException:
```
chris48s, line 57: `except RequestException` will also catch a `JSONDecodeError` from `resp.json()`

wcivf/apps/people/management/commands/import_wikipedia_bios.py
```diff
@@ -15,30 +16,47 @@ def add_arguments(self, parser):
             help="Only import bios for current candidates",
         )
 
-    def handle(self, **options):
+    def update_people(self, current):
         people = Person.objects.exclude(wikipedia_url=None)
-        if options["current"]:
+        if current:
             current_candidacies = PersonPost.objects.current()
             people = (
```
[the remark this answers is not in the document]
symroe: Oh sorry I'd mis-read that

symroe (approved):

PR #2299, opened 2025-10-02
render good markdown, but not bad markdown

chris48s:
Refs https://app.asana.com/1/1204880536137786/project/1204880927741389/task/1211366443110685?focus=true
Refs https://docs.google.com/spreadsheets/d/1dS2Ihj_irbpJ8PSca3qVFOUBAyNGPMdwegD1ESKMKSI

In this PR I have implemented a template filter which renders a very limited subset of markdown. That is:

```md
- unordered
- bullet
- lists

**bold** _italic_ and

1. ordered
2. bullet
3. lists
```

Everything else is just passed through as text. Then I am using that filter to render candidate statements.

There's a few things worth noting here:

## Blockquotes

When we made https://docs.google.com/spreadsheets/d/1dS2Ihj_irbpJ8PSca3qVFOUBAyNGPMdwegD1ESKMKSI we said blockquotes were a maybe.

Originally I did allow these, but then I backtracked on that in 45d07e55
The reason I did this was because we always render candidate statements inside a blockquote element. I'm aware nested blockquotes are valid, but I'm less convinced they are useful. Open to feedback on it though. I could easily revert 45d07e55 if you like.

## Other places we use markdown

We use the existing markdown template filter in quite a lot of places
https://github.com/search?q=repo%3ADemocracyClub%2FWhoCanIVoteFor+markdown_filter+language%3AHTML&type=code&l=HTML
In at least some of those, we do want a full markdown renderer rather than this cut down subset.
One of the things I did consider is whether I should also switch the "full" markdown renderer to use markdown-it-py as well.
The advantage of that would be we could drop markdown and nh3 as dependencies.
The downside is: Markdown rendering is notoriously a bit finicky so we might introduce some subtle rendering differences (see the next section on lists :tada:) or just need to do a lot of checking templates.
For the moment, I've just left that as it stands but happy to do a bit more work on that if we think it is useful.

## List termination

When it comes down to it, lists are basically the one markdown feature we care about here. So it is worth thinking about this.

In CommonMark, you can start a list with a single newline. A list is terminated by **two** linebreaks. So this:

```md
This is my agenda:
- Point 1
- Point 2
- Point 3

Then this is the start of a new paragraph
```

or

```md
This is my agenda:

- Point 1
- Point 2
- Point 3

Then this is the start of a new paragraph
```

both render to

```html
<p>This is my agenda:</p>
<ul>
  <li>Point 1</li>
  <li>Point 2</li>
  <li>Point 3</li>
</ul>
<p>Then this is the start of a new paragraph</p>
```

but

```md
This is my agenda:
- Point 1
- Point 2
- Point 3
Then this is the start of a new paragraph
```

renders to

```html
<p>This is my agenda:</p>
<ul>
  <li>Point 1</li>
  <li>Point 2</li>
  <li>Point 3<br>Then this is the start of a new paragraph</li>
</ul>
```

If you regularly write markdown, this is fine. Sometimes linebreaks in list items are useful.
If you don't (or aren't even aware that you are writing markdown) it may be counter-intuitive or unexpected. I pushed a (failing) test in d5eb64d7 just to demonstrate what a user might expect in this case. I can back it out if we want.

markdown-it-py doesn't have any kind of relaxed list termination mode or third-party plugin.
If we consider this an issue, we could look at trying to write a plugin or pre-processor.

The termination behaviour of our other renderer (using the pypi package called `markdown`) is similar, except you also need 2 linebreaks to _start_ a list. So at the moment, our existing filter behaves like this (there are 3 cases):

```md
This is my agenda:

- Point 1
- Point 2
- Point 3

Then this is the start of a new paragraph
```

-->

```html
<p>This is my agenda:</p>
<ul>
  <li>Point 1</li>
  <li>Point 2</li>
  <li>Point 3</li>
</ul>
<p>Then this is the start of a new paragraph</p>
```

---

```md
This is my agenda:

- Point 1
- Point 2
- Point 3
Then this is the start of a new paragraph
```

-->

```html
<p>This is my agenda:</p>
<ul>
  <li>Point 1</li>
  <li>Point 2</li>
  <li>Point 3<br>Then this is the start of a new paragraph</li>
</ul>
```

---

```md
This is my agenda:
- Point 1
- Point 2
- Point 3
Then this is the start of a new paragraph
```

-->

```html
<p>
This is my agenda:<br />
- Point 1<br />
- Point 2<br />
- Point 3<br />
Then this is the start of a new paragraph
</p>
```

You could consider that last case as more lenient 🙃 Happy to discuss what we do about this one (if anything?), even though it is a bit bikesheddy.

## Next job: YNR

I've started off by implementing this on WCIVF. One of the goals of this ticket is consistent rendering in YNR and WCIVF so once we nail it down here we will want to make corresponding changes on YNR.
I'm sort of in 2 minds about whether to just copy and paste the whole lot (I'm mainly thinking of the test cases) or make it a shared library
Maybe it would make sense to ultimately live in dc_django_utils once it is possible to install that in YNR?
https://github.com/DemocracyClub/yournextrepresentative/blob/68b61910883f4281fcb68c9fba77af81006e920c/ynr/apps/ynr_refactoring/middleware.py#L1-L11

## Tangent: `statement_intro` / `statement_remainder`

I haven't done anything about it here, but while I was looking for statements to test I noticed that often the intro isn't very useful when we split the statement. This one particularly amused me:

Screenshot at 2025-10-10 09-11-42

While we are thinking about statements, I wonder if we should try and split this in a slightly more clever place. Maybe not today..

### DIFF
--- a/wcivf/apps/core/templatetags/markdown_filter.py
+++ b/wcivf/apps/core/templatetags/markdown_filter.py
@@ -7,6 +7,16 @@ from markdown_it import MarkdownIt
 register = template.Library()
 
 
+"""
+Note:
+We are using two different markdown rendering libraries in this module
+This is for historic reasons rather than because we strictly need to
+
+See
+https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2299#:~:text=Other%20places%20we%20use%20markdown
+"""
+
+
 @register.filter(name="markdown")
 def markdown_filter(text):
     clean_text = nh3.clean(text)


symroe:
We mostly talked about this already, but here's the written version:

>     * Do we want to enable blockquotes (`>`)?

No, not at the moment. 

>     * Do we want to try and get down to a single markdown renderer in WCIVF, or are we happy with 2 different markdown rendering packages for different contexts?

I think we want things like blockquotes in _our own_ markdown. It might make most sense to have MarkdownIt render both, one with a 'normal' parser and one with the stripped down parser. This would save on installing two libraries, but I also don't think it's a big deal . 

>     * List formats: Do you want me to have a go at a relaxed list termination plugin or pre-processor?

I think whatever we do here will be confusing to someone in some way. Better to just leave it as is and see if people can't manage it. better to spend the time on preview tooling in YNR than this. 

>     * If you're happy with saying "moving forwards, lists must be terminated by two linebreaks", do we want to do anything to try and identify/fix existing statements that are now going to render unexpectedly?

We only show statements for current candidacies, so these aren't all going to break at once. I guess if it's easy to get a sense of how much we're breaking it would be useful, but for the most part I think we'll just end up 'fixing' this as new statements replace old.   

>     * What are your thoughts on where this code should ultimately live?

Ideally in the dc utils, but as we're not using that in YNR abstracting it there is a bit pointless at this point. I think we can duplicate it in both projects and make an issue to merge it if/when we move YNR to using the utils package.

symroe:
> > We only show statements for current candidacies
> 
> Is that true on YNR as well? The intention is we adopt the same rendering code for both YNR and WCIVF, although I accept YNR is primarily nerd-facing rather than public-facing.

Yeah, we show them all the time on YNR, but remove those pages from search indexes. Things being broken there matter less, and it might be useful to seem them broken there first. I guess there's a good case for doing some checking of how many we're breaking when we get on to that project, but the impact for this project is small

symroe (approved):

PR #2300, opened 2025-10-06
Don't save in get method

symroe:
This is a bit of a refactor of the way we set and ranks. 

Previously we were calling `save()` _every time_ we got a candidate's rank. Now we use the stored rank field and ensure we set this when we import or update ballots.

---
  - https://app.asana.com/0/0/1211300441380986

### DIFF
--- a/wcivf/apps/elections/models.py
+++ b/wcivf/apps/elections/models.py
@@ -5,8 +5,8 @@ import pytz
 from django.conf import settings
 from django.contrib.humanize.templatetags.humanize import apnumber
 from django.db import models
-from django.db.models import DateTimeField, JSONField, Q
-from django.db.models.functions import Greatest
+from django.db.models import DateTimeField, F, JSONField, Q, Window
+from django.db.models.functions import DenseRank, Greatest
 from django.template.defaultfilters import pluralize
 from django.urls import reverse
 from django.utils import timezone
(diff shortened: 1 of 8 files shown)


wcivf/apps/elections/models.py
```diff
@@ -544,6 +544,28 @@ def has_results(self):
...
+        # Get all candidates with calculated ranks
+        candidates = self.personpost_set.annotate(
+            new_rank=Window(
+                expression=DenseRank(), order_by=F("votes_cast").desc()
+            )
+        ).order_by("new_rank")
+
+        # Set the rank on each candidate
+        for candidate in candidates:
+            candidate.rank = candidate.new_rank
```
awdem, line 564: It's possible the vote counts have been changed, but the ranks didn't right? I'm wondering if you should check that the rank has actually changed before overwriting it, or maybe it doesn't matter.

symroe: I didn't think it mattered, but you managed to uncover another bug in me thinking about this...we rely on modified timestamps in this project, but `bulk_update` doesn't update them. So I've moved to saving only if the rank changes. This way we update the time stamp for candidates that need it, too

wcivf/apps/people/management/commands/import_people.py
```diff
@@ -240,9 +245,33 @@ def update_candidacies(self, person_data, person_obj):
...
         self.ballot_importer.do_import(params={"election_date": date})
 
+    def calculate_ranks_for_updated_ballots(self):
+        """Calculate ranks for all ballots that had vote counts updated."""
+        if not self.updated_ballots:
+            return
+
+        ballots_with_results = [
+            ballot for ballot in self.updated_ballots if ballot.has_results
+        ]
```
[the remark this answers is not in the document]
symroe: Good point, updated

awdem:
Nice. Looks good to clean up and merge.

PR #2303, opened 2025-10-14
use filter widget from dc_utils

awdem:
We moved DSLinkWidget to dc_utils in[ 2.5.0](https://github.com/DemocracyClub/dc_django_utils/releases/tag/2.5.0) but never migrated over to using it.


symroe (approved):
Good catch!

PR #2305, opened 2025-10-14
Don't duplicate person redirects

symroe:
Few different things going on here, each commit should speak for itself.

This will fail deployment because of the existing duplicate IDs. Before we deploy I'm going to delete these, and I'll then update the created time to be some time in the past for app objects. This will force a new import of all redirects.

### DIFF
--- a/wcivf/apps/people/managers.py
+++ b/wcivf/apps/people/managers.py
@@ -179,7 +179,9 @@ class PersonManager(models.Manager):
 
             try:
                 return self.get(
-                    PersonRedirect.objects.get(old_person_id=pk).new_person_id
+                    pk=PersonRedirect.objects.get(
+                        old_person_id=pk
+                    ).new_person_id
                 )
             except PersonRedirect.DoesNotExist:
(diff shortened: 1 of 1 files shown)


wcivf/apps/people/managers.py
```diff
@@ -174,9 +174,12 @@ def update_or_create_from_ynr(self, person):
...
+        except self.model.DoesNotExist as initial_e:
             from people.models import PersonRedirect
 
-            return self.get(
-                PersonRedirect.objects.get(old_person_id=pk).new_person_id
-            )
+            try:
+                return self.get(
+                    PersonRedirect.objects.get(old_person_id=pk).new_person_id
+                )
```
chris48s: this was wrong in the original PR too, but this needs to be

```
return self.get(pk=PersonRedirect.objects.get(old_person_id=pk).new_person_id)
```

Just calling `.get(1)` rather than `.get(pk=1)` will throw `TypeError: cannot unpack non-iterable int object`

PR #2306, opened 2025-10-21
fix post caching during import

awdem:
Addresses the issue that caused: https://app.asana.com/1/1204880536137786/project/1204880927741389/task/1210995577125990?focus=true

Neither id or slug are unique on EE (and therefore YNR) so if two ballots with the same post identifier or slug are imported at the same time then only the the first one's post is saved to the cache. Appending the created date to the post or slug should make it so the the cache key is unique.

### DIFF
--- a/wcivf/apps/elections/import_helpers.py
+++ b/wcivf/apps/elections/import_helpers.py
@@ -165,14 +165,16 @@ class YNRPostImporter:
         if not post_id:
             # if no id to use return None to indicate to skip this ballot
             return None
+        created_date = ballot_dict["post"]["created"]
+        post_cache_key = f"{post_id}-{created_date}"
 
-        if post_id not in self.post_cache:
+        if post_cache_key not in self.post_cache:
             post, _ = Post.objects.update_or_create(
                 ynr_id=post_id,
                 defaults={"label": ballot_dict["post"]["label"]},
             )
-            self.post_cache[post_id] = post
-        return self.post_cache[post_id]
+            self.post_cache[post_cache_key] = post
+        return self.post_cache[post_cache_key]
 
 
 class YNRBallotImporter:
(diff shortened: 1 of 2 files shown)


chris48s:
Hmm. I'd expect it would be quite common that in a case where you've got a CA mayor and PCC election happening on the same date we'd also create the election objects on the same date in EE. Given this happens when running `import_ballots` I'd expect that is what happened in this situation where we've flagged this problem.

Also, is this actually fixing the root cause.. by which I mean: The `Post` object in WCIVF assumes the YNR Post ID/slug _is_ unique to the extent it uses it as the PK:
https://github.com/DemocracyClub/WhoCanIVoteFor/blob/0edfca50f7ca4df32cda530bb75a182276c2f4bf/wcivf/apps/elections/models.py#L356 so the WCIVF data model can't accommodate 2 different `Post` objects with the same `ynr_id`. So even if you make 2 entries in `importer.post_cache` aren't they just going to be 2 different pointers to the same `Post` object in the DB because we just did a `Post.get_or_create` on the same slug value?

awdem:
> so the WCIVF data model can't accommodate 2 different `Post` objects with the same `ynr_id`. So even if you make 2 entries in `importer.post_cache` aren't they just going to be 2 different pointers to the same `Post` object in the DB because we just did a `Post.get_or_create` on the same slug value?

Ah yeah you're right. I'm not sure about the best way to address this. 

WCIVF is wrong in assuming that the YNR Post ID/slug is unique, so ideally we'd need to change the value we're using for `ynr_id`  to something that is sufficiently unique and unchanging. As far as I can tell, the ballot_dict["post"] doesn't have any immutable, truly unique fields in it, nor does `ballot_dict` have any org info in it, so I think we're required to construct a composite key? Also if the value we're using for `Post.ynr_id` changes then we'll also need to update all the existing posts. I'm not sure if that's a simple or complex thing to do.

Alternatively, we could add some kind of check for ambiguous post ids/slugs at the point before we import them and do something to disambiguate them.

Going even further up, It would be better to add a unique key for posts to the YNR API - maybe that's the best solution? We'd still have to update all the historic posts on YNR, though.

PR #2308, opened 2025-10-23
fix leaflet importer

awdem:
The recent [django 5.2 upgrade to leaflets](https://github.com/DemocracyClub/electionleaflets/pull/405) changed the API to include timezone info in datetime string representations.

This broke the WCIVF leaflet importer because we were assuming naive datetimes.

This PR fixes that by using the dateutils parser to parse the `date_uploaded` field. It also adds some tests that include the two str formats that I encountered, plus a naive datetime. It's possible there are other formats in the database, but I'm not sure what they might be.

---
  - https://app.asana.com/0/0/1211717953002006

### DIFF
--- a/wcivf/apps/leaflets/management/commands/import_leaflets.py
+++ b/wcivf/apps/leaflets/management/commands/import_leaflets.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 
 import requests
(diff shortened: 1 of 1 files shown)


chris48s (approved):
nice :+1:

PR #2346, opened 2026-04-01
Feat/non fptp election results

awdem:
This PR adds a `results_source_url` field to `PostElection` and imports valid URLs to that field from the YNR API.

It also adds a link using the above field to non-FPTP ballot pages: 
With results table (numbers in the table are made up for testing):
image
Without table:
image


symroe:
This code is good, but I suspect we might want to hide these when the election is no longer current. I bet that the link rot is strong here, so most historic elections will show broken links. 

Let's ship this and see what happens

symroe (approved):

awdem:
> This code is good, but I suspect we might want to hide these when the election is no longer current. I bet that the link rot is strong here, so most historic elections will show broken links.
> 
> Let's ship this and see what happens

This button will only display on non-FPTP ballots, which hypothetically shouldn't have a ResultSet object (although I suspect I may have added some when I imported the historic election data) so there hopefully shouldn't be any broken links.

PR #2353, opened 2026-04-14
Boundary changes follow up frontend

awdem:
This PR makes some tweaks to the boundary change maps. It:

- Simplifies the line interpolation. The minimum line width is now 3 to distinguish it from highways on OSM and it there are less unneccesary steps
- Adds popups on click for the old and new layers, as well as for the postcode centroid.
- changes the cursor to a pointer when passing over interactive layers

boundarychangestweaks


wcivf/apps/elections/templates/elections/boundary_reviews_view.html
```diff
@@ -354,20 +357,18 @@ <h4>{% trans "Map" %}</h4>
...
-                            map.setFilter("new_division_set_layer_fill", ['==', ["get", "division__official_identifier"], change.new_division_official_identifier])
-                        }
+                        map.addLayer({
+                            "id": "new_division_set_layer_fill",
+                            "type": "fill",
+                            "source": "new_division_set",
+                            "source-layer": `${newDivisionsetID}_${change.division_type}`,
+                            "paint": {
+                                "fill-color": "#E6007C",
+                                "fill-opacity": ShowNewBoundaryFill ? 0.2 : 0
```
GeoWill: I'm assuming this is done to make the popups work I.e. we need a transparent fill. 
If this is the case, I don't think this needs to be a conditional. It looks like  `ShowNewBoundaryFill` is set to `true`, then not used, then set to `false` and evaluated here. Shouldn't we just bin it, and set opacity to `0`?
If I've missed something and we do need `ShowNewBoundaryFill`, can it be camel case: `showNewBoundaryFill`.

wcivf/apps/elections/templates/elections/boundary_reviews_view.html
```diff
@@ -249,6 +249,14 @@ <h4>{% trans "Map" %}</h4>
             legendDiv.appendChild(row)
         }
 
+        const getFeatureData = (feature) => {
+            return {
+                name: feature.properties.division__name || "Name Missing",
+                layerColor: feature.layer.paint['fill-color'] || feature.layer.paint['line-color'],
+                isNew: feature.source.startsWith("new") ?  true : false
```
GeoWill: [`startsWith`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) returns a bool, so I don't think you need the ternary (`?  true : false`)

wcivf/apps/elections/templates/elections/boundary_reviews_view.html
```diff
@@ -407,6 +415,40 @@ <h4>{% trans "Map" %}</h4>
...
+                            featsData.push(getFeatureData(feat));
+
+                            seen.add(featId);
+                        }
+
+                        const popUpHTML = featsData.map(feat => `
+                            <div style="color: ${feat.layerColor};">
+                                ${feat.isNew ? "New" : "Old"}: <strong style="color: ${feat.layerColor};">${feat.name}</strong>
+                            </div>
+                            `).join('');
```
GeoWill: I know we control the pmtiles, but I think it's probably better here not to just chuck the colour and name strings into the html without sanitizing it. This is more so that we don't go and copy and paste it somewhere else when we shouldn't, rather than it being a massive concern here. It's also confusing because there's a js built in [`setHTML`](https://developer.mozilla.org/en-US/docs/Web/API/Element/setHTML) which does sanitize input, while maplibre popup [`setHTML`](https://maplibre.org/maplibre-gl-js/docs/API/classes/Popup/#sethtml) doesn't...

I think if you build the popup up out of dom elements, and then use [`setDOMContent`](https://maplibre.org/maplibre-gl-js/docs/API/classes/Popup/#setdomcontent), it will avoid the issue. 

```javascript
const popupContent = document.createElement('div');
featsData.forEach(feat => {
    const featContainer = document.createElement('div');
    featContainer.style.color = feat.layerColor;

    const label = document.createTextNode(feat.isNew ? "New: " : "Old: ");
    const strong = document.createElement('strong');
    strong.style.color = feat.layerColor;
    strong.textContent = feat.name;

    featContainer.appendChild(label);
    featContainer.appendChild(strong);
    popupContent.appendChild(row);
});

const popup = new maplibregl.Popup({ closeButton: true, closeOnClick: true })
    .setLngLat(coords)
    .setDOMContent(popupContent);
showPopup(popup);
```

wcivf/apps/elections/templates/elections/boundary_reviews_view.html
```diff
@@ -418,24 +435,19 @@ <h4>{% trans "Map" %}</h4>
...
 
                         if (!feats.length) {
                             return;
                         }
 
                         let featsData = [];
-                        let seen = new Set();
 
                         for (const feat of feats) {
                             const featId = feat.properties.id
```
GeoWill: You don't  use `featId` any more now.

wcivf/apps/elections/templates/elections/boundary_reviews_view.html
```diff
@@ -410,8 +426,22 @@ <h4>{% trans "Map" %}</h4>
                         if (bounds) {
                             map.fitBounds(bounds, {padding: 20, duration: 0, });
                         };
+                        const markerPopup = new maplibregl.Popup({
+                            closeButton: true,
+                            closeOnClick: true,
+                            offset: 25,
+                        })
+                            .setLngLat(postcodeLocation.geometry.coordinates)
+                            .setText("Postcode Centroid");
```
GeoWill: I assume the actual postcode is available in the context, so I'd lean to including it here rather than just poscode.

GeoWill (changes requested):
I haven't run this locally, but I found the screencast extremely useful! I would mostly be checking that the popups open and close as I would expect, and it works nicely on constituencies and regions (i.e. there's no accidental sharing of feature data between the maps or some such).

There's a few comments on the code.  

Regarding showing the postcode in the popup I think that's a nice to have rather than needed.

wcivf/apps/elections/templates/elections/boundary_reviews_view.html
awdem: This is so the fill appears when when we're only showing the new boundaries (e.g. no change, or all Welsh postcodes), but is transparent when the old boundaries are also on the map.

GeoWill: Got it. 
In that case I think it might actually be clearer to make this: 
```javascript
"fill-opacity": !showOldBoundary ? 0.2 : 0
```

wcivf/apps/elections/templates/elections/boundary_reviews_view.html
awdem: Done in [7e30807](https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2353/commits/7e30807d2995be1d98922f14798b9110f3c71a4d)

wcivf/apps/elections/templates/elections/boundary_reviews_view.html
awdem: That makes more sense. Have changed it in [0b33dac](https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2353/commits/0b33dac043f66c3b0f87a3d37be2a6ce1840188e)

wcivf/apps/elections/templates/elections/boundary_reviews_view.html
awdem: You're right. Fixed in [6a58dd9](https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2353/commits/6a58dd9f41f77e5ef567ed1a49266167b9bb12a1)

wcivf/apps/elections/templates/elections/boundary_reviews_view.html
awdem: Ah I had not thought about that. I've implemented the above in [c8f59a7](https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2353/commits/c8f59a732c19a0eae80f1a502e166910b2483fce).

wcivf/apps/elections/templates/elections/boundary_reviews_view.html
awdem: Whoops, fixed in https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2353/commits/ba5f123fe5d1fae92a24af606b056f4de0851f18

awdem:
> I would mostly be checking that the popups open and close as I would expect, and it works nicely on constituencies and regions (i.e. there's no accidental sharing of feature data between the maps or some such).

I just double-checked a few scottish postcodes and didn't find any accidental interactions between maps. popups acting normal as well.

GeoWill:
I've pushed https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2353/commits/ba5f123fe5d1fae92a24af606b056f4de0851f18 to `development` to trigger a deploy. I'll have a quick play around, but  I think this looks good.

awdem:
> One more small issue. It's a bit annoying that the popup initialises with the `close` button highlighted, as it obscures the last letter of the postcode. I'm not sure why it's happening, I guess and easy fix would be to add a bit more padding?

Turning this Mablibre popup option to `False` stops the button from being focused: https://maplibre.org/maplibre-gl-js/docs/API/type-aliases/PopupOptions/#focusafteropen

But I don't if that has any accessibility implications? Happy to change the padding but wanted to see what you think.

GeoWill:
I'm not sure I'm afraid. 
It does take a lot of tabbing to get round the whole page though. So maybe best to just pad it?

awdem:
> I'm not sure I'm afraid. It does take a lot of tabbing to get round the whole page though. So maybe best to just pad it?

image

Done in [ac35c9a](https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2353/commits/ac35c9ab2afba2dd512652308e9720c6f45e8e26)

PR #2354, opened 2026-04-14
Lucy errors

jargonautical:
```[tasklist]
### PR Checklist
-between last commit and this one, something is going wrong with decoding (unicode error on runserver, instantaneous in Chrome but only happens on changing the language to Welsh in either Safari or Firefox.
```


chris48s:
Replaced by https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2360

PR #2357, opened 2026-04-15
Use node

awdem:
This PR moves the JS packages used for boundary change maps to be managed by npm rather than imported from a CDN. I've tested by deploying to [dev.wcivf.club](https://dev.wcivf.club/) and everything seems to be working.

I also did a small clean up of the CI config in https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2357/commits/46e3a5276d2e9eba50b0e4f9ab5e4af316d8c679 by removing the py312 execution env. I think we used to do some steps in a matrix with multiple python versions, but we no longer do and the execution env is redundant.

One thing I think will need a close inspection is [65b20ab](https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2357/commits/65b20abe2a52098964821028635547a613a6ca1a) where I add node to CI. I'm not sure I've put it in all the places it needs to be.

Follow up:

We are still importing leaflet from a CDN here: 
https://github.com/DemocracyClub/WhoCanIVoteFor/blob/aba09d6ae19ea8cecbcc8ef9965ca85190f81f8f/wcivf/apps/elections/templates/elections/includes/_polling_place.html#L166-L167

We could move to just using maplibre rather than depending on two mapping libraries, but I decided against including this in the PR. I will make a follow up card to return to this after May.

### DIFF
--- a/deploy/after_install/install_npm_deps.sh
+++ b/deploy/after_install/install_npm_deps.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+set -xeEo pipefail
+
+cd /var/www/wcivf/code/
+npm ci
--- a/package.json
+++ b/package.json
@@ -1,0 +1,28 @@
+{
+  "name": "whocanivotefor",
+  "version": "1.0.0",
+  "description": "",
+  "directories": {
+    "doc": "docs"
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/DemocracyClub/WhoCanIVoteFor.git"
+  },
+  "author": "",
+  "license": "BSD-3-Clause",
+  "bugs": {
+    "url": "https://github.com/DemocracyClub/WhoCanIVoteFor/issues"
+  },
+  "homepage": "https://github.com/DemocracyClub/WhoCanIVoteFor#readme",
+  "dependencies": {
+    "maplibre-gl": "5.7.3",
+    "pmtiles": "4.3.0"
+  },
+  "engines": {
+    "node": ">=18"
+  }
+}
--- a/wcivf/apps/elections/templates/elections/boundary_reviews_view.html
+++ b/wcivf/apps/elections/templates/elections/boundary_reviews_view.html
@@ -1,5 +1,7 @@
 {% extends "base.html" %}
 {% load i18n %}
+{% load pipeline %}
+{% load static %}
 
 {% block page_title %}{% trans "Boundary Reviews"%}{% endblock page_title %}
 {% block page_meta %}
@@ -8,6 +10,7 @@
 
 {% block extra_site_css%}
     {{ block.super }}
+    {% stylesheet 'maplibre-gl' %}
     <style>
         div[class$="-map-legend"] {
             display: none;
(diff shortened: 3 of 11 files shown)


deploy/after_install/install_npm_deps.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+set -xeE pipefail
```
GeoWill: `pipefail` won't do anything without `o`. i.e. `set -xeEo pipefail`. 
But having searched I realise this mistake is all over our code bases.

wcivf/apps/elections/templates/elections/boundary_reviews_view.html
```diff
@@ -158,13 +161,13 @@ <h4>{% trans "Map" %}</h4>
...
+
     {{ boundary_reviews|json_script:"boundaryReviewsData" }}
     {{ postcode_location|json_script:"postcodeLocation" }}
     {{ nation|json_script:"nation" }}
 
-    <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/maplibre-gl/5.7.3/maplibre-gl.css" integrity="sha512-IylZQnfgxyZ4rs0/u4TjuOthOSWKwh47nrCPuscGnyidnt1mrhKdja/uUxfvkLaMusDAWV3DgKHHpfzNdTdGhw==" crossorigin=""/>
-    <script src="//cdnjs.cloudflare.com/ajax/libs/maplibre-gl/5.7.3/maplibre-gl.js" integrity="sha512-4qLefPK7TDle+A82wrHOa/B75+B8HH0w6Zwu6YaoXxwBrLl8tQ7UmoyS0PCWOobcq3fFe51G+AN+jVZpwN5qOw==" crossorigin=""></script>
-    <script src="//unpkg.com/pmtiles@4.4.0/dist/pmtiles.js" integrity="sha256-SqjwRpVddxFvwcm7cbxJ0XRlQ0KvjxKEqjTGJccf1nY=" crossorigin=""></script>
+    <script src="{% static 'maplibre-gl/dist/maplibre-gl.js' %}"></script>
+    {% javascript 'pmtiles' %}
```
GeoWill, line 170: Is there a reason these are done differently?

package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "whocanivotefor",
+  "version": "1.0.0",
+  "description": "[![CircleCI](https://circleci.com/gh/DemocracyClub/WhoCanIVoteFor.svg?style=svg)](https://circleci.com/gh/DemocracyClub/WhoCanIVoteFor) [![Coverage Status](https://coveralls.io/repos/github/DemocracyClub/WhoCanIVoteFor/badge.svg?branch=master)](https://coveralls.io/github/DemocracyClub/WhoCanIVoteFor?branch=master) ![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)",
```
GeoWill: I don't think this badge markdown should be here

GeoWill (commented):
Couple of comments, but looking pretty good.

wcivf/apps/elections/templates/elections/boundary_reviews_view.html
awdem: Copied from here: https://github.com/DemocracyClub/EveryElection/commit/32ac04bf68f0bdd591f07e0b90b1420247749147 

I don't remember the exact details, but I think that the maplibre-gl entry file is minified in some way and when I tried to import it with pipeline it broke everything so instead I imported it directly from static.

package.json
awdem: It was the default when I left it blank during `npm init` - I don't know why. We also have it in the EE [package.json](https://github.com/DemocracyClub/EveryElection/blob/master/package.json) so I decided to leave it, but happy to remove it as well.

wcivf/apps/elections/templates/elections/boundary_reviews_view.html
GeoWill: Fair enough! Let's do it like that.

package.json
GeoWill: Seems tidier to delete but, doesn't really matter.

deploy/after_install/install_npm_deps.sh
awdem: Fixed in f012722f2083a1f369e7057b961924c68b8a56f5

I copied unknowingly from https://github.com/DemocracyClub/WhoCanIVoteFor/blob/a97ad6b4c59f40749edb702c4026808e33ac9890/deploy/after_install/install_python_deps.sh

Does it matter here if the script isn't using pipes at all?

package.json
awdem: Removed in https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2357/commits/a97ad6b4c59f40749edb702c4026808e33ac9890

GeoWill (approved):
drop the deploy commit and rebase and it's good to go.

PR #2360, opened 2026-04-16
errors fixed, final translations included

jargonautical:
```[tasklist]
### PR Checklist

```


jargonautical:
A few minor bits I haven't been able to resolve; they're tagged but not translating.
reason for by election translates but the rest of the sentence doesnt
Button text tagged but not translating
Footer list item DONATE tagged but not translating

locale/cy/LC_MESSAGES/django.po
```diff
@@ -7,13 +7,13 @@
 # William Roper <roper.william@gmail.com>, 2022
 # Atebol, 2022
 # Sym Roe <sym@talusdesign.co.uk>, 2022
-#
-#, fuzzy
+#: wcivf/apps/people/templates/people/not_current_person_detail.html:33
```
jargonautical, line 10: makemessages keeps adding this back in, going to investigate why

chris48s:
OK, so the reason why you've got a load of text where you've put translations in the po file but it isn't coming out translated on the web page is because they are still marked as fuzzy.

Django won't display translations marked as fuzzy (this is the inverse design decision to what we see on the Electoral Commission site).

So if you take, for example:

```po
#: wcivf/apps/feedback/models.py:16
#, fuzzy
#| msgid "Boundary changes"
msgid "No change"
msgstr "Dim newid"
```

..and edit that to

```po
#: wcivf/apps/feedback/models.py:16
msgid "No change"
msgstr "Dim newid"
```

Then re-compile the translations, then you'll see the "No change" button is translated to "Dim newid" on the website.

You've got about 100 translations in the PO file still marked as fuzzy. Can you go through those and push another commit to this branch updating them. I'm guessing in most cases, we can just remove the fuzzy comments as you've reviewed and updated the translations hopefully. If they're still bad, just explicitly set `msgstr ""`.

I can take it from there.

locale/cy/LC_MESSAGES/django.po
chris48s: I can pick up fixing this one.

PR #2372, opened 2026-04-27
Fix local parties importer

awdem:
This transaction wasn't actually doing anything because it was using the default database but all the writing was being done on the principal. See:
https://github.com/DemocracyClub/UK-Polling-Stations/pull/8224
https://github.com/DemocracyClub/UK-Polling-Stations/pull/8605
[Chris's blog](https://chris48s.github.io/blogmarks/posts/2025/django-db-routers-1/#:~:text=expected%20DB%20connection.-,Transactions,-This%20one%20is)

That meant that when `LocalPartyImporter` deleted all the parties/manifestos, they would disappear from the site until they were re-imported.

We're going to have to do a big review of all the other transactions in WCIVF, but I that's for another PR.

### DIFF
--- a/wcivf/apps/core/db_routers.py
+++ b/wcivf/apps/core/db_routers.py
@@ -27,3 +27,9 @@ class PrincipalRDSRouter:
 
     def allow_migrate(self, db, app_label, model_name=None, **hints):
         return True
+
+
+def get_principal_db_name():
+    if PRINCIPAL in settings.DATABASES:
+        return PRINCIPAL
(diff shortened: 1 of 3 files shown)


chris48s:
Yeah. Lets get this merged :+1:

In terms of next steps, I think the core `import_people`, `import_ballots`, `import_parties` commands are the next things to look at. Then we move on to more peripheral things.

chris48s (approved):

PR #2375, opened 2026-04-28
Refactor results text

awdem:
This PR refactors the text in the results column of the previous candidacies table. It:
- makes the text for non-FPTP elections consistent
- Makes the Elected/Not elected text bold and makes the vote count text smaller
- Elected/Not elected always appears before the vote count

To do this I've simplified `get_results_text` by moving the vote count text logic to a new property, `get_vote_count_text` and putting the logic for determining text when an election is in the past back into to the template.
Old:
image
New:
image
Old:
image

New:
image
 

---
  - https://app.asana.com/0/0/1214259695023894

### DIFF
--- a/wcivf/apps/people/models.py
+++ b/wcivf/apps/people/models.py
@@ -63,11 +63,10 @@ class PersonPost(models.Model):
     @property
     def get_results_text(self) -> str:
         """
-        Text describing the outcome of the election, as relates to this person.
+        Returns text describing whether or not this person was elected.
+        For example, if they were elected, return:
 
-        For example, if they were elected and we have the vote count, return:
-
-            "Elected (12345 votes)"
+            "Elected"
         """
 
         # Cancelled ballot:
@@ -79,42 +78,30 @@ class PersonPost(models.Model):
         if self.post_election.cancelled:
             if self.elected:
                 return _("Elected unopposed")
-            return _("(election cancelled)")
+            return _("(Election cancelled)")
 
-        # If the election is in the past, we can't have results!
-        # NOTE: this needs to be after the cancelled checks, as an election
-        # is always cancelled before the election day(!)
-        if "election" in self._state.fields_cache:
-            election = self._state.fields_cache["election"]
-        else:
-            election = self.post_election.election
-
-        if election and not election.in_past:
-            return _("(Results due after %(date)s)") % {
-                "date": election.election_date.strftime("%d %B %Y")
-            }
-
-        # If this candidacy was marked as elected, we need to show that!
-        # But we might not always have votes case for them.
-        # This is either because no one has gone and entered them, or
-        # because it's not an election we don't support votes cast for
-        # (e.g. non-FPTP)
-        num_votes = intcomma(self.votes_cast)
         if self.elected:
-            if self.votes_cast:
-                return _("%(num_votes)s votes (elected)") % {
-                    "num_votes": num_votes
-                }
-            return _("Elected (vote count not available)")
+            return _("Elected")
+        return _("Not elected")
 
-        # If we have votes cast, but we've not marked the candidacy as elected
-        # then the person wasn't elected, but we should still show the votes
-        # they got.
-        if self.votes_cast:
-            return _("%(num_votes)s votes (not elected)") % {
-                "num_votes": num_votes
+    @property
+    def get_vote_count_text(self) -> str:
+        """
+        Returns text describing the vote count for this candidacy, formatted with commas.
+
+        For example, "12,345 votes".
+        """
+        if self.post_election.get_voting_system.slug != "FPTP":
+            return _(
+                "We do not collect voting data for %(voting_system)s elections"
+            ) % {"voting_system": self.post_election.get_voting_system.get_name}
+
+        if self.votes_cast is not None:
+            return _("%(vote_count)s votes") % {
+                "vote_count": intcomma(self.votes_cast)
             }
-        return _("Not elected (vote count not available)")
+
+        return _("vote count not available")
 
     @property
     def get_results_rank(self):
--- a/wcivf/apps/people/templates/people/includes/_person_previous_elections_card.html
+++ b/wcivf/apps/people/templates/people/includes/_person_previous_elections_card.html
@@ -31,7 +31,10 @@
                                 (Results due after {{ date }})
                             {% endblocktrans %}
                         {% else %}
-                            <strong>{{ person_post.get_results_text }}</strong> <span class="ds-font-size-s1">({{ person_post.get_vote_count_text }})</span>
+                            <strong>{{ person_post.get_results_text }}</strong>
+                            {% if not person_post.post_election.cancelled %}
+                                <span class="ds-font-size-s1">({{ person_post.get_vote_count_text }})</span>
+                            {% endif %}
                         {% endif %}
                     </td>
                     {% if object.previous_party_count %}
--- a/wcivf/apps/people/templates/people/includes/_person_previous_elections_card.html
+++ b/wcivf/apps/people/templates/people/includes/_person_previous_elections_card.html
@@ -26,12 +26,10 @@
                     </td>
                     <td>{{ person_post.party_name }}</td>
                     <td>
-                        {% if person_post.election.voting_system.slug == "sv" %}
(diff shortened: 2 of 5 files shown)


wcivf/apps/people/templates/people/includes/_person_previous_elections_card.html
```diff
@@ -26,12 +26,10 @@
                     </td>
                     <td>{{ person_post.party_name }}</td>
                     <td>
-                        {% if person_post.election.voting_system.slug == "sv" %}
-                            {% trans "We do not collect voting data for Supplementary Vote elections." %}
-                        {% elif person_post.election.voting_system.slug == "STV" %}
-                            {% trans "We do not collect voting data for Single Transferable Vote elections." %}
-                        {% else  %}
-                            {{ person_post.get_results_text }}
+                        {% if not person_post.election.in_past %}
```
GeoWill: Does this mean that a future cancelled election will show `results due after ... `? Feels like there should be a `test_future_cancelled_elections` test. I think this reverts the behaviour described in the deleted comment in the `get_results_text` method. Now the behaviour is split over model and template I think it should be tested.

wcivf/apps/people/templates/people/includes/_person_previous_elections_card.html
```diff
@@ -26,12 +26,10 @@
...
+                            (Results due after {{ person_post.election.election_date|date:"d M Y" }})
```
GeoWill: This won't be translated. 
It either wants to be pushed up into the `get_results_text` method, and wrapped in `_()` or should be something like: 

```
{% blocktrans trimmed with date=person_post.election.election_date|date:"d M Y" %}
  (Results due after {{ date }})
{% endblocktrans %}
```

I think as long as `{% load i18n %}` is included, and we have the locale available, then the date will translate. But I haven't checked this.

wcivf/apps/people/models.py
```diff
@@ -79,42 +78,30 @@ def get_results_text(self) -> str:
...
+                "(We do not collect voting data for %(voting_system)s elections)"
+            ) % {"voting_system": self.post_election.get_voting_system.get_name}
+
+        if self.votes_cast is not None:
+            return _("(%(vote_count)s votes)") % {
+                "vote_count": intcomma(self.votes_cast)
             }
-        return _("Not elected (vote count not available)")
+
+        return _("(vote count not available)")
```
GeoWill: This is minor, but all these return values are wrapped in brackets now. I actually find it makes parsing the strings quite a bit harder. So I'd be tempted to put the brackets in the template, rather than here in the method. That might just be me though!

awdem: Agreed! I think this was an artifact of a previous iteration. Fixed here: [bcb694d](https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2375/commits/bcb694dc469ccabc0197c2643db1b09a3050939f)

wcivf/apps/people/templates/people/includes/_person_previous_elections_card.html
awdem: Have added a blocktrans tag: [b0692c4](https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2375/commits/b0692c401a86bb3c3793158939f884bbdc9009e7)

Also, not sure if I need to generate/update the `.po` file?

wcivf/apps/people/templates/people/includes/_person_previous_elections_card.html
awdem: You're right, have added tests and fixed in: [2140a3e](https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2375/commits/2140a3e750b2790a9e0092f6f847735728456bdc)

wcivf/apps/people/tests/test_person_views.py
```diff
@@ -393,9 +393,79 @@ def test_previous_elections_card(self):
...
+            name="Welsh Assembly Election",
+            current=False,
+            election_date="2100-05-01",
+            slug="local.welsh.assembly.2100-05-01",
+        )
+        future_cancelled_post_election = PostElectionFactory(
+            election=election,
+            post=PostFactory(label="Welsh Assembly", territory="WLS"),
+            ballot_paper_id="local.welsh.assembly.2100-05-01",
+            voting_system=VotingSystem(slug="FPTP"),
```
GeoWill, line 483: Shouldn't this use `VotingSystemFactory` once and set it for the `Election` and `PostElection`?

Related I guess you added `VotingSystemFactory` because a bunch of tests errored if it wasn't set. So follow up question, is do all `Election`/`PostElection` objects in the DB have it set?

wcivf/apps/people/tests/test_person_views.py
```diff
@@ -393,9 +393,79 @@ def test_previous_elections_card(self):
...
+        """Test that the postcode search results with the text
+        'Results due after [election date]'
+        """
+        election = ElectionFactory(
+            name="Welsh Assembly Election",
+            current=False,
+            election_date="2100-05-01",
+            slug="local.welsh.assembly.2100-05-01",
+        )
+        future_cancelled_post_election = PostElectionFactory(
```
GeoWill: I think this should be `future_post_election`

wcivf/apps/people/models.py
```diff
@@ -79,42 +78,30 @@ def get_results_text(self) -> str:
...
+        Returns text describing the vote count for this candidacy, formatted with commas.
+
+        For example, "12,345 votes".
+        """
+        if self.post_election.get_voting_system.slug != "FPTP":
+            return _(
+                "We do not collect voting data for %(voting_system)s elections"
+            ) % {"voting_system": self.post_election.get_voting_system.get_name}
+
+        if self.votes_cast is not None:
```
GeoWill, line 99: I just thought of something about this on second reading. Do we have any counts in the DB that are `0` instead of `None`? i.e. has `0` been used as a place holder anywhere? If so we will incorrectly show candidates with `0` votes cast when it should be `Vote count not available`.

wcivf/apps/people/templates/people/includes/_person_previous_elections_card.html
GeoWill: Does this fit in neatly with any of the translation work ticketed in sprint plan? if so I'd loop it in with that, rather than blocking this pr over it.

GeoWill (commented):
Couple of small comments/questions, but think this is pretty much there.

wcivf/apps/people/tests/test_person_views.py
awdem: Ah yes in fact I can just remove this line because I've made it so that FPTP is the default `VotingSystem` for all `Election`s created with the  Election factory class and when `PostElection`s don't have a `VotingSystem` set they default to using their `Election`s `VotingSystem`. 

Had a look in the DB and there is only one `PostElection` that doesn't have it set, but it looks like it shouldn't exist anyway judging from the name and the 500 when I try to go the page: `tmp_local.south-buckinghamshire.farnham-royal-and-hedgerley.2016-05-05.DIW:E05010574`. 

There are 47 `Election`s that don't have a voting system set, but I don't think anything will go wrong since all of the `PostElection`s (except the one above) have them set?

wcivf/apps/people/models.py
awdem: There 305 instances of votes_cast = 0 . However I don't know how to tell if 0 is a placeholder that means we don't actually know the votes, or a case of someone getting 0 votes?

wcivf/apps/people/templates/people/includes/_person_previous_elections_card.html
awdem: There is a ticket to fill out missing translations for work that wasn't captured in the last translations PR so yeah I think it's captured elsewhere.

wcivf/apps/people/tests/test_person_views.py
awdem: Fixed in https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2375/commits/59e075f037b5d09e1188cd3013ea6e8e8913ae34

GeoWill (approved):

PR #2399, opened 2026-05-28
fixes for postal votes text

chris48s:
Refs https://app.asana.com/1/1204880536137786/project/1204880927741389/task/1215202625478087?focus=true

7574a725 is not strictly related, but worth fixing while I was here

Refs https://github.com/DemocracyClub/ec-postcode-lookup-pages/pull/239

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,7 +100,7 @@ jobs:
 
   sam_build:
     docker:
-      - image: public.ecr.aws/sam/build-python3.12:latest
+      - image: public.ecr.aws/sam/build-python3.12:1.158
     working_directory: ~/repo
     steps:
       - checkout
@@ -117,7 +117,7 @@ jobs:
 
   sam_deploy:
     docker:
-      - image: public.ecr.aws/sam/build-python3.12:latest
+      - image: public.ecr.aws/sam/build-python3.12:1.158
     working_directory: ~/repo
     parameters:
       dc-environment:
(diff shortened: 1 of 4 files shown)


chris48s:
I will come back to this build failure next week. It is not really related to the changes in this PR

.circleci/config.yml
```diff
@@ -100,7 +100,7 @@ jobs:
 
   sam_build:
     docker:
-      - image: public.ecr.aws/sam/build-python3.12:latest
+      - image: public.ecr.aws/sam/build-python3.12:1.158
```
chris48s, line 103: This is a temporary solution - it allows us to deploy this work

If you look at https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2400 that demonstrates the problem (basically, `public.ecr.aws/sam/build-python3.12` now ships with UV pre-installed and ignores the version we install).

For now, lets pin it so we can deploy this.. but I think we need to look at:
a) how to get the SAM build image to prefer a specific version of UV so this doesn't keep breaking
b) upgrading UV across projects

chris48s: This same issue will be affecting other projects that use this image

wcivf/apps/elections/views/mixins.py
```diff
@@ -292,11 +292,13 @@ def get_global_postal_vote_card(self, post_elections, council):
             "application_deadline": next_ballot.postal_vote_application_deadline,
             "election_date": next_ballot.election.election_date,
             "sopn_date": next_ballot.expected_sopn_date,
+            "show_dispatch_date_fallback": False,
         }
 
         # we only hold postal votes dispatch data data for one
         # election. TODO: remove/review after 2026-05-07
         if next_ballot.election.election_date == date(2026, 5, 7):
```
symroe, line 300: Is it worth making this more future proof by defining a 'first Thursday in May' check?

symroe: Happy to punt this to an issue, actually.

symroe (approved):

PR #2401, opened 2026-06-03
Prioritise our UV over the UV pre-installed in the SAM image

chris48s:
Here's an explanation of the problem:
https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2399#discussion_r3333562216

In #2399 I fixed this temporarily by just pinning to an old version of the docker image.
In this PR I've gone back to the `:latest` tag but fixed it so that "our" version of UV takes priority over the one that comes pre-installed.

I've tested this deploys cleanly to dev

Assuming we are happy with this, I will apply the same fix on our other projects that use the `public.ecr.aws/sam/build-python3.12` image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,13 +100,16 @@ jobs:
 
   sam_build:
     docker:
-      - image: public.ecr.aws/sam/build-python3.12:1.158
+      - image: public.ecr.aws/sam/build-python3.12:latest
     working_directory: ~/repo
     steps:
       - checkout
       - run:
           name: Install UV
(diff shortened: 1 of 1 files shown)


GeoWill (approved):
Looks good. Also TIL about `/var/lang/`.

chris48s:
I don't think /var/lang is special or standard. I think it is just a Circle CI thing.

If we `pip install` something, our packages go into `/var/lang/lib/python3.12/site-packages` and the corresponding bin directory is `/var/lang/bin`

PR #2407, opened 2026-06-08
improve parsing of usernames from social links

### LINKED ISSUE #2356: Buggy parsing of Facebook names
A Facebook URl like `https://m.facebook.com/Tony.Blair/?locale=en_GB` is rendered as `?locale=en_GB`

This is because the name extraction is only looking at the last element after the last slash:

https://github.com/DemocracyClub/WhoCanIVoteFor/blob/aba09d6ae19ea8cecbcc8ef9965ca85190f81f8f/wcivf/apps/people/models.py#L295-L299

Result:

Candidate with badly rendered name.

I'd suggest either stripping query strings before rendering, or using a URl parsing library to extract the path without the query string.

chris48s:
closes #2356

add a bunch of test cases for various cases + fix the URL parsing bugs

Ideally, we should also ensure these are clean at source as well but it makes sense to be defensive here too.

### DIFF
--- a/wcivf/apps/people/models.py
+++ b/wcivf/apps/people/models.py
@@ -275,33 +275,30 @@ class Person(models.Model):
 
     @property
     def facebook_personal_username(self):
-        facebook_personal_url = self.facebook_personal_url
-        facebook_split = list(filter(None, facebook_personal_url.split("/")))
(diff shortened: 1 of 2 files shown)


awdem (approved):
LGTM

PR #2408, opened 2026-06-10
WAF

symroe:
Disclosure: I did about 90% of this with Claude Code / Opus 4.8. On a new branch, I prompted:

> This project deploys to AWS using SAM. See @sam-template.yaml. It uses CloudFront, and I want to install Web Application Firewall in front of it to stop bot traffic. Please read @../yournextrepresentative/cdk/stack/waf.py and copy that implementation from CDK into SAM. I want the same rules, features and  blocks as that application uses, but defined here in SAM's YAML. If possible, please make this using a new template YMAL that's included in the main one, but ignore this if that's complex. I just want maintainable code.

After some thinking it came back with:

>  This confirms the architecture. The key finding: YNR deploys the WAF as a completely separate stack in us-east-1 (cdk/app.py:42), and passes the resulting ARN into the CloudFront distribution. This is mandatory — CloudFront-scoped WAFv2 WebACLs must live in us-east-1, but this SAM stack deploys to eu-west-2 (samconfig.toml).
> 
>   That means a nested/included stack won't work — a nested stack inherits the parent's region (eu-west-2), which a CLOUDFRONT scope WAF rejects. So the maintainable equivalent of the CDK setup is: a standalone waf-template.yaml deployed to us-east-1, with its WebACL ARN wired into the main template's CloudFront distribution via a parameter. Let me build that.

I'll not paste the whole chat here, but there was some follow up to get it to change tagging. I then asked it to update CircleCI so that WAF deploys with that. 

I did a dev deploy (manually changing the branch). I needed to update some IAM permissions on the dev account. I've not done this on the other accounts, but I can do before deploying. 

---

## My review of Claude's performance:

I'm sort of over being impressed by this any more, but the whole thing took about 5 minutes from initial prompt to the first two commits here. Faffing about with IAM took a bit longer. This was helped by the fact that the deploys are very similar (in terms of the CloudFront passing on to an ALB, not in terms of YNR's ECS deployment). 

The LLM work was to convert one concept in one language to the same concept in another, and this is something I'd expect them to be fast at. 

I've read the code, and done a deploy. There are more comments than I'd have written, but that's no bad thing. I think it's about as understandable and maintainable as anything I'd have written (everyone loves YAML, afterall).  

---
  - https://app.asana.com/0/0/1215579151691311

### DIFF
--- a/waf-template.yaml
+++ b/waf-template.yaml
@@ -1,0 +1,235 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Description: >
+  WCIVF WAF
+
+  Simple bot-protection Web ACL for the WCIVF CloudFront distribution.
+
+  IMPORTANT: CloudFront-scoped WAFv2 resources MUST be created in us-east-1,
+  regardless of which region the main application stack is deployed to. This
+  template is therefore deployed as a SEPARATE stack in us-east-1, and its
+  output WebACL ARN is passed into the main sam-template.yaml as the
+  `WebACLArn` parameter.
+
+  This is a direct port of the CDK WafStack used by yournextrepresentative
+  (cdk/stack/waf.py) - same rules, same blocks, same managed rule overrides.
+
+  Tagging is applied at the stack level via the deploy command's `--tags`
+  flag (dc-environment / dc-product / CreatedVia) and propagates to every
+  resource here, so the resources themselves carry no Tags blocks.
+
+Resources:
+
+  # WAF regex pattern sets are capped at 10 patterns each, so the list of
+  # known bot user-agent substrings is split across several sets and combined
+  # with an OrStatement in the known-bots rule below.
+  KnownBotsPatternSet0:
+    Type: AWS::WAFv2::RegexPatternSet
+    Properties:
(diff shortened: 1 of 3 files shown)


waf-template.yaml
```diff
@@ -0,0 +1,235 @@
...
+      Name: SimpleBotProtectionWAF
+      Scope: CLOUDFRONT
+      DefaultAction:
+        Allow: {}
+      VisibilityConfig:
+        CloudWatchMetricsEnabled: true
+        MetricName: SimpleBotProtectionWAF
+        SampledRequestsEnabled: true
+      Rules:
+        # Rate limiting rule: 500 requests per 5 min per IP
```
chris48s, line 158: I think it is probably fine, but is there any risk that BallotCacheWriter
https://github.com/DemocracyClub/BallotCacheWriter/blob/e24bf0a39981a183fb58d4fade6af016d9281387/ballot_writer/get_ballots.py#L45
might ever exceed this?

chris48s:
One question, but I've had a poke about on dev and this looks good :+1:

PR #2409, opened 2026-06-11
don't fiddle with PCC/mayor titles

chris48s:
If it is a division-less election, just spit out the ballot title


GeoWill (approved):
For completeness we should add `by-election` to the election name on `pcc.north-yorkshire.by.2021-11-25` and `pcc.northumbria.by.2019-07-18`.

PR #2413, opened 2026-07-09
improve 'recent leaflets' section

chris48s:
Refs https://app.asana.com/1/1204880536137786/project/1204880927741389/task/1216109485888378?focus=true

- Show leaflets from the last year only
- If we know older leaflets are available, link to EL but don't show them inline

### DIFF
--- a/locale/cy/LC_MESSAGES/django.po
+++ b/locale/cy/LC_MESSAGES/django.po
@@ -2594,7 +2594,7 @@ msgstr ""
 #: wcivf/apps/people/templates/people/includes/_person_leaflets.html:51
 #, python-format
 msgid "Leaflets from %(person_name)s"
-msgstr "Taflenni gan %(person_name)s"
+msgstr ""
 
 #: wcivf/apps/people/templates/people/includes/_person_leaflets.html:56
 #, python-format
@@ -2603,9 +2603,6 @@ msgid ""
 "%(person_id)s\"> View leaflets from %(person_name)s</a> <a class=\"ds-cta\" "
 "href=\"https://electionleaflets.org/leaflets/add/\"> Upload a leaflet </a>"
 msgstr ""
-"<a class=\"ds-cta\" href=\"https://electionleaflets.org/person/"
-"%(person_id)s\">Gweld taflenni gan %(person_name)s</a> <a class=\"ds-cta\" "
-"href=\"https://electionleaflets.org/leaflets/add/\">Uwchlwythwch daflen</a>"
 
 #: wcivf/apps/people/templates/people/includes/_person_local_party_card.html:6
 #, python-format
--- a/locale/cy/LC_MESSAGES/django.po
+++ b/locale/cy/LC_MESSAGES/django.po
@@ -11,7 +11,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2026-06-17 14:25+0000\n"
+"POT-Creation-Date: 2026-07-09 10:18+0000\n"
 "PO-Revision-Date: 2022-03-24 09:04+0000\n"
 "Last-Translator: Sym Roe <sym@talusdesign.co.uk>, 2022\n"
 "Language-Team: Welsh (https://www.transifex.com/democracy-club/teams/61326/"
@@ -308,28 +308,28 @@ msgstr " (wedi'i ohirio)"
 msgid " (uncontested)"
 msgstr "(dim cystadleuaeth)"
 
-#: wcivf/apps/elections/models.py:870
+#: wcivf/apps/elections/models.py:874
 #, python-format
 msgid "This by-election was called because %(reason)s."
 msgstr "Canslwyd yr etholiad hwn %(reason)s."
 
-#: wcivf/apps/elections/models.py:1007
+#: wcivf/apps/elections/models.py:1011
 msgid "First-past-the-post"
 msgstr "Y Cyntaf i’r Felin"
 
-#: wcivf/apps/elections/models.py:1009
+#: wcivf/apps/elections/models.py:1013
 msgid "Additional Member System"
 msgstr "Y System Aelodau Ychwanegol"
 
-#: wcivf/apps/elections/models.py:1011
+#: wcivf/apps/elections/models.py:1015
 msgid "Supplementary Vote"
 msgstr "Pleidlais Atodol"
 
-#: wcivf/apps/elections/models.py:1013
+#: wcivf/apps/elections/models.py:1017
 msgid "Single Transferable Vote"
 msgstr "Pleidlais Sengl Drosglwyddadwy"
 
-#: wcivf/apps/elections/models.py:1015
+#: wcivf/apps/elections/models.py:1019
 msgid "Closed-List Proportional Representation"
 msgstr "Cynrychiolaeth Gyfrannol â Rhestrau Caeedig"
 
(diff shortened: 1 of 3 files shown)


locale/cy/LC_MESSAGES/django.po
```diff
@@ -2591,6 +2591,22 @@ msgstr ""
...
+#, python-format
+msgid ""
+"<a class=\"ds-cta\" href=\"https://electionleaflets.org/person/"
+"%(person_id)s\"> View leaflets from %(person_name)s</a> <a class=\"ds-cta\" "
+"href=\"https://electionleaflets.org/leaflets/add/\"> Upload a leaflet </a>"
+msgstr ""
+"<a class=\"ds-cta\" href=\"https://electionleaflets.org/person/"
+"%(person_id)s\">Gweld taflenni gan %(person_name)s</a> <a class=\"ds-cta\" "
+"href=\"https://electionleaflets.org/leaflets/add/\">Uwchlwythwch daflen</a>"
+
```
chris48s: I've had a go at these as they're only 1 word changes, but maybe it better to se them to `msgstr ""`?

wcivf/apps/leaflets/models.py
```diff
@@ -1,10 +1,15 @@
+import datetime as dt
+
 from django.db import models
 from people.models import Person
 
 
 class LeafletQuerySet(models.QuerySet):
     def latest_four(self):
-        return self.order_by("date_uploaded_to_electionleaflets")[:4]
```
chris48s, line 7: One of the issues with this was that this was selecting the 4 oldest leaflets, not the 4 most recent.

This should have really been `.order_by("-date_uploaded_to_electionleaflets")`

wcivf/apps/leaflets/models.py
```diff
@@ -1,10 +1,15 @@
...
+        return self.filter(
+            date_uploaded_to_electionleaflets__gte=dt.datetime.now()
```
GeoWill: Doesn't functionally matter with a time delta of 1year, but I noticed it so will mention. This causes noise in the logs like:

```
RuntimeWarning: DateTimeField Leaflet.date_uploaded_to_electionleaflets received a naive datetime while time zone support is active.
```

I think you can solve this by changing 

```python
dt.datetime.now()  - dt.timedelta(days=365)
```

to

```python
from django.utils import timezone
...
timezone.now()  - dt.timedelta(days=365)
```

wcivf/apps/people/templates/people/includes/_person_leaflets.html
```diff
@@ -3,39 +3,64 @@
...
+            {% endfor %}
+        </ul>
+
+        <div class="ds-cluster-left">
+            <div>
+                {% blocktrans trimmed with person_name=object.name person_id=object.ynr_id %}
+                    <a class="ds-cta" href="https://electionleaflets.org/person/{{ person_id }}">
+                        More leaflets from {{ person_name }}</a>
+                    <a class="ds-cta" href="https://electionleaflets.org/leaflets/add/">
+                        Upload a leaflet
```
GeoWill: This is already covered in:
https://github.com/DemocracyClub/WhoCanIVoteFor/blob/44aa90b247ed99d9da1f0a68920584e6ef46a15d/wcivf/apps/people/templates/people/includes/_person_edit_details_card.html#L18

That card only shows if the person has current or future candidacies. I think we should only show it once. Either in this partial or the other. ~I don't mind which.~ I've made suggestion below that would always include it in this template, so maybe we can just delete the other one. I think this template also only shows if the person has current or future candidacies.

wcivf/apps/people/templates/people/includes/_person_leaflets.html
```diff
@@ -3,39 +3,64 @@
...
+            </div>
+        </div>
+    </details>
+{% else %}
+
+    {# We know one or more leaflets exist #}
+    {# but all of them are >1 year old #}
+    <h2 class="ds-h3">
+        {% blocktrans with person_name=object.name %}Leaflets from {{ person_name }}{% endblocktrans %}
+    </h2>
```
GeoWill, line 53: I think this heading is more in keeping with the rest of the page, than the `<h4>` used in the `<summary>` block for recent leaflets. 

-------------------

**No recent leaflets;**

Image

-------------------

vs 

-------------------

**Has recent leaflets:**

Image

-------------------
I would be tempted to make the 'has recent leaflets' case look something like: 

Image

wcivf/apps/leaflets/models.py
GeoWill: 👍 

Unrelated, it's a shame there's no equivalent to `@cached_property` like on a model.

locale/cy/LC_MESSAGES/django.po
GeoWill: I guess it depends how confident you are! setting them to `msgstr ""` will make it more likely they'll get a second set of eyes. I have no way (beyond google translate) to judge.

GeoWill (commented):
I think this is fine, but do have a couple of thoughts.

If you're busy with other stuff it would be fine to merge as is.

wcivf/apps/people/templates/people/includes/_person_leaflets.html
GeoWill: Here's a patch for what I suggested. 

<details><summary>Patch</summary>

```diff
diff --git a/wcivf/apps/people/templates/people/includes/_person_leaflets.html b/wcivf/apps/people/templates/people/includes/_person_leaflets.html
index d4caa4a1..e7c03df2 100644
--- a/wcivf/apps/people/templates/people/includes/_person_leaflets.html
+++ b/wcivf/apps/people/templates/people/includes/_person_leaflets.html
@@ -5,10 +5,13 @@
 
 
 {% if object.leaflet_set.latest_four %}
+    <h2 class="ds-h3">
+        {% blocktrans with person_name=object.name %}Leaflets from {{ person_name }}{% endblocktrans %}
+    </h2>
     <details class="ds-stack">
         <summary>
             <h4>
-                {% blocktrans with person_name=object.name %}Recent leaflets from {{ person_name }}{% endblocktrans %}
+                {% blocktrans with person_name=object.name %}Recent leaflets{% endblocktrans %}
             </h4>
         </summary>
 
@@ -31,18 +34,19 @@
             {% endfor %}
         </ul>
 
-        <div class="ds-cluster-left">
-            <div>
-                {% blocktrans trimmed with person_name=object.name person_id=object.ynr_id %}
-                    <a class="ds-cta" href="https://electionleaflets.org/person/{{ person_id }}">
-                        More leaflets from {{ person_name }}</a>
-                    <a class="ds-cta" href="https://electionleaflets.org/leaflets/add/">
-                        Upload a leaflet
-                    </a>
-                {% endblocktrans %}
-            </div>
-        </div>
+
     </details>
+    <div class="ds-cluster-left">
+        <div>
+            {% blocktrans trimmed with person_name=object.name person_id=object.ynr_id %}
+                <a class="ds-cta" href="https://electionleaflets.org/person/{{ person_id }}">
+                    More leaflets from {{ person_name }}</a>
+                <a class="ds-cta" href="https://electionleaflets.org/leaflets/add/">
+                    Upload a leaflet
+                </a>
+            {% endblocktrans %}
+        </div>
+    </div>
 {% else %}
 
     {# We know one or more leaflets exist #}
```
</details>

locale/cy/LC_MESSAGES/django.po
chris48s: Yeah. Me either.
I've decided to just delete them.
24b7e40a

wcivf/apps/leaflets/models.py
chris48s: Good call 96620bca

wcivf/apps/people/templates/people/includes/_person_leaflets.html
chris48s: Agreed. I've applied your patch in f167955c

wcivf/apps/people/templates/people/includes/_person_leaflets.html
chris48s: Yeah good spot.
I think you're right that we should only have the "add leaflets" CTA in one place.

I think I'm going to put this on ice until next week. I'll run through it with Peter quickly once he is back and decide what we want to do then finish this off.

chris48s:
Had a quick chat with Peter about this. We decided to just go with the "Upload your leaflets" CTA at the bottom of the page under "That's all we know..."
We think separating the (read-only) "profile" from the CTAs to add data is good.
Also small wording change on the button if we know all leaflets are >1 year old.

locale/cy/LC_MESSAGES/django.po
```diff
@@ -2580,16 +2585,26 @@ msgstr ""
...
 msgid ""
 "<a class=\"ds-cta\" href=\"https://electionleaflets.org/person/"
-"%(person_id)s\"> More leaflets from %(person_name)s</a> <a class=\"ds-cta\" "
-"href=\"https://electionleaflets.org/leaflets/add/\"> Upload a leaflet </a>"
+"%(person_id)s\"> View past leaflets from %(person_name)s</a>"
 msgstr ""
 "<a class=\"ds-cta\" href=\"https://electionleaflets.org/person/"
-"%(person_id)s\">Mwy o daflenni gan %(person_name)s</a> <a class=\"ds-cta\" "
-"href=\"https://electionleaflets.org/leaflets/add/\">Uwchlwythwch daflen</a>"
+"%(person_id)s\">Mwy o daflenni gan %(person_name)s</a>"
```
chris48s, line 2607: I think keeping a translation that is off by one word is probably better than just deleting it here.

GeoWill (approved):

PR #2414, opened 2026-07-15
import timetable from EE

chris48s:
Similar PR to
https://github.com/DemocracyClub/yournextrepresentative/pull/2780
and to a lesser extent
https://github.com/DemocracyClub/ec-postcode-lookup-pages/pull/245


wcivf/apps/elections/models.py
```diff
@@ -620,67 +641,41 @@ def update_candidate_ranks(self):
...
-    @property
-    def registration_deadline(self):
-        try:
-            date = get_election_timetable(
-                self.ballot_paper_id, self.post.territory
-            ).registration_deadline
-        except AttributeError:
-            return None
-
-        return date.strftime("%d %B %Y")
```
chris48s, line 641: This was a bit of a mess

Some of these functions were returning a `dt.date()` object

Some of them were calling `.strftime("%d %B %Y")` on them here at the model layer

I've changed it so everything is just a `dt.date()` object at the model layer and formatting happens in the template. I've made sure everything formats right. For some reason we were already passing everything through a `|naturalday` filter on the template side even if we'd already formatted it here, so no updates were needed at the template layer.

Side note:
One thing I really like about the EC site is we've got a single `date_format` filter we use everywhere.
This application is a right old mess of calls to `|naturalday` which don't all use the same params.
I would like to tidy that up at some point, but not in this PR

wcivf/apps/elections/models.py
```diff
@@ -573,6 +572,28 @@ class PostElection(TimeStampedModel):
...
+    registration_deadline = models.DateField(
+        blank=True, null=True, help_text="Register to vote deadline"
+    )
+    postal_vote_application_deadline = models.DateField(
+        blank=True, null=True, help_text="Postal vote application deadline"
+    )
+    vac_application_deadline = models.DateField(
+        blank=True, null=True, help_text="VAC application deadline"
+    )
+
```
chris48s, line 596: We don't use every single date in this application
but we use enough that we might as well just sync/store them all tbh

pyproject.toml
```diff
@@ -20,7 +20,7 @@ dependencies = [
     "dc-logging-utils",
     "djangorestframework-jsonp==1.0.2",
     "feedparser==6.0.10",
-    "uk-election-timetables==4.4.0",
+    "uk-election-timetables==5.0.0",
```
chris48s, line 23: I'm going to handle this in the same style as
https://github.com/DemocracyClub/yournextrepresentative/pull/2780
and
https://github.com/DemocracyClub/yournextrepresentative/pull/2781

This PR upgrades the lib and sets the historic dates on everything

..and then another PR to remove the library.

chris48s: https://github.com/DemocracyClub/WhoCanIVoteFor/pull/2415

symroe (approved):

